PR #67, opened 2020-04-02
Update filleeventqueries.md

code42-ceciliastevens:
Added the markdown for generating file event client docs


code42-ceciliastevens:
Ah, silly me, should've checked. I'll close this one then.

PR #72, opened 2020-04-08
Added high risk employee feature

kiran-chaudhary:
Added high risk employee feature.


src/py42/_internal/client_factories.py
```diff
@@ -94,6 +98,18 @@ def _get_stored_value(self, key):
             self._key_value_store_client = KeyValueStoreClient(session)
         return self._key_value_store_client.get_stored_value(key).text
 
+    def get_high_risk_employee_client(self):
+        # TODO figure out the correct session instance to be created.
```
alanag13, line 102: `session = self._get_jwt_session(u"employeecasemanagement-API_URL")`

src/py42/_internal/client_factories.py
```diff
@@ -94,6 +98,18 @@ def _get_stored_value(self, key):
...
+        session = ""
+        if not self._high_risk_employee_client:
+            self._high_risk_employee_client = HighRiskEmployeeClient(session, self._user_context)
+
+        # TODO similarly detection_list too but since there is no property to instantiate it, need
```
alanag13, line 107: this client would also use `session = self._get_jwt_session(u"employeecasemanagement-API_URL")`

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def get(self, user_id):
+        data = {"tenantId": self._user_context.get_current_tenant_id(), "userId": user_id}
+        resource = u"/user/getbyid"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def search(
```
alanag13, line 47: We would want this to be a generator that automatically handles paging, similar to the `get_all` functions in `user`, `org`, and `device`. In fact, this should probably be called `get_all` to be consistent.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+        page_size=100,
+        page_num=1,
+    ):
+        """
+        Search High Risk employee list. Filter results by filter_type or risk factors.
+
+        #TODO Param description
+        """
+
+        # Defined default values for page_size & page_num even though in requirement
```
alanag13, line 62: this is fine, these are good defaults

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+
+        data = {
+            "tenantId": self._user_context.get_current_tenant_id(),
+            "userId": user_id,
+            "notes": notes,
+        }
+        uri = u"{0}/{1}".format(self._uri_prefix, u"updatenotes")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add_risk_tag(self, user_id, tags=None):
```
alanag13, line 105: tags should not be an optional param here

src/py42/_internal/client_factories.py
kiran-chaudhary: Here I meant to ask, When will `DetectionListUserClient be` instantiated? 
In this method only `HighRiskEmployeeClient` should be instantiated.

since we directly call `sdk.detectionlist.update_notes` when will `self._detection_list_user_client` be defined?

Whereas for `self.detectionlist.high_risk_employee` will define `self._high_risk_employee_client`.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: ok, I will revert this change.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+
+        tags = tags if tags else []
+        # should we return if risk_factors is empty?
+        # if so, in that case the response would not be a Py42Response, or should we create it?
```
alanag13, line 109: in all cases where we are adding or removing list items (for risk tags and cloud aliases), we could throw a `ValueError` if `tags` is None or an empty list.

src/py42/_internal/client_factories.py
alanag13: `get_detection_list_user_client` should be its own function though

src/py42/modules/detectionlists.py
```diff
@@ -5,3 +5,37 @@ def __init__(self, microservice_client_factory):
...
+        # Changed method signature, used 'tags' instead of 'tag' as compared to requirement,
+        # as the API expects a list.
+
+        # Suggestion:
+        # we can allow user to pass either of string or list and manage it on our side as
+        # below, i.e if its not a list, make a list and pass it to the API.
+        # if type(tags) is str:
+        #   tags = [tags]
+
+        # I don't see much validation done in other modules as well, do we need to add
```
alanag13, line 29: In this case i think it would be helpful to throw an error. We can re-evaluate other modules on an as-needed basis but I'm trying to keep the code simple and in many cases we just allow the user to fail.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+        data = {
+            "tenantId": self._user_context.get_current_tenant_id(),
+            "userId": user_id,
+            "riskFactors": tags,
+        }
+        uri = u"{0}/{1}".format(self._uri_prefix, u"removeriskfactors")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add_cloud_alias(self, user_id, aliases=None):
```
alanag13, line 130: aliases should not be an optional argument

src/py42/modules/detectionlists.py
```diff
@@ -5,3 +5,37 @@ def __init__(self, microservice_client_factory):
     @property
     def departing_employee(self):
         return self._microservice_client_factory.get_departing_employee_client()
+
+    @property
+    def high_risk_employee(self):
```
alanag13, line 10: all the user methods should get the detection_list_user_client by calling `self._microservice_client_factory.get_detection_list_user_client()` once it exists.

alanag13:
This is very good, essentially exactly what I was thinking. You just have to create the `microservice_client_factory.get_detection_list_user_client()` function and use it in each relevant method in the `DetectionListModule` (the client will only be created once, the other times it will just be retrieved).

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,150 @@
...
+    # In other modules URI is a single variable, breaking it up to have better control
+    # if we want to refactor version and prefix to a common configuration.
+    _api_version = u"v2"
+    _uri_prefix = u"/svc/api/{0}".format(_api_version)
+
+    def __init__(self, session, user_context):
+        super(HighRiskEmployeeClient, self).__init__(session)
+        self._user_context = user_context
+
+    def add(self, user_id):
```
kiran-chaudhary, line 20: @alanag13 
Here, we don't create user, `/user/create/` endpoint, directly we add `/highriskemployee/add`. I removed that part after change in requirement, Is this ok? From my understanding and testing I think, user existence is a pre-requisite for adding.

Difference in code can be seen between commits, in `create` method v/s this method.

src/py42/modules/detectionlists.py
```diff
@@ -1,7 +1,58 @@
...
+        return self._microservice_client_factory.get_high_risk_employee_client()
+
+    def update_notes(self, user_id, notes):
+        # TODO Refactor, this condition should not be needed, relates to TODO fix above
+        if not self._detection_list_user_client:
```
alanag13, line 21: You don't need these if statements (or the private `_detection_list_user_client` and can simply get the client using `get_detection_list_user_client` in each method because the client only gets created once in that method.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,205 @@
...
+        return self._session.post(uri, data=json.dumps(data))
+
+    def _add_high_risk_employee(self, tenant_id, user_id):
+
+        data = {"tenantId": tenant_id, "userId": user_id}
+        resource = u"/highriskemployee/add"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add(self, user_id=None, username=None):
```
alanag13, line 39: The `userId` is the userUid of the the code42 user (can be seen using `/api/User`, so `username` doesn't have to be a parameter here, we can have this module take a dependency on `userclient` and get the user name for us.

src/py42/_internal/client_factories.py
kiran-chaudhary: Done.

src/py42/_internal/client_factories.py
kiran-chaudhary: Done

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Done

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Done

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Observed, API response with 400 Bad request error, during such cases, so left it as is.

src/py42/modules/detectionlists.py
kiran-chaudhary: API returns bad request error, same as earlier comment.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed

src/py42/modules/detectionlists.py
kiran-chaudhary: Done.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed to allow only user_id.

src/py42/modules/detectionlists.py
kiran-chaudhary: Removed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed to accept only used_id.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+
+        tenant_id = self._user_context.get_current_tenant_id()
+        return self._add_high_risk_employee(tenant_id, user_id)
+
+    def set_alerts_enabled(self, enabled=True):
+        """
+        Enable alerts.
+        #TODO Confirm
+        # `toggle_alerts` would be a better method name, if we intend to allow to disable
```
alanag13, line 51: `toggle` implies that you are changing something to the opposite of what it currently is. a `toggle` function would not require an argument.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        self._user_context = user_context
+
+    def _add_high_risk_employee(self, tenant_id, user_id):
+
+        data = {"tenantId": tenant_id, "userId": user_id}
+        resource = u"/highriskemployee/add"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add(self, user_id=None):
```
alanag13, line 30: user_id should not be an optional parameter

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        Args:
+            user_id (str/int): The user_id whose tag(s) needs to be updated.
+            tags (str or list of str ): A single tag or multiple tags in a list to be added.
+                e.g "tag1" or ["tag1", "tag2"]
+
+        Returns:
+            :class:`py42.sdk.response.Py42Response`
+        """
+
+        if type(tags) is str:
```
alanag13, line 190: when using `str`, import it from `py42._internal.compat`.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        Get user information.
+
+        Args:
+            user_id (str/int): Id of the user who needs to be added to HRE lens.
+
+        Returns:
+            :class:`py42.sdk.response.Py42Response`
+        """
+        data = {"tenantId": self._user_context.get_current_tenant_id(), "userId": user_id}
+        resource = u"/user/getbyid"
```
alanag13, line 93: this should use `highriskemployee/get`.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+    """
+
+    # In other modules URI is a single variable, breaking it up to have better control
+    # if we want to refactor version and prefix to a common configuration.
+    _api_version = u"v2"
+    _uri_prefix = u"/svc/api/{0}".format(_api_version)
+
+    def __init__(self, session, user_context):
+        super(HighRiskEmployeeClient, self).__init__(session)
+        self._user_context = user_context
```
alanag13, line 21: pass in a `DetectionListUserClient` here.

src/py42/_internal/client_factories.py
```diff
@@ -84,6 +90,18 @@ def _get_stored_value(self, key):
             self._key_value_store_client = key_value_store.KeyValueStoreClient(session)
         return self._key_value_store_client.get_stored_value(key).text
 
+    def get_high_risk_employee_client(self):
+        if not self._high_risk_employee_client:
+            session = self._get_jwt_session(u"employeecasemanagement-API_URL")
+            self._high_risk_employee_client = HighRiskEmployeeClient(session, self._user_context)
```
alanag13, line 96: pass in a `DetectionListUserClient` here

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+            self._get_high_risk_employees_page,
+            "items",
+            tenant_id=self._user_context.get_current_tenant_id(),
+            filter_type=filter_type,
+            sort_key=sort_key,
+            sort_direction=sort_direction,
+        )
+
+
+class DetectionListUserClient(BaseClient):
```
alanag13, line 147: this client should probably go in its own file.

alanag13, line 147: We need the `get` and `create` api methods to be added here

docs/methoddocs/high_risk_employee.md
```diff
@@ -0,0 +1,15 @@
+# High Risk Employee
```
unparalleled-js, line 1: I think these headings are out of order

docs/methoddocs/high_risk_employee.md
```diff
@@ -0,0 +1,15 @@
...
+
+```eval_rst
+.. autoclass:: py42.modules.detectionlists.DetectionListsModule
+    :members:
+    :show-inheritance:
+```
+
+## Detection List
```
unparalleled-js, line 9: Should be `Detection Lists` (plural). Also, probably should include a newline under it like on the other one

docs/methoddocs/high_risk_employee.md
```diff
@@ -0,0 +1,15 @@
+# High Risk Employee
+
+```eval_rst
+.. autoclass:: py42.modules.detectionlists.DetectionListsModule
```
unparalleled-js, line 4: There is also a departing employee client that is a detection list but it is on its .md file. Do we want to include that one here and just name the file `detectionlists.md`?

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        resource = u"/highriskemployee/remove"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def get(self, user_id):
+        """
+        Get user information.
+
+        Args:
+            user_id (str/int): Id of the user who needs to be added to HRE lens.
```
alanag13, line 87: id of user who _has been added_ to the High Risk Employee detection list.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+            "tenantId": self._user_context.get_current_tenant_id(),
+            "alertsEnabled": enabled,
+        }
+        resource = u"/highriskemployee/setalertstate"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def remove(self, user_id):
+        """
+        Remove an user from high risk employee lens.
```
alanag13, line 69: use `detection list` instead of `lens`

docs/methoddocs/high_risk_employee.md
unparalleled-js, line 1: We should also include an` autoclass` for `DetectionListUserClient`

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def _add_high_risk_employee(self, tenant_id, user_id):
+
+        data = {"tenantId": tenant_id, "userId": user_id}
+        resource = u"/highriskemployee/add"
+        uri = u"{0}{1}".format(self._uri_prefix, resource)
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add(self, user_id=None):
+        """
```
alanag13, line 31: Once they are added to `DetectionListUserClient`, use the `get` method to see if the id has already been created and `create` method if it has not, _then_ call `add_high_risk_employee()`.

docs/methods.md
```diff
@@ -24,6 +24,7 @@ Explore the complete public documentation for `py42` below.
 * [Security Data](methoddocs/securitydata.md)
 * [Legal Hold](methoddocs/legalhold.md)
 * [Departing Employee](methoddocs/departingemployee.md)
+* [High Risk Employee Lens](methoddocs/high_risk_employee.md)
```
unparalleled-js, line 27: I think we are avoiding calling them `lens` in the SDK since that is more of a branding term and could go away. We are sticking to `detection lists`. Also, it should follow the same convention / format as the `Departing Employee` item listed right above it, unless they get consolidated (see my other PR comments). 

I think `High Risk Employee` would work, Code42 users should know what that is.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class HighRiskEmployeeClient(BaseClient):
+    """
+    Administrator utility to manage High Risk employees.
+    `Support Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Detection_list_management_APIs>`__
+
+    """
+
+    # In other modules URI is a single variable, breaking it up to have better control
```
unparalleled-js, line 14: You can remove this comment, it is fine either way. We don't have to refactor anything right now. It may be a good idea for some APIs. Others, like this one, have changed dramatically between v1 and v2.

docs/methoddocs/high_risk_employee.md
alanag13: all of its methods are exposed on `DetectionListModule`, so no need

alanag13: `DetectionListUserClient` could be made internal.

src/py42/clients/detectionlists/high_risk_employee.py
unparalleled-js: The user has more control when they can explicitly enable or disable, I think. That way, if there is a script where they just want to turn it on always, they can just always do `set_alerts_enabled(True)` without having to care about the current state of `alerts_enabled`.

src/py42/clients/detectionlists/high_risk_employee.py
unparalleled-js, line 69: Remove _a_ user

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+    def __init__(self, session, user_context):
+        super(DetectionListUserClient, self).__init__(session)
+        self._user_context = user_context
+
+    def update_notes(self, user_id, notes):
+        """Add or update notes related to the user.
+
+        Args:
+            user_id (str/int): The user_id whose notes need to be updated.
+            notes (str): Note.
```
unparalleled-js, line 165: How about instead of `Note.`, we do `User profile notes.`

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+            "notes": notes,
+        }
+        uri = u"{0}/{1}".format(self._uri_prefix, u"updatenotes")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add_risk_tag(self, user_id, tags):
+        """ Add one or more tags.
+
+        Args:
+            user_id (str/int): The user_id whose tag(s) needs to be updated.
```
unparalleled-js, line 182: `str or int` to match the arg style below?

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        data = {
+            "tenantId": self._user_context.get_current_tenant_id(),
+            "userId": user_id,
+            "notes": notes,
+        }
+        uri = u"{0}/{1}".format(self._uri_prefix, u"updatenotes")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def add_risk_tag(self, user_id, tags):
```
unparalleled-js, line 178: Should this method be named `add_risk_tags` so that it is more evident it can take a list

unparalleled-js: Same with `remove_risk_tags`

alanag13: Yes, I agree that that is a better name

docs/methoddocs/high_risk_employee.md
unparalleled-js: Let's make it internal!

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Renamed

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Removed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: cannot call create method, since we need `username` to call it, just added a check to verify user_id exists in detection list.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Done

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Added.

kiran-chaudhary: Refactored.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Done.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Changed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Done.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: > Missed removing it. Changed.

docs/methods.md
kiran-chaudhary: Changed.

docs/methoddocs/high_risk_employee.md
kiran-chaudhary: Changed.

src/py42/clients/detectionlists/high_risk_employee.py
kiran-chaudhary: Passing `False` to `set_alerts_enabled` sounds contradictory, enabled and false !!  
So I thought a user might call this function only when they intend to change the state and hence suggested for toggling. 
Removed comment.

docs/methoddocs/high_risk_employee.md
unparalleled-js: Don't forget to rename it to `detectionlists.md` if we go this route

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,140 @@
...
+            - a user already exists in high risk employee detection list
+            - user_id is not found in detection lists
+
+        Args:
+            user_id (str or int): Id of the user who needs to be added to HRE detection list.
+
+        Returns:
+            :class:`py42.sdk.response.Py42Response`
+        """
+        if self._user_client.get_by_id(user_id):
```
alanag13, line 48: Let's bring in `py42.clients.UserClient` as a dependency and use it to determine the id's username if it has not already been already added to `DetectionLists`. Once have the username, we can call `DectectionUserClient.create()`, then call `_add_high_risk_employee()` as normal.

kiran-chaudhary: Done, do note, used `user.client`.`get_by_uid` instead of `get_by_id` as `user_id` in detection list refers to `uid` of `user_client`

docs/methoddocs/high_risk_employee.md
kiran-chaudhary: Changed.

docs/methoddocs/high_risk_employee.md
kiran-chaudhary: After making this change, facing below error while generating docs. Not able to figure out the reason. 

I have tried, `make clean ` too.

```
WARNING: autodoc: failed to import class 'high_risk_employee.HighRiskEmployeeClient' from module 'py42.clients.detectionlists'; the following exception was raised:
cannot import name 'HighRiskEmployeeClient'
```

src/py42/clients/detectionlists/high_risk_employee.py
alanag13: Works for me!

docs/methoddocs/Detectionlists.md
```diff
@@ -0,0 +1,25 @@
+# Detection Lists
```
unparalleled-js, line 1: None of the other files in `methoddocs` begin with a capital letter.
This should be `detectionlists.md`.

unparalleled-js: Otherwise, this looks good!

src/py42/clients/detectionlists/detection_list_user_client.py
```diff
@@ -0,0 +1,180 @@
...
+        """Create a detection list profile for a user.
+
+        Args:
+            username (str): Username of the user.
+
+        Returns:
+            :class:`py42.sdk.response.Py42Response`
+        """
+        data = {
+            "tenantId": self._user_context.get_current_tenant_id(),
```
unparalleled-js, line 34: Do we want to still allow users to specify a tenant ID if they want?
@alanag13

unparalleled-js: An example would be departing employee client - it only gets the current tenant ID if none is passed in.

src/py42/clients/detectionlists/high_risk_employee.py
unparalleled-js: TIL

docs/methoddocs/high_risk_employee.md
kiran-chaudhary: @alanag13 @unparalleled-js : I am facing a weird error,  due to below line in `high_risk_employee.py` document generation is failing as mentioned in previous comment. 

`from py42.sdk.exceptions import Py42NotFoundError, Py42BadRequestError`

Any pointers on how to fix this?!! Can you please checkout this branch and confirm that you also are facing the same error.  

Also tested the same by creating a fresh virtual environment. If I remove this line document gets generated without any error.

docs/methoddocs/Detectionlists.md
kiran-chaudhary: Yes, I have fixed this but as mentioned in other comment, I am facing an error and hence have not pushed the change.

alanag13:
@unparalleled-js the spec didn't include it, and I think I'm in favor of leaving it out because nearly every use case does not want or need it. Only someone internal could have cross-tenant access, so this wouldn't be used in ordinary integrations. Leaving it off puts a little less mental burden of "what is this? do i need this?" on the user, and if we find out that there  _is_ a legitimate case where somebody needs this option, it's very easy to add in without breaking anything.

docs/methoddocs/high_risk_employee.md
alanag13: @kiran-chaudhary I will look into this, we have seen similar problems before.

alanag13:
@kiran-chaudhary besides the error that you mentioned (that I will look into), this looks great!

unparalleled-js:
> @unparalleled-js the spec didn't include it, and I think I'm in favor of leaving it out because nearly every use case does not want or need it. Only someone internal could have cross-tenant access, so this wouldn't be used in ordinary integrations. Leaving it off puts a little less mental burden of "what is this? do i need this?" on the user, and if we find out that there _is_ a legitimate case where somebody needs this option, it's very easy to add in without breaking anything.

Works for me! Thanks for the explanation.

unparalleled-js:
@kiran-chaudhary @alanag13 I figured out what is causing that error... it's importing the exceptions in `high_risk_employee.py`

README.md
```diff
@@ -102,7 +102,7 @@ To override these settings, import `py42.sdk.settings` and override values as ne
```
unparalleled-js, line 102: Looks like on line 100, still is using `py42.sdk.settings`

unparalleled-js:
`AlertClient` still references `py42.sdk.reponse.Py42Response` in return types

src/py42/_internal/clients/detection_list_user.py
```diff
@@ -0,0 +1,174 @@
+import json
```
unparalleled-js, line 1: There are numerous missing `u""` on all these data strings... probably want to make sure we add those.

alanag13: i'll look over those

README.md
alanag13: good catch

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,145 @@
...
+            sort_direction (str): "ASC" or "DESC"
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response` objects
+            that each contain a page of users.
+        """
+
+        return get_all_pages(
+            self._get_high_risk_employees_page,
+            "items",
```
unparalleled-js, line 140: Missing `u""`

unparalleled-js:
`exceptions.py` mentions the old path in the docstr of method `raise_py42_error`

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -0,0 +1,145 @@
...
+        tenant_id,
+        filter_type=None,
+        sort_key=None,
+        sort_direction=None,
+        page_num=None,
+        page_size=None,
+    ):
+        # Overwriting page_size since default value 1000 returns error
+        page_size = 100
+        data = {
```
unparalleled-js, line 113: Missing `u""`s heres

docs/methoddocs/Detectionlists.md
unparalleled-js, line 1: This file name still starts with a capital letter, causing the docs not to build

alanag13: wat, i just changed this howwwwww

unparalleled-js: It is because of GitHub
it's annoying

unparalleled-js: I think the only way to fix it is to first delete it. Commit. then recreate it

alanag13: you can rename it in the browser

unparalleled-js: TIL

PR #83, opened 2020-04-21
[WIP] Proposed design structure for Alert Rules.

kiran-chaudhary:
This PR is raised to finalise design approach for adding/removing users to/from alert rules.
https://jira.corp.code42.com/browse/INTEG-990

Added comments inline describing the idea.

Added some code to relate to the idea. This PR is not intended to be merged.

### DIFF
--- a/src/py42/clients/alertrules.py
+++ b/src/py42/clients/alertrules.py
@@ -1,0 +1,82 @@
+from py42.clients import BaseClient
+
+# Below comments are to discuss design and is not intended to be official documentation.
+
+
+class AlertRulesClient(BaseClient):
+    """
+    `add` and `remove_users` require ids, i.e ruleIds and userIds, irrespective of which
+    type of rules they belong to, whereas search is dependent on rule type.
+
+    {
+      "tenantId": "string",
+      "ruleId": "string",
+      "userList": [
+        {
+          "userIdFromAuthority": "string",
+          "userAliasList": [
+            "string"
+          ]
+        }
+      ]
+    }
+
+    {
+      "tenantId": "string",
+      "ruleId": "string",
+      "userIdList": [
+        "string"
+      ]
+    }
+
+    Search create & update have 3 different endpoints for each task, e.g for search
+        /api/v1/Rules/query-cloud-share-permissions-rule
+        /api/v1/Rules/query-endpoint-exfiltration-rule
+        /api/v1/Rules/query-file-type-mismatch-rule
+
+    1. Design it similar to detectionlists
+        Now based on future requirement to add/update rules etc,
+        We can have module AlertRules and below clients
+            CloudShare
+            EndpointExfiltration
+            FileTypeMismatch
+
+         e.g sdk.alertrules.cloudshare.search(),
+             sdk.alertrules.endpointexfiltration.search(),
+             sdk.alertrules.filetypemismatch.search(), similarly for add and update.
+
+             sdk.alertrules.add_users(), sdk.alertrules.remove_users(), etc
+
+    2. Create an AlertRulesClient focusing only on current requirement.
+
+        sdk.alertrules.add_users(),
+        sdk.alertrules.remove_users(),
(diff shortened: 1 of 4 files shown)


src/py42/clients/alertrules.py
```diff
@@ -0,0 +1,82 @@
...
+            EndpointExfiltration
+            FileTypeMismatch
+
+         e.g sdk.alertrules.cloudshare.search(),
+             sdk.alertrules.endpointexfiltration.search(),
+             sdk.alertrules.filetypemismatch.search(), similarly for add and update.
+
+             sdk.alertrules.add_users(), sdk.alertrules.remove_users(), etc
+
+    2. Create an AlertRulesClient focusing only on current requirement.
```
unparalleled-js, line 50: I like this one better because I think it is similar to other clients in py42. A single client can use multiple endpoints if they are closely related. However, your first idea is pretty cool. Do we suspect there will be / could be other methods for the three different rule types? Do they have different IDs or is getting a rule by ID just get any rule type that has that ID?

Another idea might be combining some of the ideas behind other querying py42 does. For example, we can have a search method that takes a parameter that is either an object or a str that contains the endpoint str. That object or str could be one of several enum options.

For example

class CloudSharePermissionsRule(object):
  _endpoint = "query-cloud-share-permissions-rule"

  def __str__(): 
    return _endpoint

def search(rule_type):
  url = "{}/{}".format(base_uri, str(rule_type))
  request(url)
`

alanag13: I like the api for option 2 with the exception of `get`.

the thing I don't like about `get` is that if we add additional rule types, py42 will not automatically be searching them without an update, and that fact won't be obvious to the user.

I think I like @unparalleled-js 's idea here -- we can expose the types of rules as constants and require the type as a parameter to `get`, making it more clear that that is the only type of rule being searched.

kiran-chaudhary: ok, makes sense. 

Also analysing Web UI, `console.us.code42.com`, 'manage alert' tab it fetches all Rules with below endpoint
   https://alert-service-east.us.code42.com/svc/api/v1/rules/query-rule-metadata, as opposed to what is mentioned in swagger document, https://fed-observer-default.core-int.cloud.code42.com/svc/swagger/index.html#/. 

We can choose to use the metadata endpoint instead of individual ones, then we don't have to specify the rule_type, we need to make a call on the same.

Meanwhile, I will start work with option 2 for add and remove users.

kiran-chaudhary:
Queries: 

1. Should `tenant_id` be an argument to the methods?

2. For the task `Alert rules should be searchable by name` , there are no endpoints as such that provide searching by name, do we need to fetch all results and then compare name and return the response?

src/py42/clients/alertrules.py
alanag13: makes sense to me!

unparalleled-js: Sounds good!

unparalleled-js:
> Queries: @alanag13 @unparalleled-js
> 
> 1. Should `tenant_id` be an argument to the methods?
> 2. For the task `Alert rules should be searchable by name` , there are no endpoints as such that provide searching by name, do we need to fetch all results and then compare name and return the response?

1.) I would say we are leaning towards not making tenant ID a parameter until it is requested by customers / endusers.

2.) If there no way to do this with the endpoint alone, we can loop through the `get_all()` method until we find a match. We did that for v1 departing employee.

alanag13:
@kiran-chaudhary:
1. no, i feel that `tenant-id` should be determined for the user. We can revisit this and add it in if we ever come across a use case where specifying the tenant manually makes sense, but right now the api only uses this for internal service accounts that I would expect anyone to ever script against.

2. yes, if the api itself does not allow you to filter by name, that is the most reasonable way to accomplish this.

kiran-chaudhary:
Thank you @unparalleled-js @alanag13  

I have all the information I need, I will close this PR now.

PR #85, opened 2020-04-21
Update departingemployee.md

annie-payseur:
IX feedback per INTEG-979

### DIFF
--- a/docs/userguides/departingemployee.md
+++ b/docs/userguides/departingemployee.md
@@ -1,6 +1,8 @@
-# Add or Remove users from the Departing Employee List
+# Add or Remove Users From the Departing Employees List
 
-Let's say you want to add a user to the departing employee list. All you need to know is the user's Code42 user UID.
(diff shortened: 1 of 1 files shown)


docs/userguides/departingemployee.md
```diff
@@ -1,6 +1,8 @@
-# Add or Remove users from the Departing Employee List
+# Add or Remove Users From the Departing Employees List
```
unparalleled-js, line 1: Departing Employees List sounds like a list of lists to me

annie-payseur: hmm. "list" is the term we're using in IX documentation (instead of "lens"). it's a list of employees/users.

unparalleled-js: The "S" on Employees is what I was referring to. If this how normal talk, then I am ok with it.

unparalleled-js: An "employee list" is also a list of employees though. The s is implied by the "list", isn't it?

unparalleled-js: Consistency is good! Best to match the IX docs.

PR #86, opened 2020-04-21
Chore/ix feedback

timabrmsn:


docs/methoddocs/alerts.md
```diff
@@ -5,3 +5,60 @@
...
+
+```eval_rst
+.. autoclass:: py42.sdk.queries.alerts.alert_query.AlertQuery
+    :members:
+    :show-inheritance:
+```
+
+## Filter Classes
+
+The following classes construct filters for file event queries. Each filter class corresponds to a alert detail.
```
unparalleled-js, line 17: "a" should be "an", 3rd to last word on this line

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -112,12 +109,12 @@ def _get_high_risk_employees_page(
         return self._session.post(uri, data=json.dumps(data))
 
     def get_all(self, filter_type=u"OPEN", sort_key=None, sort_direction=None):
-        """Search High Risk employee list. Filter results by filter_type.
+        """Search High Risk Employee list. Filter results by filter_type.
```
unparalleled-js, line 112: `High Risk Employees` to be consistent, right?

src/py42/clients/devices.py
```diff
@@ -53,37 +53,31 @@ def get_all(
...
         active=true and blocked=true.
-
         `REST Documentation <https://console.us.code42.com/apidocviewer/#Computer-get>`__
 
         Args:
             active (bool, optional): Filters results by device state. When set to True, gets all
                 active devices. When set to False, gets all deactivated devices. When set to None
                 or excluded, gets all devices regardless of state. Defaults to None.
-            blocked (bool, optional): Filters results by blocked status, True or False. Defaults
+            blocked (bool, optional): Filters results by blocked status: True or False. Defaults
```
unparalleled-js, line 62: Do we need to escape this colon? I've seen weird behaviors with using colons like this when building the docs.

src/py42/clients/devices.py
```diff
@@ -102,13 +96,12 @@ def get_all(
...
-
         `REST Documentation <https://console.us.code42.com/apidocviewer/#Computer-get>`__
 
         Args:
-            device_id (int): The ID of the device.
-            include_backup_usage (bool, optional): A flag to denote whether to include its
-              backup summary. Defaults to None.
+            device_id (int): The identification number of the device.
+            include_backup_usage (bool, optional): A flag to denote whether to include the
+            destination and its backup stats. Defaults to None.
```
unparalleled-js, line 104: `Args` docs does permit indenting on new lines, makes it easier to read.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -6,11 +6,7 @@
 
 
 class HighRiskEmployeeClient(BaseClient):
-    """Administrator utility to manage High Risk employees.
-
-    `Support Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Detection_list_management_APIs>`__
-
-    """
+    """A client for interacting with High Risk Employee APIs."""
```
unparalleled-js, line 9: Employees? If it is plural for `departing employees`, it's  plural for HR as well probably

src/py42/clients/detectionlists/high_risk_employee.py
timabrmsn: @annie-payseur didn't call out the need for pluralization on HRE anywhere.

src/py42/clients/devices.py
```diff
@@ -119,13 +112,12 @@ def get_by_id(self, device_id, include_backup_usage=None, **kwargs):
...
         """Gets device information by GUID.
-
         `REST Documentation <https://console.us.code42.com/apidocviewer/#Computer-get>`__
 
         Args:
-            guid (str): The GUID of the device.
-            include_backup_usage (bool, optional): A flag to denote whether to include its
-              backup summary. Defaults to None.
+            guid (str): The globally unique identifier of the device.
+            include_backup_usage (bool, optional): A flag to denote whether to include the
```
unparalleled-js, line 119: Other places for args indenting too in this file

docs/methoddocs/alerts.md
timabrmsn: yep

src/py42/clients/devices.py
timabrmsn: Looks good for me in local build.

src/py42/modules/detectionlists.py
```diff
@@ -86,12 +87,13 @@ def add_user_risk_tags(self, user_id, tags):
         self._detection_list_user_client.add_risk_tags(user_id, tags)
 
     def remove_user_risk_tags(self, user_id, tags):
-        """Remove one or more tags.
+        """Remove one or more risk factors.
```
unparalleled-js, line 90: @alanag13 Is arguing for us to all these `tags` last I checked, believing that `factors` is a marketing term.

src/py42/modules/detectionlists.py
```diff
@@ -86,12 +87,13 @@ def add_user_risk_tags(self, user_id, tags):
...
 
         Args:
-            user_id (str or int): The user_id whose tag(s) needs to be removed.
-            tags (str or list of str ): A single tag or multiple tags in a list to be removed.
-               e.g u"tag1" or ["tag1", "tag2"], for python version 2.X, pass u"str" instead of "str"
+            user_id (str or int): The Code42 userUid whose tag(s) needs you want to remove.
```
unparalleled-js, line 93: Tags or factors, we need to pick and be consistent

src/py42/modules/detectionlists.py
```diff
@@ -102,12 +104,13 @@ def remove_user_risk_tags(self, user_id, tags):
         self._detection_list_user_client.remove_risk_tags(user_id, tags)
 
     def add_user_cloud_aliases(self, user_id, aliases):
-        """Add one or more cloud alias.
+        """Add one or more cloud aliases.
```
unparalleled-js, line 107: `alias`, should be singular. I believe @alanag13  is fixing this though.

src/py42/clients/devices.py
unparalleled-js: It's probably fine then! I just wanted to make sure.

src/py42/modules/securitydata.py
```diff
@@ -107,14 +108,14 @@ def get_all_user_security_events(
         min_timestamp=None,
         max_timestamp=None,
     ):
-        """Gets legacy security events for the user with the given UID.
+        """Gets file activity events for the user with the given UID.
```
unparalleled-js, line 111: As an end user, how do I know that this is FED and not AED? That is why the word "legacy" was there. I'm curious as to the feedback that prompted it's removal and/or alternative ideas for how we can convey the limitations of this method.

unparalleled-js: Does `file activity` say enough? Cloud activity is still file activity.

src/py42/clients/detectionlists/high_risk_employee.py
unparalleled-js: But if you look at other IX docs, https://duckduckgo.com/?q=high+risk+employees+code42&ia=web,
you will see it is plural.

annie-payseur: This is tricky - I pluralized HRE in the guide because that matches the console UI and therefore our IX docs, however, I realize it's singular in the detection list management APIs. Now I'm not sure which direction to go here.

src/py42/modules/securitydata.py
timabrmsn: @annie-payseur thoughts?

annie-payseur: End users don't know the terms FED and AED at all - those are internal terms/acronyms. "Legacy security events" doesn't tell them much, either. To your point on file activity & cloud activity, we could say "endpoint file activity" here?

src/py42/modules/detectionlists.py
annie-payseur: I know Alan is looking to use tags as a more generic term, but my thought was to tie it back to risk factors in the explanation here, since both the API and UI use "risk factors".

timabrmsn: Should we say "risk factor tags"?

src/py42/clients/detectionlists/high_risk_employee.py
unparalleled-js: As long as they are consistent, either works! Probably easier to make them singular since most places in the code and docs, they are already singular.

src/py42/modules/securitydata.py
timabrmsn: The header of the checkboxes for enabling FED in Console is "Endpoint Monitoring" so I think that's a good compromise.

src/py42/modules/detectionlists.py
annie-payseur: That seems like a good compromise to me.

alanag13:
@timabrmsn let me know when youve completed feedback, but we we don't need split hair over `Employee` vs `Employees`, etc. As long as the documentation conveys the general correct info these are finer decisions that we don't have to make right now

src/py42/modules/securitydata.py
alanag13: `legacy Endpoint Monitoring` works for me, I still want to indicate that this is the "old" way of doing things.

src/py42/modules/detectionlists.py
alanag13: documentation is more editable than function names -- im fine with docs containing whatever info will best be able to help the user figure out what's going on.

timabrmsn:
Ok I think it's ready.

PR #90, opened 2020-04-22
[WIP] Feature/user alert rules

kiran-chaudhary:
Implemented AlertRulesClient. PR created for early review.

Pending task: 
   Implement `get_by_name`

### DIFF
--- a/src/py42/clients/alertrules.py
+++ b/src/py42/clients/alertrules.py
@@ -1,0 +1,127 @@
+import json
+from py42.clients import BaseClient
+from py42.clients.util import get_all_pages
+
+
+_ALERT_RULE_ENDPOINTS = {
+    "exfiltration": u"query-endpoint-exfiltration-rule",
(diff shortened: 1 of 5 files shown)


src/py42/clients/alertrules.py
```diff
@@ -0,0 +1,127 @@
...
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        tenant_id = self._user_context.get_current_tenant_id()
+        data = {
+            u"tenantId": tenant_id,
+            u"ruleId": rule_id,
+            u"userList": [
+                {
+                    # not sure what should be passed here and its significance.
```
kiran-chaudhary, line 41: I need to confirm the behaviour and difference between the usages of both the fields, whether one of them is optional, whether they are mutually exclusive, etc. I will do some testing and play around a bit here, right now just added as is since the API expects it.

PR #92, opened 2020-04-22
Delete section on dates since they are no longer consistent

unparalleled-js:


docs/userguides/basics.md
```diff
@@ -79,36 +79,11 @@ print(cloud_usernames)
...
-import py42.sdk
-import py42.util
-from datetime import datetime, timedelta
-sdk = py42.sdk.from_local_account("https://console.us.code42.com", "my_username", "my_password")
-
-
-# Prints all the departing employee cases on or after two weeks
-
-# get a date in the future
-departing_date = datetime.utcnow() + timedelta(days=14)
```
unparalleled-js, line 98: departure date is supposed to a str any way.

An example using posix timestamps might be more relevant somewhere else that is not basics.md

PR #93, opened 2020-04-22
Bugfix/exceptions wrap responses

timabrmsn:


timabrmsn:
I only needed it for figuring out my own errors, nothing in the CLI would depend on this. So it can wait until the next release.

PR #96, opened 2020-04-22
make departure_date optional

alanag13:


src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -25,7 +25,7 @@ def add(self, user_id, departure_date):
         Args:
             user_id (str or int): The Code42 userUid of the user you want to add to the departing
                 employees list.
-            departure_date (date): Date in YYYY-MM-DD format. Date is treated as UTC.
+            departure_date (str, optional): Date in YYYY-MM-DD format. Date is treated as UTC.
```
unparalleled-js, line 28: Forgot to end with `Defaults to None.`

PR #97, opened 2020-04-24
Feature/user alert rules

kiran-chaudhary:
Implemented [AlertRulesClient](https://jira.corp.code42.com/browse/INTEG-990) . 

EDIT Changed design as per [last discussion](https://github.com/code42/py42/pull/97#issuecomment-620831910)


src/py42/_internal/client_factories.py
```diff
@@ -114,6 +117,20 @@ def get_detection_list_user_client(self, user_client):
             )
         return self._detection_list_user_client
 
+    def get_alert_rules_client(self):
+        if not self._alert_rules_client:
+            session = self._get_jwt_session(u"FedService-API_URL")
```
kiran-chaudhary, line 122: Need to know exact string in place of `u"FedService-API_URL"`.  Tried to find it through `_get_sts_base_url` but it didn't help.

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,161 @@
...
+            :class:`py42.response.Py42Response`
+        """
+        tenant_id = self._user_context.get_current_tenant_id()
+        data = {"tenantId": tenant_id, "ruleIds": [rule_id]}
+        uri = u"{0}{1}".format(self._api_prefix, _ALERT_RULE_ENDPOINTS[rule_type])
+        return self._session.post(uri, data=json.dumps(data))
+
+
+class AlertRulesManagerClient(BaseClient):
+    """A client to manage Alert Rules."""
```
kiran-chaudhary, line 100: Added both client in same file, because of lack of proper naming ideas.

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,161 @@
...
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        tenant_id = self._user_context.get_current_tenant_id()
+        data = {
+            u"tenantId": tenant_id,
+            u"ruleId": rule_id,
+            u"userList": [
+                {
+                    # not sure what should be passed here and its significance.
```
kiran-chaudhary, line 41: I need to confirm the behaviour and difference between the usages of both the fields, whether one of them is optional, whether they are mutually exclusive, etc. I will do some testing and play around a bit here, right now just added as is since the API expects it. 
This is pending as I need to figure out the key to connect to `fed-observer` instance

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,161 @@
...
+            user_ids (str): List of comma separated user ids who needs to be added to the criteria.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        tenant_id = self._user_context.get_current_tenant_id()
+        data = {u"tenantId": tenant_id, u"ruleId": rule_id, u"userIdList": [user_ids]}
+        uri = u"{0}{1}".format(self._api_prefix, u"remove-users")
+        return self._session.post(uri, data=json.dumps(data))
+        pass
```
unparalleled-js, line 64: Unneeded `pass`

src/py42/_internal/clients/alertrules.py
unparalleled-js: Do we need to have two different clients?

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,161 @@
...
+
+        Args:
+            rule_name (str): Name to search for, case insensitive search.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        rules = self.get_all()
+
+        for rule in rules:
```
unparalleled-js, line 156: See https://py42docs.code42.com/en/stable/userguides/basics.html#paging if you get stuck here

src/py42/_internal/clients/alertrules.py
kiran-chaudhary: As we need connection to 2 different hosts, we need this. I had raised a query yesterday in this context on Slack, as for different methods we need connection to different servers.

src/py42/_internal/clients/alertrules.py
kiran-chaudhary: Thank you, In the latest changes, I am sending a json response instead of Py42Response, is that fine?

src/py42/_internal/clients/alertrules.py
kiran-chaudhary: Removed

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+_ALERT_RULE_ENDPOINTS = {
+    u"exfiltration": u"query-endpoint-exfiltration-rule",
+    u"cloudshare": u"query-cloud-share-permissions-rule",
+    u"typemismatch": u"query-file-type-mismatch-rule",
+}
+
+
+class AlertRulesClient(BaseClient):
```
unparalleled-js, line 14: I am wondering if there are better names for these... Maybe this one should be called `AlertRuleUsersClient` and the other one should be called `AlertRulesClient` ?

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,157 @@
...
+            u"ruleMetadata",
+            tenant_id=tenant_id,
+            sort_key=sort_key,
+            sort_direction=sort_direction,
+        )
+
+    def get_by_name(self, rule_name):
+        """Fetch a rule by its name.
+
+            Raises Py42NotFoundError when no match is found.
```
unparalleled-js, line 143: :class:\`py42.exceptions.Py42NotFoundError`

unparalleled-js:
Could you add a changelog entry?

src/py42/_internal/clients/alertrules.py
unparalleled-js: I think the `get_by_name()` function / logc only needs to be in the Alerts Rules Module and not this internal client. I think it makes sense for a module to do this sort of logic using clients, which are more expected to return Py42Responses.

src/py42/_internal/clients/alertrules.py
unparalleled-js: Does this first client have any other endpoints besides ones for managing users? If not, I think we should make the renaming suggestions (or other suggestions you may have).

tests/_internal/clients/test_alertrules.py
```diff
@@ -0,0 +1,142 @@
...
+            and posted_data["groups"] == []
+            and posted_data["groupClause"] == "AND"
+            and posted_data["pgNum"] == 0
+            and posted_data["pgSize"] == 1000
+            and posted_data["srtKey"] == "key"
+            and posted_data["srtDirection"] == "ASC"
+        )
+
+    @pytest.mark.skip("TODO add assert statements")
+    def test_get_by_name_posts_expected_data(self, mocker, py42_response):
```
kiran-chaudhary, line 116: Need help to figure out how to test this. `get_all` returns a generator containing `py42Response`, so instead of generator I have set its return value to a list. However this doesn't work for me. 

I want to test that the code indeed fetches the record with given name.

tests/_internal/clients/test_alertrules.py
```diff
@@ -0,0 +1,142 @@
...
+        alert_rule_client = mocker.MagicMock(spec=AlertRulesManagerClient)
+        py42_response.text = TEST_RESPONSE
+        alert_rule_client.get_all.return_value = [py42_response]
+        alert_rule_client.get_by_name(u"TestName")
+
+    @pytest.mark.skip("TODO add assert statements")
+    def test_get_by_name_filters_correct_record(self, mocker, py42_response):
```
kiran-chaudhary, line 123: Need help to figure out how to test this. `get_all` returns a generator containing `py42Response`, so instead of generator I have set its return value to a list. However this doesn't work for me.

I want to test that the code indeed fetches the record with given name.

unparalleled-js: You can create a mock generator.

Create a local function inside this test that yields the response (or for a better test, multiple responses):

```
responses = [wrongResponseA, py42_response, wrongResponeB]
def get_all():
  for r in responses:
    yield r

alert_rule.get_all.side_effect = get_all

```

CHANGELOG.md
```diff
@@ -10,6 +10,16 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Added
+
+-  `sdk.alertrules` methods:
+    - `add_user()`
```
kiran-chaudhary, line 16: Method names needs to be rethought. 

Here `add_user` takes a custom identifier and multiple email aliases in the user alert rules. Hence `add_user` is not  appropriate.  `add_user_group` would be more appropriate. 
   i.e. add_user_group(rule_id, group_identifier, [email1, email2] 

Similarly for `remove_users` and `remove_all_users` , 

I was not happy with any of the names I could think of and hence have left it for discussion in review.

tests/_internal/clients/test_alertrules.py
alanag13: @unparalleled-js 's suggestion should work, but im still confused about why @kiran-chaudhary 's current setup _doesnt_ work. What happens exactly? Is a generator still what is returned?

alanag13:
@unparalleled-js @kiran-chaudhary what do you think about accessing these methods as `sdk.alerts.rules` instead of `sdk.alertrules`? This would essentially involve:
 - Renaming `AlertRulesModule` to `AlertModule`
- passing `AlertClient` into `AlertModule`
- exposing all of `AlertClient`'s methods on `AlertModule`
- expose `AlertRulesClient` as a `rules` property.

tests/_internal/clients/test_alertrules.py
kiran-chaudhary: The function doesn't return the intended dictionary/json (`TEST_RESPONSE`) and hence can't assert the result.

CHANGELOG.md
kiran-chaudhary: Moreover there is one more endpoint in the API, which accepts rule_id, user_group_id and aliases, exactly like add_users and removes only the aliases from the rule. 

We might want to support that too, do we need a new method for that !?

kiran-chaudhary:
> @unparalleled-js @kiran-chaudhary what do you think about accessing these methods as `sdk.alerts.rules` instead of `sdk.alertrules`? This would essentially involve:
> 
> * Renaming `AlertRulesModule` to `AlertModule`
> * passing `AlertClient` into `AlertModule`
> * exposing all of `AlertClient`'s methods on `AlertModule`
> * expose `AlertRulesClient` as a `rules` property.

Based on [swagger doc](https://fed-observer-default.core-int.cloud.code42.com/svc/swagger/index.html#/Rules/Rules_CreateEndpointExfiltrationSystemRule) there are 3 types of rules so at some point if we need creation of these rules, we might need something like

`sdk.alerts.rules.exfiltration`,
`sdk.alerts.rules.cloudshare`

Then alike `User-Criteria` ,  current work, there are multiple criteria's we might need to support 
so that would make it, 

`sdk.alerts.rules.users`
`sdk.alerts.rules.[criteria2]`. e.g `fileActivity` and so on.

Or may be we want to make it...

`sdk.alerts.rules.exfiltration.[criteria1]`
`sdk.alerts.rules.exfiltration.[criteria2]` and so on

kiran-chaudhary:
> Could you add a changelog entry?

Added.

src/py42/_internal/clients/alertrules.py
```diff
@@ -0,0 +1,156 @@
...
+        Returns
+            :dict: Dictionary containing rule-details.
+        """
+        rule_pages = self.get_all()
+        for rule_page in rule_pages:
+            rules = rule_page["ruleMetadata"]
+            for rule in rules:
+                if rule_name.lower() in rule["name"].lower():
+                    return rule
+        raise Py42NotFoundError("No Alert Rules found with name: {0}".format(rule_name))
```
alanag13, line 156: I think creating this Exception will fail on the latest version of py42 -- we updated all of the HTTP-based error to take the root exception in the constructor instead of a message. We can raise a generic `Py42Error` here.

CHANGELOG.md
alanag13: `add_user` refers to a single code42 user who could have have multiple email addresses used by other services. The `group_identifier` you mention is the code42 username, while `[email1, email2]` is a set of known _alternate_ email addresses that the user has that otherwise may be unknown to code42.

I _think_ that this list in practice should be the same as the list of cloud aliases from `detectionlists.users` -- I will verify this. If this is the case, we should change our implementation to simply look those up and not require that the user specifiy them. I will do some research into this and get back to you here with details.

alanag13: I did some research into this today and learned a lot -- a lot of the api doesn't work like it should appear to on the surface. Apologies for the confusion, as this also was not obvious to me.

It turns out that the are two different kinds of rules: system rules and "normal" rules.
_system_ rules are automatically created by the system, one for each detection list type / rule type combination. In the UI, these are labeled with a blue oval with the detection list name inside them.

The following api methods are only relevant to _system_ rules, and are designed to only be called by Code42-owner service accounts:
- add-users
- remove-users
- remove-all-users
- remove-user-aliases

Because users are not typically allowed to modify system rules, modifying them will not be part of this feature.

Unfortunately, the interface for modifying "normal" (user-defined) rules is all contained within a single `POST` for each api type:
- /api/v1/Rules/update-cloud-share-permissions-rule
- /api/v1/Rules/update-endpoint-exfiltration-rule
- /api/v1/Rules/update-file-type-mismatch-rule

Adding / removing users from a rule involves setting the exact desired list of users on the "userList" property in the post body: This example uses `cloud-share-permissions` rules but is relevant to all of them. You can observe this behavior by creating your own rule in the console and then modifying it.

```json
{
  "rule": {
    "tenantId": "string",
    "name": "string",
    "description": "string",
    "severity": 0,
    "isEnabled": true,
    "fileBelongsTo": {
      "usersToAlertOn": 0,
      "userList": [ // <--- this list needs to be modified
        "string"
      ]
    },
    "notificationConfig": {
      "enabled": true,
      "notificationInfo": [
        {
          "notificationType": 0,
          "notificationAddress": "string"
        }
      ]
    },
    "fileCategoryWatch": {
      "watchAllFiles": true,
      "fileCategoryList": [
        0
      ]
    },
    "ruleSource": "string",
    "watchGoogleDrive": {
      "publicOnTheWeb": true,
      "publicViaLink": true,
      "outsideTrustedDomains": true
    },
    "watchMicrosoftOneDrive": {
      "publicViaLink": true,
      "outsideTrustedDomains": true
    },
    "watchBox": {
      "publicViaLink": true,
      "outsideTrustedDomains": true
    },
    "id": "string"
  }
```

essentially, our `add_user` method will need to first call the corresponding `query` method for the correct rule type, then modify the "userList" on it, then `POST` the entire body. Fortunately, the response from `query` is mostly similar to what `update` expects, except for some metadata near the bottom:

```json
    {
      "tenantId": "string",
      "name": "string",
      "description": "string",
      "severity": 0,
      "isEnabled": true,
      "fileBelongsTo": {
        "usersToAlertOn": 0,
        "userList": [
          "string"
        ]
      },
      "notificationConfig": {
        "enabled": true,
        "notificationInfo": [
          {
            "notificationType": 0,
            "notificationAddress": "string"
          }
        ]
      },
      "fileCategoryWatch": {
        "watchAllFiles": true,
        "fileCategoryList": [
          0
        ]
      },
      "ruleSource": "string",
      "watchGoogleDrive": {
        "publicOnTheWeb": true,
        "publicViaLink": true,
        "outsideTrustedDomains": true
      },
      "watchMicrosoftOneDrive": {
        "publicViaLink": true,
        "outsideTrustedDomains": true
      },
      "watchBox": {
        "publicViaLink": true,
        "outsideTrustedDomains": true
      },
      "id": "string",
      // --- everything below this line should be removed before posting to update
      "createdAt": "2020-04-27T22:18:48.007Z",
      "createdBy": "string",
      "modifiedAt": "2020-04-27T22:18:48.007Z",
      "modifiedBy": "string",
      "isSystem": true
    }
```

alanag13:
@kiran-chaudhary see my comment above for some clarification on to implement add/remove user. I apoligize for not clarifying this earlier -- I am unfamiliar with this api and this was also news to me.  https://github.com/code42/py42/pull/97#discussion_r416193133

> Based on [swagger doc](https://fed-observer-default.core-int.cloud.code42.com/svc/swagger/index.html#/Rules/Rules_CreateEndpointExfiltrationSystemRule) there are 3 types of rules so at some point if we need creation of these rules, or adding/updating/removing other criteria we might need something like
> 
> `sdk.alerts.rules.exfiltration`,
> `sdk.alerts.rules.cloudshare`
> 
> Then alike `User-Criteria` , current work, there are multiple criteria's we might need to support
> so that would make it,
> 
> `sdk.alerts.rules.users`
> `sdk.alerts.rules.[criteria2]`. e.g `fileActivity` and so on.
> 
> Or may be we want to make it...
> 
> `sdk.alerts.rules.exfiltration.[criteria1]`
> `sdk.alerts.rules.exfiltration.[criteria2]` and so on

Based on what I learned above, we will want the api to be something like
sdk.alerts.rules.add_users(rule_id, list_users) 
sdk.alerts.rules.remove_users(rule_id, list_users)
sdk.alerts.rules.get(rule_id)
sdk.alerts.rules.get_all <--- as is

`get`, `remove_users`, and `add_users` would implemented as follows:
- call `get_all`
- find matching rule_id in collection returned by `get_all`
- inspect the `type` property of that rule. it will be one of the following:
    - `FED_ENDPOINT_EXFILTRATION`
    - `FED_CLOUDSHARE_PERMISSIONS`
    - `FED_FILE_TYPE_MISMATCH`
- based on the type, call the appropriate `query` and `update` commands.

Let me know if you have any questions. Thanks!

kiran-chaudhary:
> @kiran-chaudhary see my comment above for some clarification on to implement add/remove user. I apoligize for not clarifying this earlier -- I am unfamiliar with this api and this was also news to me. [#97 (comment)](https://github.com/code42/py42/pull/97#discussion_r416193133)
> 
> > Based on [swagger doc](https://fed-observer-default.core-int.cloud.code42.com/svc/swagger/index.html#/Rules/Rules_CreateEndpointExfiltrationSystemRule) there are 3 types of rules so at some point if we need creation of these rules, or adding/updating/removing other criteria we might need something like
> > `sdk.alerts.rules.exfiltration`,
> > `sdk.alerts.rules.cloudshare`
> > Then alike `User-Criteria` , current work, there are multiple criteria's we might need to support
> > so that would make it,
> > `sdk.alerts.rules.users`
> > `sdk.alerts.rules.[criteria2]`. e.g `fileActivity` and so on.
> > Or may be we want to make it...
> > `sdk.alerts.rules.exfiltration.[criteria1]`
> > `sdk.alerts.rules.exfiltration.[criteria2]` and so on
> 
> Based on what I learned above, we will want the api to be something like
> sdk.alerts.rules.add_users(rule_id, list_users)
> sdk.alerts.rules.remove_users(rule_id, list_users)
> sdk.alerts.rules.get(rule_id)
> sdk.alerts.rules.get_all <--- as is
> 
> `get`, `remove_users`, and `add_users` would be implemented as follows:
> 
> * call `get_all`
> * find matching rule_id in collection returned by `get_all`
> * inspect the `type` property of that rule. it will be one of the following:
>   
>   * `FED_ENDPOINT_EXFILTRATION`
>   * `FED_CLOUDSHARE_PERMISSIONS`
>   * `FED_FILE_TYPE_MISMATCH`
> * based on the type, call the appropriate `query` and `update` commands.
> 
> Based on everything we've learned, I don't feel the "dynamic" approach will be very safe -- it assumes too much about what future urls, etc will be named. if we add a new type of rule, we will have to update our library to accommodate it.
> 
> Let me know if you have any questions. Thanks!

@alanag13 
Even I had the same impression while analysing the API first time, that we need whole response to make an update request, I raised it [here on slack](https://code42sw.slack.com/archives/G010NGJCSGL/p1587376365267700) and in the thread @mikemccollowc42  pointed us to the correct endpoints. Those endpoints are not specific to System rules only, they do work for normal rules as well. 

I can explain the same with current py42 implementation.

`>>> rules_client = sdk.alertrules
`
Now  I intend to update this rule 
`{'type$': 'RULE_METADATA', 'modifiedBy': 'kiran.chaudhary+partner@code42.com', 'modifiedAt': '2020-04-24T16:49:42.9767920Z', 'name': 'Explore cloud user rule alerts', 'description': '', 'severity': 'LOW', 'isSystem': False, 'isEnabled': True, 'ruleSource': 'Alerting', 'tenantId': '1d71796f-af5b-4231-9d8e-df6434da4663', 'observerRuleId': 'e57ae2e7-b1f8-4332-92e7-1ff4d47bd951', 'type': 'FED_CLOUD_SHARE_PERMISSIONS', 'id': '6930a1fe-73fc-4f89-8c45-96cba5547b71', 'createdBy': 'kiran.chaudhary+partner@code42.com', 'createdAt': '2020-04-20T06:51:10.9420250Z'}`

`>>> rules_client.add_user('e57ae2e7-b1f8-4332-92e7-1ff4d47bd951', "RANDOM-KEY", ['kiran.chaudhary+partner@code42.com', 'alan.grgic+schume@code42.com', 'mike.mccollow+partner@code42.com'])
<Py42Response [status=200, data='']>`

Check whether rule is updated.

`>>> updated_rule = rules_client.get('cloudshare', 'e57ae2e7-b1f8-4332-92e7-1ff4d47bd951')
>>> updated_rule.text
'{"type$": "CLOUD_SHARE_PERMISSIONS_RULE_DETAILS_RESPONSE", "rules": [{"type$": "CLOUD_SHARE_PERMISSIONS_RULE_DETAILS", "tenantId": "1d71796f-af5b-4231-9d8e-df6434da4663", "name": "Explore cloud user rule alerts", "description": "", "severity": "LOW", "isEnabled": true, "fileBelongsTo": {"type$": "FILE_BELONGS_TO", "usersToAlertOn": "SPECIFIED_USERS", "userList": ["kiran.chaudhary+partner@code42.com", "alan.grgic+schume@code42.com", "mike.mccollow+partner@code42.com"]}, "notificationConfig": {"type$": "NOTIFICATION_CONFIG", "enabled": false}, "fileCategoryWatch": {"type$": "FILE_CATEGORY_WATCH", "watchAllFiles": true}, "ruleSource": "Alerting", "watchGoogleDrive": {"type$": "WATCH_GOOGLE_DRIVE", "publicOnTheWeb": true, "publicViaLink": true, "outsideTrustedDomains": true}, "watchMicrosoftOneDrive": {"type$": "WATCH_MICROSOFT_ONE_DRIVE", "publicViaLink": false, "outsideTrustedDomains": false}, "watchBox": {"type$": "WATCH_BOX", "publicViaLink": false, "outsideTrustedDomains": false}, "id": "e57ae2e7-b1f8-4332-92e7-1ff4d47bd951", "createdAt": "2020-04-20T06:51:12.0585280Z", "createdBy": "kiran.chaudhary+partner@code42.com", "modifiedAt": "2020-04-28T06:43:05.9087800Z", "modifiedBy": "kiran.chaudhary+partner@code42.com", "isSystem": false}]}'`


So we might not need the approach of `get_all`, the current implemented endpoints should suffice.  However as you advised we need the design changes.

kiran-chaudhary:
> @unparalleled-js @kiran-chaudhary what do you think about accessing these methods as `sdk.alerts.rules` instead of `sdk.alertrules`? This would essentially involve:
> 
> * Renaming `AlertRulesModule` to `AlertModule`
> * passing `AlertClient` into `AlertModule`
> * exposing all of `AlertClient`'s methods on `AlertModule`
> * expose `AlertRulesClient` as a `rules` property.

In design context this is better, but in our current implementation there are two clients (based on hosts ), so are you ok with current naming convention used. 

Apart from above point, whether this would suffice in future ? i.e as we see that there are two ways to update a rule:

-  one where we pass the whole response with the updated criteria.  

- second where we have new endpoints for add_users, remove_users, etc...

Now if similar changes would be done for other criteria in API and if we need to support adding/updating those criteria to the rule, will this design hold ?

I am sorry if I am not able to convey the point and adding confusion, may be we can have a call if that is convenient.

src/py42/_internal/clients/alertrules.py
kiran-chaudhary: Changed to Py42Error with the message.

alanag13:
@kiran-chaudhary okay, so after our call and after asking the team for additional clarifications, let's start from beginning and define the design:

```
sdk.alerts.rules.add_user(rule_id, user_id)
sdk.alerts.rules.remove_user(rule_id, user_id)
sdk.alerts.rules.remove_all_users(rule_id)
sdk.alerts.rules.get_all <--- as is
```

you`ll notice that I have `user` as singular -- this is because there will not be a good way of tracking failures to a single item in a large list. If `add_user` or `remove_user` fails, we need to know which user caused the failure.

secondly, aliases have been removed, this is because as we discussed, these are intended to be the cloud aliases for `DetectionListUserClient` -- so the implementation will need to look up the user's aliases and send them all.

```
sdk.alerts.rules.exfiltration.get(rule_id)
sdk.alerts.rules.cloudshare.get(rule_id)
sdk.alerts.rules.filetypemismatch.get(rule_id)
```
I changed my mind about these because i don't want a generic `get` method that returns different JSON schemas. The user will need to figure out which one to call based on the `type` field that you get back from `get_all`, which you are going to have to call in order to get the rule_id, anyway. I also like that the `get` and any future related `update` functions for each will be in the same place.

src/py42/_internal/clients/alertrules/exfiltration.py
```diff
@@ -0,0 +1,26 @@
...
+from py42.clients import BaseClient
+
+
+class Exfiltration(BaseClient):
+
+    _version = u"v1"
+    _resource = u"query-endpoint-exfiltration-rule"
+    _api_prefix = u"/svc/api/{0}/Rules/{1}".format(_version, _resource)
+
+    def __init__(self, session, tenant_id):
```
kiran-chaudhary, line 12: Instead of passing `user_context`, passed `tenant_id` , a little optimisation since `get_current_tenant_id`  is the only method in `UserContext`, I felt passing it is unnecessary and it makes a network call each time it is called. Is this ok?

Did this change only in few places considering may be I have to revert it back if this is incorrect.

tests/_internal/clients/test_alertrules.py
kiran-chaudhary: Resolving this as fixed. The response sample used had some errors due to which json to dictionary conversion was failing, changed the response to bare minimum to fix.

src/py42/_internal/clients/alertrules/exfiltration.py
alanag13: makes sense to me!

src/py42/clients/alerts.py
```diff
@@ -103,3 +105,56 @@ def _add_tenant_id_if_missing(self, query):
...
+            "groupClause": "AND",
+            "pgNum": 0,  # This API expects first page to start with zero, hence hard-coded.
+            "pgSize": page_size,
+            "srtKey": sort_key,
+            "srtDirection": sort_direction,
+        }
+        uri = self._uri_prefix.format(u"rules/query-rule-metadata")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def get_all(self, sort_key="CreatedAt", sort_direction="DESC"):
```
unparalleled-js, line 124: Let's name this `get_all_rules()` so that we know it returns rules and not alerts, since it is now in the `AlertsClient`

src/py42/_internal/client_factories.py
```diff
@@ -59,6 +60,8 @@ def __init__(
         self._high_risk_employee_client = None
         self._detection_list_user_client = None
         self._ecm_session = None
+        self._alert_rules_client = None
+        self._alert_rules_manager_client = None
```
unparalleled-js, line 64: Is this still a thing?

src/py42/_internal/clients/alertrules/file_type_mismatch.py
```diff
@@ -0,0 +1,26 @@
+import json
+
+from py42.clients import BaseClient
+
+
+class FileTypeMisMatch(BaseClient):
```
unparalleled-js, line 6: Should we name these ending with the word `client`, as in our convention?
It does not matter too much since they are internal

src/py42/_internal/clients/alert_rules.py
```diff
@@ -0,0 +1,93 @@
...
+        }
+        uri = u"{0}{1}".format(self._api_prefix, u"add-users")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def remove_user(self, rule_id, user_id):
+        """Update alert rule criteria to remove all aliases specified against the user-uid.
+
+        Args:
+            rule_id (str): Observer rule Id of a rule to be updated.
+            user_id (str): A valid user Uid.
```
alanag13, line 70: Let keep this description consistent with other places we describe the user id, something like:
```
The Code42 userUid  of the user to remove from the alert
```

src/py42/_internal/client_factories.py
kiran-chaudhary: Removed.

src/py42/_internal/clients/alertrules/file_type_mismatch.py
kiran-chaudhary: Renamed.

src/py42/sdk/__init__.py
```diff
@@ -147,3 +147,12 @@ def detectionlists(self):
             :class:`py42.modules.detectionlists.DetectionListsModule`
         """
         return self._sdk_dependencies.detection_lists_module
+
+    @property
+    def alerts(self):
+        """A collection of methods related to retrieving, updating alerts rules.
+
+        Returns:
+            :class:`py42.clients.AlertRules
```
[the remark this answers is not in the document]
kiran-chaudhary: Added.

src/py42/sdk/__init__.py
```diff
@@ -147,3 +147,12 @@ def detectionlists(self):
             :class:`py42.modules.detectionlists.DetectionListsModule`
         """
         return self._sdk_dependencies.detection_lists_module
+
+    @property
+    def alerts(self):
+        """A collection of methods related to retrieving, updating alerts rules.
```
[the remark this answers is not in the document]
kiran-chaudhary: Changed.

src/py42/_internal/clients/alert_rules.py
```diff
@@ -0,0 +1,93 @@
...
+        Args:
+            rule_id (str): Observer Id of a rule to be updated.
+            user_id (str): A custom identifier to uniquely identify the specified user aliases
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        tenant_id = self._user_context.get_current_tenant_id()
+        user_details = self._detection_list_user_client.get_by_id(user_id)
+        print(user_details)
```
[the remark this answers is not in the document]
kiran-chaudhary: Removed.

src/py42/_internal/clients/alert_rules.py
```diff
@@ -0,0 +1,93 @@
...
+        data = {
+            u"tenantId": tenant_id,
+            u"ruleId": rule_id,
+            u"userList": [{u"userIdFromAuthority": user_id, u"userAliasList": user_aliases,}],
+        }
+        uri = u"{0}{1}".format(self._api_prefix, u"add-users")
+        return self._session.post(uri, data=json.dumps(data))
+
+    def remove_user(self, rule_id, user_id):
+        """Update alert rule criteria to remove all aliases specified against the user-uid.
```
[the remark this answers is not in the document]
kiran-chaudhary: Changed

src/py42/clients/alerts.py
```diff
@@ -103,3 +105,56 @@ def _add_tenant_id_if_missing(self, query):
...
             return str(query)
+
+    def _get_alert_rules(
+        self, tenant_id, sort_key=None, sort_direction=None, page_num=None, page_size=None
+    ):
+        data = {
+            "tenantId": tenant_id,
+            "groups": [],
+            "groupClause": "AND",
+            "pgNum": 0,  # This API expects first page to start with zero, hence hard-coded.
```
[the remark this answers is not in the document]
kiran-chaudhary: Updated.

src/py42/clients/alerts.py
```diff
@@ -103,3 +105,56 @@ def _add_tenant_id_if_missing(self, query):
...
+        tenant_id = self._user_context.get_current_tenant_id()
+        return get_all_pages(
+            self._get_alert_rules,
+            u"ruleMetadata",
+            tenant_id=tenant_id,
+            sort_key=sort_key,
+            sort_direction=sort_direction,
+        )
+
+    def get_by_name(self, rule_name):
```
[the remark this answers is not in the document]
kiran-chaudhary: Changed.

src/py42/modules/alertrules.py
```diff
@@ -0,0 +1,38 @@
...
+        return self.microservice_client_factory.get_alert_rules_client(
+            self.microservice_client_factory.get_detection_list_user_client(self._user_client)
+        )
+
+    def get_all(self, sort_key="CreatedAt", sort_direction="DESC"):
+        """Fetch all available rules.
+
+        Args:
+            sort_key (str): Sort results based by field, default 'CreatedAt'.
+            sort_direction (str): ``ASC`` or ``DESC``, default "DESC"
```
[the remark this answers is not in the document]
kiran-chaudhary: Changed.

src/py42/_internal/clients/alert_rules.py
kiran-chaudhary: Changed.

src/py42/modules/alertrules.py
```diff
@@ -0,0 +1,38 @@
...
+        self._user_client = user_client
+        self._alert_client = None
+
+    @property
+    def rules(self):
+        return self.microservice_client_factory.get_alert_rules_client(
+            self.microservice_client_factory.get_detection_list_user_client(self._user_client)
+        )
+
+    def get_all(self, sort_key="CreatedAt", sort_direction="DESC"):
```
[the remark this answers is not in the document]
kiran-chaudhary: If  `AlertRulesModule` contains `AlertClient`, it will allow excess to other methods  as well, that existed earlier. Is that the expectation? 
 Do we want to support access to only `get_all` and `get_rules_by_name` or all?

src/py42/clients/alerts.py
kiran-chaudhary: Method names changed.

src/py42/clients/alerts.py
```diff
@@ -103,3 +104,56 @@ def _add_tenant_id_if_missing(self, query):
             return json.dumps(query_dict)
         else:
             return str(query)
+
+    def _get_alert_rules(
+        self, tenant_id, sort_key=None, sort_direction=None, page_num=None, page_size=None
+    ):
+        data = {
```
unparalleled-js, line 111: Missing `u""`s here

src/py42/modules/alertrules.py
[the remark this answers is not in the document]
unparalleled-js: I think the idea is that they are private, so you only get access to the methods off that client that you need.

src/py42/modules/alertrules.py
```diff
@@ -0,0 +1,37 @@
+class AlertRulesModule(object):
+    def __init__(self, microservice_client_factory, user_client):
+        self.microservice_client_factory = microservice_client_factory
+        self._user_client = user_client
+        self._alert_client = None
+
+    @property
+    def rules(self):
```
unparalleled-js, line 8: Is this supposed be property or just private? `sdk.alertrules.rules` sounds unecessary.

tests/_internal/clients/alertrules/test_file_type_mismatch.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+import json
+
+from py42._internal.clients.alertrules.file_type_mismatch import FileTypeMismatchClient
+
+
+class TestFileTypeMisMatch(object):
```
unparalleled-js, line 7: This test classes should also end in `Client`

alanag13:
@kiran-chaudhary i got this updated to use the structure that I was trying to articulate. Could you finish this up by adding tests for `AlertRulesModule` and `AlertsModule` classes that I created, and also change `docs` to use those classes instead of `AlertClient`?

src/py42/_internal/client_factories.py
```diff
@@ -2,10 +2,11 @@
 
 from requests import HTTPError
 
-from py42._internal.clients import archive
+from py42._internal.clients import archive, alerts
 from py42._internal.clients import key_value_store
```
unparalleled-js, line 6: These three import statement can be combined

src/py42/_internal/initialization.py
```diff
@@ -4,6 +4,7 @@
 from py42._internal.storage_session_manager import StorageSessionManager
 from py42._internal.token_providers import StorageTokenProviderFactory
 from py42.modules import archive as archive_module, detectionlists, securitydata as sec_module
+from py42.modules import alertrules, alerts
```
unparalleled-js, line 7: This can be combined with the import above it

src/py42/clients/alertrules/exfiltration.py
```diff
@@ -0,0 +1,26 @@
...
+    def get(self, rule_id):
+        """Fetch exfiltration alert rule by rule id.
+
+        Args:
+            rule_id (str): Observer rule Id of a rule to be fetched.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        data = {"tenantId": self._tenant_id, "ruleIds": [rule_id]}
```
unparalleled-js, line 25: Missing `u""s here

src/py42/clients/alertrules/cloud_share.py
```diff
@@ -0,0 +1,26 @@
...
+    def get(self, rule_id):
+        """Fetch cloud share alert rule by rule id.
+
+        Args:
+            rule_id (str): Observer rule Id of a rule to be fetched.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        data = {"tenantId": self._tenant_id, "ruleIds": [rule_id]}
```
unparalleled-js, line 25: The three specific rule clients all seem really similar and could share code. Do we want to make a base `RuleClient` that handles creating the data and making the request and takes a URL?

alanag13: this is the case right now, but the type of data required to update each type of rule is very different. I'd rather keep them separate.

src/py42/clients/alertrules/exfiltration.py
unparalleled-js: The three muska-rule-type clients all have this here

src/py42/clients/alertrules/cloud_share.py
kiran-chaudhary: This idea of similar code just allows me to share a thought I was having, all we do as business logic here is

-  map the method parameters to request fields of the API, 

- define uri

- define http method 

so even that can be abstracted out to a configuration file and a function,  this would reduce the verbosity, and only focus would remain to design the feature layout.

Obviously this would be a wrapper around a wrapper and as it adds code complexity, not convinced enough to prefer such approach.

src/py42/clients/alertrules/exfiltration.py
kiran-chaudhary: I am sorry, this u"" thing is coming up again and again and I feel like I have fixed all, I wonder if there was command which could let us know if all string are python 2 compatible, I would add it to `pre-commit` . Fixed.

src/py42/clients/alertrules/cloud_share.py
alanag13: that's an interesting thought for py42 as a whole! probably out of scope for this PR though

src/py42/_internal/initialization.py
```diff
@@ -40,7 +45,7 @@ def __init__(self, host_address, session_factory, root_session):
         )
 
         microservice_client_factory = MicroserviceClientFactory(
-            host_address, root_session, session_factory, self.user_context
+            host_address, root_session, session_factory, self.user_client, self.user_context
```
unparalleled-js, line 48: Passing user_context and user_client in the wrong order.

Maybe we should be passing in key value arguments here

src/py42/modules/alerts.py
```diff
@@ -29,9 +29,8 @@ def search(self, query):
             :class:`py42.response.Py42Response`: A response containing the alerts that match the given
             query.
         """
-        query = self._add_tenant_id_if_missing(query)
-        uri = self._uri_prefix.format(u"query-alerts")
-        return self._session.post(uri, data=query)
+        alert_client = self._microservice_client_factory.get_alerts_client()
+        return alert_client.search(query)
 
     def get_details(self, alert_ids, tenant_id=None):
```
unparalleled-js, line 35: Why do we no longer accept single Id args? This is a breaking change that I think is unnecessary

unparalleled-js: Oh wait it is probably in the client

unparalleled-js: In that case, the doc strings should indicate that. `str or list of str`

src/py42/sdk/__init__.py
```diff
@@ -147,3 +147,12 @@ def detectionlists(self):
             :class:`py42.modules.detectionlists.DetectionListsModule`
         """
         return self._sdk_dependencies.detection_lists_module
+
+    @property
+    def alerts(self):
+        """A collection of methods related to retrieving and updating alerts rules.
+
+        Returns:
+            :class:`py42.clients.AlertRules`
```
unparalleled-js, line 156: This is not the correct type listed

src/py42/sdk/__init__.py
```diff
@@ -147,3 +147,12 @@ def detectionlists(self):
             :class:`py42.modules.detectionlists.DetectionListsModule`
         """
         return self._sdk_dependencies.detection_lists_module
+
+    @property
+    def alerts(self):
```
unparalleled-js, line 152: Is it ok that we still have `alerts` off of `securitydata` as well?

src/py42/clients/alertrules/__init__.py
```diff
@@ -0,0 +1,66 @@
+import json
+
+from py42.clients import BaseClient
+from py42.clients.alertrules.cloud_share import CloudShareClient
+from py42.clients.alertrules.exfiltration import ExfiltrationClient
+from py42.clients.alertrules.file_type_mismatch import FileTypeMismatchClient
+
+
+class AlertRulesClient(BaseClient):
```
unparalleled-js, line 9: This is a public-facing client without docs

PR #98, opened 2020-04-24
added update_cold_storage_purge_date to archive client

ceciliastevens:
This request adds an update_cold_storage_purge_date to the archive client. This functionality is used by PS in our deactivate devices script, which we would like to re-implement using py42.

I'm pretty new to all of this and I'm not sure I did things right, so please let me know if I should make any changes, or if there's better ways to do things.

### DIFF
--- a/src/py42/modules/archive.py
+++ b/src/py42/modules/archive.py
@@ -109,13 +109,13 @@ class ArchiveModule(object):
 
     def update_cold_storage_purge_date(self, archive_guid, purge_date):
         """Updates the cold storage purge date for a specified archive.
-        REST Documentation <https://www.crashplan.com/apidocviewer/#ColdStorage-put>
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#ColdStorage-put>`__
 
         Args:
             archive_guid (str): The identification number of the archive that should be updated
             purge_date (str): The date on which the archive should be purged in yyyy-MM-dd format
 
         Returns:
-            :class: py42.response.Py42Response: the response from the ColdStorage API.
+            :class: `py42.response.Py42Response`: the response from the ColdStorage API.
         """
         return self._archive_client.update_cold_storage_purge_date(archive_guid, purge_date)
--- a/src/py42/modules/archive.py
+++ b/src/py42/modules/archive.py
@@ -106,3 +106,16 @@ class ArchiveModule(object):
             that each contain a page of restore history.
         """
         return self._archive_client.get_all_restore_history(days, u"computerId", device_id)
+
+    def update_cold_storage_purge_date(self, archive_guid, purge_date):
+        """Updates the cold storage purge date for a specified archive.
+        REST Documentation <https://www.crashplan.com/apidocviewer/#ColdStorage-put>
+
+        Args:
+            archive_guid (str): The identification number of the archive that should be updated
+            purge_date (str): The date on which the archive should be purged in yyyy-MM-dd format
+
+        Returns:
+            :class: py42.response.Py42Response: the response from the ColdStorage API.
+        """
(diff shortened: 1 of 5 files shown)


src/py42/modules/archive.py
```diff
@@ -106,3 +106,16 @@ def get_all_device_restore_history(self, days, device_id):
             that each contain a page of restore history.
         """
         return self._archive_client.get_all_restore_history(days, u"computerId", device_id)
+
+    def update_cold_storage_purge_date(self, archive_guid, purge_date):
+        """Updates the cold storage purge date for a specified archive.
+        REST Documentation <https://www.crashplan.com/apidocviewer/#ColdStorage-put>
```
unparalleled-js, line 112: This doc link is a little off. It should all be surrounded in tick marks and should trail with two underscores, like this:

``REST Documentation <https://www.crashplan.com/apidocviewer/#ColdStorage-put>``__

unparalleled-js: Also, for consistency's sake, can you link to `https://console.us.code42.com/apidocviewer/#ColdStorage-put` instead of crashplan?

src/py42/modules/archive.py
```diff
@@ -106,3 +106,16 @@ def get_all_device_restore_history(self, days, device_id):
...
+
+        Args:
+            archive_guid (str): The identification number of the archive that should be updated
+            purge_date (str): The date on which the archive should be purged in yyyy-MM-dd format
+
+        Returns:
+            :class: py42.response.Py42Response: the response from the ColdStorage API.
```
unparalleled-js, line 119: Similar thing here -- to get the docs links to build, I believe you have to use tick marks like this: :class:\`py42.response.Py42Response`

unparalleled-js:
Thanks for doing this! If you don't mind, could you add an entry in the ChangeLog that described these added methods (to the module only, not to the client since it's internal).

src/py42/modules/archive.py
```diff
@@ -106,3 +106,16 @@ def get_all_device_restore_history(self, days, device_id):
...
         """
         return self._archive_client.get_all_restore_history(days, u"computerId", device_id)
+
+    def update_cold_storage_purge_date(self, archive_guid, purge_date):
+        """Updates the cold storage purge date for a specified archive.
+        REST Documentation <https://www.crashplan.com/apidocviewer/#ColdStorage-put>
+
+        Args:
+            archive_guid (str): The identification number of the archive that should be updated
+            purge_date (str): The date on which the archive should be purged in yyyy-MM-dd format
```
alanag13, line 116: this made me realize that we have some places in the docs that site to use `YYYY-MM-DD` format, which is _technically_ not correct, thanks.

alanag13:
@ceciliastevens thanks for contributing! Can you describe the easiest way for us to verify that your change is working as intended? Is there anywhere in the UI where this value would be visible?

ceciliastevens:
@alanag13 Cold storage purge dates can be seen in the console, as documented here: https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Cold_storage#Change_cold_storage_duration

I've also addressed the other issues raised in the comments here.

src/py42/modules/archive.py
ceciliastevens: I think I've got it right now.

src/py42/modules/archive.py
ceciliastevens: Updated. My habit is to use crashplan.com when looking at the documentation, so I didn't even think about it.

src/py42/modules/archive.py
unparalleled-js: The tic marks were not showing up in the diff so I went in edit mode to see and there was space between :class: and the class name, so that may have been a small issue. I just went ahead and removed the space just in case!

PR #99, opened 2020-04-28
Minor/refactor tenant id from alertquery

timabrmsn:


src/py42/sdk/queries/alerts/filters/alert_filter.py
```diff
@@ -29,7 +29,7 @@ def not_contains(cls, value):
 class DateObserved(QueryFilterTimestampField):
     """Class that filters alerts based on the timestamp the alert was triggered."""
 
-    _term = u"CreatedAt"
+    _term = u"createdAt"
```
alanag13, line 32: I recall you saying that there was a reason this was set to `CreatedAt`, is this correct and `CreatedAt` is just for the sort key?

timabrmsn: Correct, the "term" for the query is `createdAt` in both what the browser sends and what the response contains (although the endpoint does accept `CreatedAt`).

But the sort key has to be `CreatedAt` and does _not_ accept `createdAt`.

src/py42/clients/alerts.py
```diff
@@ -94,3 +94,12 @@ def reopen(self, alert_ids, tenant_id=None, reason=None):
         uri = self._uri_prefix.format(u"reopen-alert")
         data = {u"tenantId": tenant_id, u"alertIds": alert_ids, u"reason": reason}
         return self._session.post(uri, data=json.dumps(data))
+
+    def _validate_tenant_exists_on_query(self, query):
```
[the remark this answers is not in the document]
timabrmsn: done

PR #103, opened 2020-05-04
Fix

unparalleled-js:


alanag13:
calls to `get_detection_list_user_client` need to also be fixed

PR #104, opened 2020-05-06
add alert rule filters

timabrmsn:


src/py42/sdk/queries/alerts/filters/alert_filter.py
```diff
@@ -45,6 +45,42 @@ class RuleName(AlertQueryFilterStringField):
...
+    """
+
+    _term = u"ruleSource"
+
+    ALERTING = u"Alerting"
+    DEPARTING_EMPLOYEE = u"Departing Employee"
+    HIGH_RISK_EMPLOYEE = u"High Risk Employee"
+
+
+class RuleType(QueryFilterStringField):
```
unparalleled-js, line 70: What about the term?

PR #105, opened 2020-05-11
Feature/setitem response

alanag13:
adds `__setitem__` support to `Py42Response`.

### DIFF
--- a/src/py42/response.py
+++ b/src/py42/response.py
@@ -1,6 +1,7 @@
 import json
 
 from py42._internal.compat import reprlib, str
+from py42.exceptions import Py42Error
 
 
 class Py42Response(object):
@@ -16,17 +17,20 @@ class Py42Response(object):
         except ValueError:
             self._data_root = self._response.text or u""
 
-        # looping over a Py42Response will loop through list items, dict keys, or str characters
-        self._iter = iter(self._data_root)
-
     def __getitem__(self, key):
         return self._data_root[key]
 
     def __setitem__(self, key, value):
-        self._data_root[key] = value
+        try:
+            self._data_root[key] = value
+        except TypeError as e:
+            data_root_type = type(self._data_root)
+            message = u"The Py42Respose root is of type {}, but __setitem__ got a key of {} and value of {}, which is incompatible."
+            raise Py42Error(message)
 
     def __iter__(self):
-        return self._iter
+        # looping over a Py42Response will loop through list items, dict keys, or str characters
+        return iter(self._data_root)
 
     @property
     def encoding(self):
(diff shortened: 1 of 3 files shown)


src/py42/response.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, requests_response):
     def __getitem__(self, key):
         return self._data_root[key]
 
+    def __setitem__(self, key, value):
+        self._data_root[key] = value
```
timabrmsn, line 26: Should we add a check here to make sure `_data_root` isn't a string, and if it is print a helpful error? Maybe something like "Response data of type str does not support item assignment"?

src/py42/response.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, requests_response):
...
+
     def __iter__(self):
         return self._iter
```
timabrmsn, line 29: I know this isn't part of this change, but I just noticed this. Since `self._iter` is a single generator result of `iter(self._data_root)`, you can only iterate over a response once. We should have `__iter__` return a new instance of `iter(self._data_root)` each time it's called, no?

src/py42/response.py
alanag13: python will make something like happen automatically without us doing anything:
```
>>> test = "test"
>>> test[0] = 1
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'str' object does not support item assignment
>>> test["test"]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: string indices must be integers, not str
```

src/py42/response.py
alanag13: it's debatable whether or not being able to do this multiple times would be expected behavior, but one of the main goals of this project is approachability / ease of use, so I think I agree with you.

src/py42/response.py
timabrmsn: yeah, I'm just thinking about the fact that since we're doing some misdirection on the __setitem__ it could be confusing to debug because the types don't match. 

I could see an end user saying: "what str object? I tried to set a value on a Py42Response object"

alanag13: shouldn't the trace help with that?

alanag13: I suppose we could catch/raise this in a `Py42Error` and add some additional context

src/py42/response.py
```diff
@@ -16,14 +17,25 @@ def __init__(self, requests_response):
...
-        # looping over a Py42Response will loop through list items, dict keys, or str characters
-        self._iter = iter(self._data_root)
-
     def __getitem__(self, key):
-        return self._data_root[key]
+        try:
+            return self._data_root[key]
+        except TypeError as e:
+            data_root_type = type(self._data_root)
+            message = u"The Py42Respose root is of type {}, but __getitem__ got a key of {}, which is incompatible."
```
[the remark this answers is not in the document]
alanag13: good find thanks!

PR #109, opened 2020-05-18
Feature/alert rule lookups

alanag13:
These will help us re: some error messages that are needed in the cli, namely, the cli currently does not error if you provide an invalid id. Tests coming soon


src/py42/_internal/clients/alerts.py
```diff
@@ -73,19 +79,45 @@ def get_all_rules(self, sort_key=u"CreatedAt", sort_direction=u"DESC"):
...
-    def get_rules_by_name(self, rule_name):
-        rule_pages = self.get_all_rules()
-        matched_rules = []
-        for rule_page in rule_pages:
-            rules = rule_page[u"ruleMetadata"]
-            for rule in rules:
-                if rule_name.lower() == rule[u"name"].lower():
-                    matched_rules.append(rule)
-        return matched_rules
+    def get_all_rules_by_name(self, rule_name, sort_key=u"CreatedAt", sort_direction=u"DESC"):
```
alanag13, line 87: Basically by looking at the source code I was able to figure out that these filter clauses work. I don't think that rules are complex that we need to implement the full query system for them, at least not yet. If there is demand for that we can keep still add that and keep these functions.

src/py42/modules/alertrules.py
```diff
@@ -84,14 +84,27 @@ def get_all(self, sort_key="CreatedAt", sort_direction="DESC"):
         alerts_client = self.microservice_client_factory.get_alerts_client()
         return alerts_client.get_all_rules(sort_key=sort_key, sort_direction=sort_direction)
 
-    def get_by_name(self, rule_name):
+    def get_all_by_name(self, rule_name):
```
alanag13, line 87: This is a breaking change for a couple reasons:
- method name change
- returns generator instead of list

However, I'm okay with it because it it was very recently introduced, nothing we know of is using it, and and it makes the API more concistent.

tests/clients/test_alerts.py
```diff
@@ -283,22 +283,3 @@ def test_get_all_posts_expected_data(self, mock_session, user_context):
             and posted_data["srtKey"] == "key"
             and posted_data["srtDirection"] == "ASC"
         )
-
-    def test_get_by_name_filters_correct_record(self, mock_get_all_session, user_context):
```
alanag13, line 287: These tests arent needed anymore because they test filtering that happens _after_ we get the response, Which will no longer be the case (the filtering will be the responsibility of the alert api now)

src/py42/_internal/clients/alerts.py
```diff
@@ -73,19 +79,45 @@ def get_all_rules(self, sort_key=u"CreatedAt", sort_direction=u"DESC"):
...
+        tenant_id = self._user_context.get_current_tenant_id()
+        return get_all_pages(
+            self._get_alert_rules,
+            u"ruleMetadata",
+            tenant_id=tenant_id,
+            groups=[
```
unparalleled-js, line 93: Should we create a shared function to create this blob for these functions? It could take the term as a parameter. We could also look into using code from the query services too, so that at any time, it is mindless to create FilterClause/Group looking JSON

alanag13: we already have one! switched to use that.

src/py42/_internal/clients/alerts.py
```diff
@@ -73,19 +81,33 @@ def get_all_rules(self, sort_key=u"CreatedAt", sort_direction=u"DESC"):
...
+        return get_all_pages(
+            self._get_alert_rules,
+            u"ruleMetadata",
+            tenant_id=tenant_id,
+            groups=[json.loads(str(create_eq_filter_group(u"Name", rule_name)))],
+            sort_key=sort_key,
+            sort_direction=sort_direction,
+        )
+
+    def get_rule_by_observer_id(self, observer_id, sort_key=u"CreatedAt", sort_direction=u"DESC"):
```
unparalleled-js, line 100: Does this not ever return more rules? Can a rule have multiple observer ids?

tests/clients/test_alerts.py
```diff
@@ -284,21 +284,52 @@ def test_get_all_posts_expected_data(self, mock_session, user_context):
...
+        filter_group = posted_data["groups"][0]["filters"][0]
+
+        assert filter_group["term"] == "ObserverRuleId"
+        assert filter_group["operator"] == "IS"
+        assert filter_group["value"] == "testid"
+        assert mock_session.post.call_args[0][0] == "/svc/api/v1/rules/query-rule-metadata"
+        assert (
+            posted_data["tenantId"] == user_context.get_current_tenant_id()
+            and posted_data["groupClause"] == "AND"
+            and posted_data["pgNum"] == 0
```
unparalleled-js, line 331: Have we tested these to make sure that pages start at 0 and not 1 ?

src/py42/_internal/clients/alerts.py
alanag13: nope

tests/clients/test_alerts.py
alanag13: it uses the same endpoint as stuff where we know that to be the case, and i've also physically tested this, so yes

PR #113, opened 2020-05-26
Saved search

kiran-chaudhary:
Created PR to review design.


src/py42/modules/investigation.py
```diff
@@ -0,0 +1,7 @@
+class InvestigateModule(object):
```
alanag13, line 1: Let structure this this way:

`sdk.securitydata.savedsearches`
- `get()`
- `get_by_id()`
- `execute()`

i.e. add the the saved_search_client to `SecurityModule` instead of a new `InvestigateModule`.

src/py42/_internal/client_factories.py
```diff
@@ -62,6 +63,7 @@ def __init__(
         self._detection_list_user_client = None
         self._ecm_session = None
         self._alert_rules_client = None
```
alanag13, line 65: add `self._file_event_session = None`  then set and return it (like we do with `ecm_session`) in `get_file_event_client()`

src/py42/_internal/client_factories.py
```diff
@@ -111,6 +113,12 @@ def get_alert_rules_client(self):
             )
         return self._alert_rules_client
 
+    def get_saved_search_client(self):
+        if not self._saved_search_client:
+            session = self._get_jwt_session(u"FORENSIC_SEARCH-API_URL")
```
alanag13, line 118:
```python
if not self._file_event_session:
    self._file_event_session = self._get_jwt_session(u"FORENSIC_SEARCH-API_URL")
self._saved_search_client = SavedSearchClient(self._file_event_session, self.get_file_event_client())
```

src/py42/_internal/clients/investigation.py
```diff
@@ -0,0 +1,40 @@
...
+        if u"searches" in response:
+            return {
+                key: search[key]
+                for search in response[u"searches"]
+                for key in self._response_header
+            }
+
+    def get(self):
+        uri = u"{}".format(self._resource)
+        response = self._session.get(uri)
```
alanag13, line 26: Is there a reason you chose to not simply return `self._session.get(uri)` here?

kiran-chaudhary: One of the acceptance criteria in the ticket is " **The output should be the same as if it were run in the console** "

Here I interpreted console as either of _web-console_ or _console.us.code42.com_ and hence set the fields that are displayed in web when a user wants to view existing saved search list and hence the parser in next line.

kiran-chaudhary: Moreover didn't add the same to `get_by_id` so that if a user intends to view the complete saved search detail they can use the same, whereas `get` helps to view the id, name and other metadata.

kiran-chaudhary: Also do note, used `get` over `get_all` because the response doesn't have paginated data. Since by convention `get_all` returns a generator to indicated paginated response.

src/py42/modules/investigation.py
kiran-chaudhary: Moved.

src/py42/_internal/client_factories.py
kiran-chaudhary: Changed.

src/py42/_internal/client_factories.py
kiran-chaudhary: Done.

src/py42/_internal/clients/investigation.py
```diff
@@ -0,0 +1,40 @@
...
+        uri = u"{}/{}".format(self._resource, search_id)
+        return self._session.get(uri)
+
+    def execute(self, search_id):
+        response = self.get_by_id(search_id)
+        if u"searches" in response:
+            query = {}
+            query[u"purpose"] = u"USER_EXECUTED_SEARCH"
+            query[u"groups"] = response["searches"][0]["groups"]
+            print(query)
```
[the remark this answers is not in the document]
kiran-chaudhary: Removed.

src/py42/_internal/clients/securitydata.py
```diff
@@ -6,3 +6,66 @@ def get_security_event_locations(self, user_uid):
...
+        Args:
+            search_id (str): Unique search Id of the saved search, Required.
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        uri = u"{}/{}".format(self._resource, search_id)
+        return self._session.get(uri)
+
+    @staticmethod
+    def _make_query(response):
```
kiran-chaudhary, line 54: Right now `execute` method fails with `BadRequest` error,  I have tried various ways to make the desired query but still can't get it right. 

`return FileEventQuery(*response[u"searches"][0][u"groups"])`.  Need to fix this, still working on it.

src/py42/_internal/clients/investigation.py
alanag13: Sorry for the confusion -- since this api is not well documented, that instruction was meant to guide you to use the same call / response seen in the web console. Please have this simply return `self._session.get(uri)`.

src/py42/_internal/clients/securitydata.py
alanag13: I can help with this if need be

src/py42/_internal/clients/investigation.py
kiran-chaudhary: ok.

src/py42/_internal/clients/securitydata.py
kiran-chaudhary: Yes, Please, I have already spent few hours on it, comparing the request against web-console.

src/py42/modules/securitydata.py
```diff
@@ -14,6 +14,15 @@ def __init__(self, security_client, storage_client_factory, microservices_client
         self._client_cache = {}
         self._client_cache_lock = Lock()
 
+    @property
+    def savedsearches(self):
+        """A collection of methods related to retrieving forensic search data.
+
+        Returns:
+            :class: `py42.clients.investigation.SavedSearchClient`
```
unparalleled-js, line 22: I think `investigation` should say `securitydata`

unparalleled-js: And `_internal`

src/py42/_internal/clients/securitydata.py
unparalleled-js: When I run with debug mode, I get this data "{'pgNum': 1, 'pgSize': 100, 'purpose': 'USER_EXECUTED_SEARCH', 'groups': [{'filters': [{'term': 'eventTimestamp', 'operator': 'ON_OR_AFTER', 'value': '2020-05-01T00:00:00.000Z', 'display': None}], 'filterClause': 'AND', 'display': '{\"data\":{\"isMultivalue\":false},\"version\":\"v1\"}'}, {'filters': [{'term': 'eventType', 'operator': 'IS', 'value': 'DELETED', 'display': None}, {'term': 'eventType', 'operator': 'IS', 'value': 'EMAILED', 'display': None}, {'term': 'eventType', 'operator': 'IS', 'value': 'MODIFIED', 'display': None}, {'term': 'eventType', 'operator': 'IS', 'value': 'READ_BY_APP', 'display': None}, {'term': 'eventType', 'operator': 'IS', 'value': 'CREATED', 'display': None}], 'filterClause': 'OR', 'display': '{\"data\":{\"isMultivalue\":true},\"version\":\"v1\"}'}]}"

The `display: None` seems concerning

unparalleled-js: I think we can just remove `display`. Isn't that a UI property? Also, https://py42docs.code42.com/en/stable/userguides/gettingstarted.html#troubleshooting-and-support shows how to turn on debug mode. It is really useful when you are not getting the expected status codes because it prints request parameters to stdout

alanag13: FYI I was able to get this working, ended up adding some other useful things. Adding some tests and such and I'll make it available shortly.

unparalleled-js: Actually, if you convert these to be "null", instead of null, I think it should work..

src/py42/modules/securitydata.py
kiran-chaudhary: Copy paste is a sin. Thanks.

src/py42/sdk/queries/__init__.py
```diff
@@ -9,6 +14,13 @@ def __init__(self, *args, **kwargs):
         self.page_number = None
         self.sort_key = None
 
+    @classmethod
```
alanag13, line 17: added `from_dict` to `BaseQuery`, `FilterGroup`, and `QueryFilter`. This allows these objects to be more easily built directly from JSON responses like those that come from saved search.

src/py42/sdk/queries/alerts/alert_query.py
```diff
@@ -36,3 +36,17 @@ def __str__(self):
             self.sort_key,
         )
         return json
+
+    def __iter__(self):
```
alanag13, line 40: I also added `__iter__` functions to `AlertQuery`, `FileEventQuery`, `FilterGroup`, and `QueryFilter`. This allows `dict(obj)` to return a dictionary representation of the JSON version of those objects easily.

src/py42/sdk/queries/query_filter.py
```diff
@@ -176,12 +180,33 @@ def __init__(self, term, operator, value=None):
         self._operator = operator
         self._value = value
 
+    @classmethod
+    def from_dict(cls, _dict):
+        return cls(_dict[u"term"], _dict[u"operator"], value=_dict.get(u"value"))
+
+    @property
+    def term(self):
```
alanag13, line 188: I exposed these so that they can be explored more easily when iterating through a list of filter groups and trying to do logic based on them

src/py42/sdk/queries/query_filter.py
```diff
@@ -190,14 +215,37 @@ class FilterGroup(object):
...
     def __init__(self, filter_list, filter_clause=u"AND"):
         self._filter_list = filter_list
         self._filter_clause = filter_clause
 
+    @classmethod
+    def from_dict(cls, _dict, filter_clause=u"AND"):
+        filter_list = [QueryFilter.from_dict(item) for item in _dict[u"filters"]]
+        return cls(filter_list, filter_clause=filter_clause)
+
+    @property
```
alanag13, line 233: I exposed these so that they can be explored more easily when iterating through a list of filter groups and trying to do logic based on them

src/py42/_internal/clients/securitydata.py
```diff
@@ -1,8 +1,58 @@
...
+        Args:
+            search_id (str): Unique search Id of the saved search.
+            pg_num (int): The consecutive group of results of size pg_size in the result set to return.
+            pg_size (int): The maximum number of results to be returned.
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        response = self.get_by_id(search_id)
+        search = response[u"searches"][0]
+        query = FileEventQuery.from_dict(search)
```
alanag13, line 57: I realize that this, in the long run, converts the response from a string (from requests) to a dict (Py42Response), to an object (here), back to a string (right before the search) -- but for maintenance's sake I think it's worth having a single concise way to do this sort of thing.

src/py42/_internal/client_factories.py
```diff
@@ -111,6 +115,15 @@ def get_alert_rules_client(self):
             )
         return self._alert_rules_client
 
+    def get_saved_search_client(self):
+        if not self._saved_search_client:
+            if not self._file_event_session:
```
unparalleled-js, line 120: Opportunity for making a `_get_file_event_session()` function that handles the lazy setting / returning and can be used by `get_file_event_client()` and `get_saved_search_client()`

src/py42/_internal/clients/securitydata.py
```diff
@@ -1,8 +1,58 @@
...
+        uri = u"{}/{}".format(self._resource, search_id)
+        return self._session.get(uri)
+
+    def execute(self, search_id, pg_num=1, pg_size=10000):
+        """
+        Execute a saved search for given search Id and return its results.
+
+        Args:
+            search_id (str): Unique search Id of the saved search.
+            pg_num (int): The consecutive group of results of size pg_size in the result set to return.
```
unparalleled-js, line 50: Missing `, optional` for the optional args.

alanag13:
@unparalleled-js @timabrmsn I still want to add docs/tests around the new properties/functions I added (that should address the codecov report differences), but otherwise still wanted to make this available for review.

src/py42/_internal/clients/securitydata.py
unparalleled-js: As well as `Defualts to 1` and `Defaults to 1000`.

unparalleled-js:
Can we get tests around the `dict()` and `from_dict()` stuff?

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+- `sdk.securitydata.savedsearches` module with functions:
```
unparalleled-js, line 13: Could mention the object type here (SavedSearchClient) since that is technically public too

PR #119, opened 2020-06-04
Feature/cold storage get

alanag13:
Adds the ability to get the list of cold storage archives for a given org Id.

This was originally also supposed to contain a method to get cold storage by destination guid, but I ended up excluding that for a couple reasons:
- I found that that api required a `destinationId` rather than a `destinationGuid` (confusing for users, I didn't even know there was such a thing).
- Even when I tried the API with that knowledge, a customer cloud admin still returned 403s for valid ids -- I think this feature may have been on-prem specific, which kind of makes sense since most of the concept of managing destinations is hidden in the cloud.
- There is no functionality in the UI for getting this info by destination, so I don't think it will be missed.

### DIFF
--- a/src/py42/modules/archive.py
+++ b/src/py42/modules/archive.py
@@ -123,31 +123,33 @@ class ArchiveModule(object):
     def get_all_cold_storage_archives_by_org_id(
         self, org_id, include_child_orgs=True, sort_key=None, sort_dir="asc"
     ):
-        """[summary]
+        """Returns a detailed list of cold storage archive information for a given org ID.
 
         Args:
-            org_id ([type]): [description]
-            include_child_orgs (bool, optional): [description]. Defaults to True.
-            sort_key ([type], optional): [description]. Defaults to None.
-            sort_dir (str, optional): [description]. Defaults to "asc".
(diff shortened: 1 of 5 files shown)


src/py42/modules/archive.py
```diff
@@ -119,3 +119,23 @@ def update_cold_storage_purge_date(self, archive_guid, purge_date):
...
+    ):
+        """Returns a detailed list of cold storage archive information for a given org ID.
+
+        Args:
+            org_id (str): The ID of a Code42 organization.
+            include_child_orgs (bool, optional): Determines whether cold storage information from
+             the Org's children is also returned. Defaults to True.
+            sort_key (str, optional): Sets the property by which the returned results will be sorted.
+             Choose from archiveHoldExpireDate, orgName, mountPointName, archiveBytes, and archiveType. Defaults to archiveHoldExpireDate.
+            sort_dir (str, optional): Sets the order by which sort_key should be sorted. Choose from
```
unparalleled-js, line 134: Little thing, but some of the ”defaults to” doc string values include quotes and some don't

PR #124, opened 2020-06-15
added detectionlists.refresh_user_scim_attributes method

ceciliastevens:
This request adds a method to refresh user SCIM data. This won't be necessary once DES can do this automatically, but for now it's useful as there isn't an intuitive way to do this.

REST documentation here: https://ecm-default.prod.ffs.us2.code42.com/svc/swagger/index.html?urls.primaryName=v2#/User/UserControllerV2_RefreshUser

Let me know if you'd like me to change or add anything else.

### DIFF
--- a/src/py42/modules/detectionlists.py
+++ b/src/py42/modules/detectionlists.py
@@ -125,10 +125,11 @@ class DetectionListsModule(object):
         self._detection_list_user_client.remove_cloud_alias(user_id, alias)
 
     def refresh_user_scim_attributes(self, user_id):
-        """Refresh SCIM attributes for the user.
+        """Refresh SCIM attributes of a user.
+        `REST documentation <https://ecm-default.prod.ffs.us2.code42.com/svc/swagger/index.html?urls.primaryName=v2#/User/UserControllerV2_RefreshUser>`__
 
         Args:
-            user_id (str or int): The Code42 userUid whose attributes you wish to refresh.
+            user_id (str or int): The Code42 userUid of the user whose attributes you wish to refresh.
 
         Returns:
             :class:`py42.response.Py42Response`
--- a/src/py42/modules/detectionlists.py
+++ b/src/py42/modules/detectionlists.py
@@ -123,3 +123,16 @@ class DetectionListsModule(object):
         factory = self._microservice_client_factory
         self._detection_list_user_client = factory.get_detection_list_user_client()
         self._detection_list_user_client.remove_cloud_alias(user_id, alias)
+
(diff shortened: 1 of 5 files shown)


alanag13:
Thanks for adding another contribution @ceciliastevens! Could you add a changelog entry and unit tests for this? Otherwise it looks good to me!

src/py42/modules/detectionlists.py
```diff
@@ -123,3 +123,16 @@ def remove_user_cloud_alias(self, user_id, alias):
         factory = self._microservice_client_factory
         self._detection_list_user_client = factory.get_detection_list_user_client()
         self._detection_list_user_client.remove_cloud_alias(user_id, alias)
+
+    def refresh_user_scim_attributes(self, user_id):
```
alanag13, line 127: Could you add a link to the relevant REST documentation that is behind this method to the docstring?

src/py42/modules/detectionlists.py
```diff
@@ -123,3 +123,16 @@ def remove_user_cloud_alias(self, user_id, alias):
...
+        """Refresh SCIM attributes for the user.
+
+        Args:
+            user_id (str or int): The Code42 userUid whose attributes you wish to refresh.
```
alanag13, line 131: `...userUid of the user whose`

src/py42/modules/detectionlists.py
```diff
@@ -123,3 +123,16 @@ def remove_user_cloud_alias(self, user_id, alias):
         factory = self._microservice_client_factory
         self._detection_list_user_client = factory.get_detection_list_user_client()
         self._detection_list_user_client.remove_cloud_alias(user_id, alias)
+
+    def refresh_user_scim_attributes(self, user_id):
+        """Refresh SCIM attributes for the user.
```
alanag13, line 128: `... the SCIM attributes of a user`

src/py42/modules/detectionlists.py
ceciliastevens: Added.

src/py42/modules/detectionlists.py
ceciliastevens: I've updated this

src/py42/modules/detectionlists.py
ceciliastevens: Now updated

ceciliastevens:
> See above for some small feedback items. Thanks again!

I've updated the things you pointed out.

alanag13:
@ceciliastevens thanks, we'll get this tested and hopefully added soon!

PR #127, opened 2020-06-17
fix logging regression

alanag13:
This fixes a regression that caused `import py42.settings.debug` to fail ("`from py42.settings import debug` still worked).

The problem was caused by the fact that `import` statements can only import modules and not variables or classes, which require using `from module import symbol`.

The solution essentially fools python into thinking that our new class is a module.

To test:
- Install the latest version of the cli into a virtual env.
- Observe that simply typing `code42` results in error re: `import py42.settings.debug` not working
`pip install -e .[dev]` this branch into that same virtual env
- Observe that typing `code42` now behaves as normal

All existing tests re: debug settings still passing.


src/py42/_internal/session.py
```diff
@@ -176,8 +176,6 @@ def _renew_authentication(self, use_cache=False):
 
     def _print_request(self, method, url, params=None, data=None):
         debug.logger.info(u"{0}{1}".format(str(method).ljust(8), url))
-        if self.headers:
```
alanag13, line 179: I removed this for two reasons:
- Security reasons (basic credentials and tokens were being logged)
- When updating to py42 1.5.0, if you were using a level of `DEBUG` (as the cli does), your logging became more verbose than it was previously (removing this makes it the same).

src/py42/settings/debug.py
```diff
@@ -0,0 +1,24 @@
+import logging
+import sys
+
+
+class _DebugSettings(object):
```
alanag13, line 5: this class is exactly the same as it was in `__init__`, moving it here and assigning it to `sys.modules` (see the last line of this file) makes the class be able to be imported via `import py42.settings.debug`. `from py42.settings import debug` also still works.

src/py42/_internal/session.py
timabrmsn: Could we add a new `debug.log_headers` setting to make it possible but also need it set explicitly?

src/py42/settings/debug.py
timabrmsn: Nice, didn’t know about this possibility.

PR #129, opened 2020-06-22
Feature/pds stream file by hash

kiran-chaudhary:
Added methods: 
   `securitydata.stream_file_by_sha256`  and `sdk.securitydata.stream_file_by_md5`


src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+            yield pds_client.find_file_versions(md5_hash, sha256_hash, device_id, paths)
+
+    def _stream_file(self, file_generator, file):
+        for response in file_generator:
+            if response.status_code != 200:
+                continue
+            session = self._microservices_client_factory.get_storage_node_session(
+                response[u"storageNodeURL"]
+            )
+            storage_node_client = StorageNodeClient(
```
kiran-chaudhary, line 193: Do we need to create `StorageNodeClient` in `client_factories` ?

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,29 @@
...
+    def get_download_token(self):
+
+        params = "archiveGuid={0}&fileId={1}&versionTimestamp={2}".format(
+            self._archive_guid, self._file_id, self._timestamp
+        )
+        resource = u"FileDownloadToken"
+        uri = "{0}{1}?{2}".format(self._base_uri, resource, params)
+        return self._session.get(uri)
+
+    def get_file(self, host_address, token, file):
```
kiran-chaudhary, line 24: Should we write content to a file or stdout??

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+                session,
+                response[u"archiveGuid"],
+                response[u"fileId"],
+                response[u"versionTimestamp"],
+            )
+            token = storage_node_client.get_download_token()
+            return storage_node_client.get_file(response[u"storageNodeURL"], token, file)
+        raise FileNotFoundError("No file available for download.")
+
+    def stream_file_by_sha256(self, checksum):
```
kiran-chaudhary, line 203: If writing data to file is fine, do we want user to pass additional parameter, optionally, "path with filename" where data needs to be downloaded? 

Right now it writes to a file with the given checksum.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+            )
+            storage_node_client = StorageNodeClient(
+                session,
+                response[u"archiveGuid"],
+                response[u"fileId"],
+                response[u"versionTimestamp"],
+            )
+            token = storage_node_client.get_download_token()
+            return storage_node_client.get_file(response[u"storageNodeURL"], token, file)
+        raise FileNotFoundError("No file available for download.")
```
kiran-chaudhary, line 201: This message is same as in web, `console.us.code42.com`

src/py42/clients/file_event.py
```diff
@@ -23,3 +23,13 @@ def search(self, query):
         query = str(query)
         uri = u"/forensic-search/queryservice/api/v1/fileevent"
         return self._session.post(uri, data=query)
+
+    def get_file_location_detail_by_sha256(self, hash):
```
kiran-chaudhary, line 27: Added a new method to `FileEventClient`.

kiran-chaudhary:
WIP, need to add change log and tests.

src/py42/_internal/clients/storage/storagenode.py
unparalleled-js: No, we should just return the stream. The user can decide where to save it on their own.

There is an example in the README we can append to explaining how to do this

src/py42/modules/securitydata.py
unparalleled-js: Probably, then we also wouldn't need `get_storage_node_session()`

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,29 @@
...
+        resource = u"GetFile"
+        uri = "{0}/{1}{2}?{3}".format(host_address, self._base_uri, resource, token)
+        with requests.get(uri, stream=True) as reader:
```
kiran-chaudhary, line 27: Using `self._session.get` returned 406  and using `requests.get` worked, will revisit it again.

src/py42/_internal/clients/storage/storagenode.py
alanag13: Yeah, the user may want to do something else with the stream besides writing to a file, (e.g. making it streamable through a browser or printing the raw stream to a terminal), so returning the stream is the expected implementation. This is what we do for `stream_file_from_backup`.

src/py42/modules/securitydata.py
alanag13: Per above, simply return the stream and do not write a file.

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def __init__(self, session, archive_guid, file_id, timestamp):
+        super(StorageNodeClient, self).__init__(session)
+        self._archive_guid = archive_guid
+        self._file_id = file_id
+        self._timestamp = timestamp
+
+    def get_download_token(self):
+
+        params = "archiveGuid={0}&fileId={1}&versionTimestamp={2}".format(
```
alanag13, line 17: in order to ensure url escaping works properly, make `params` a dict, and pass it to the `params` arg of `get`:
```python
params = {
    u"archiveGuid": self._archive_guid,
    u"fileId": self._file_id,
    u"versionTimestamp": self._timestamp
}
uri = "{0}{1}".format(self._base_uri, u"FileDownloadToken")
return self._session.get(uri, params=params)
    
``

File: src/py42/_internal/clients/storage/storagenode.py [hunk 8]
```diff
@@ -0,0 +1,29 @@
...
+            self._archive_guid, self._file_id, self._timestamp
+        )
+        resource = u"FileDownloadToken"
+        uri = "{0}{1}?{2}".format(self._base_uri, resource, params)
+        return self._session.get(uri)
+
+    def get_file(self, host_address, token, file):
+        resource = u"GetFile"
+        uri = "{0}/{1}{2}?{3}".format(host_address, self._base_uri, resource, token)
```
alanag13, line 26: use params arg as mentioned above

src/py42/clients/file_event.py
```diff
@@ -23,3 +23,13 @@ def search(self, query):
...
+        uri = u"/forensic-search/queryservice/api/v1/filelocations?sha256={0}".format(hash)
```
alanag13, line 28: use params arg as mentioned above

src/py42/clients/file_event.py
```diff
@@ -23,3 +23,13 @@ def search(self, query):
...
+        response = self._session.get(uri)
+        file_name = response[u"locations"][0][u"fileName"]
+        paths = []
+        for location in response[u"locations"]:
+            device_id = location[u"deviceUid"]
+            paths.append("{0}{1}".format(location[u"filePath"], file_name))
+            yield device_id, paths
```
alanag13, line 35: clients should always return `Py42Response` or a generator of `Py42Response` objects, so this method should simply return the `response` on line 29.

src/py42/_internal/clients/storage/storagenode.py
alanag13, line 24: This method needs a docstring

src/py42/clients/file_event.py
alanag13, line 27: This method needs a docstring

src/py42/clients/pds.py
```diff
@@ -0,0 +1,24 @@
+from py42.clients import BaseClient
+
+
+class PreservationDataServiceClient(BaseClient):
```
alanag13, line 4: This client should be internal

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+
+    def stream_file_by_sha256(self, checksum):
+        """Stream file based on SHA256 checksum.
+
+        Args:
+            checksum (str): SHA256 encoded hash of the file.
```
alanag13, line 207: the word `encoded` is extraneous here

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+    def stream_file_by_md5(self, checksum):
+        """Stream file based on MD5 checksum.
+
+        Args:
+            checksum (str): MD5 encoded hash of the file.
+
+        Returns:
+            Downloads the requested file if found.
+        """
+        events = self._search_by_hash(checksum, MD5)
```
alanag13, line 228: the word `encoded` is extraneous here

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,29 @@
...
+        with requests.get(uri, stream=True) as reader:
+            with open(file, "wb") as local_file:
```
alanag13, line 28: don't write to a file, as discussed

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,29 @@
+from py42.clients import BaseClient
+import requests
+
+
+class StorageNodeClient(BaseClient):
```
alanag13, line 5: can we call this `StoragePreservationDataClient`? All of the other methods in this folder also apply to storage nodes, so this name is a little generic.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+
+    def _find_file_versions(self, md5_hash, sha256_hash):
+        file_event_client = self._microservices_client_factory.get_file_event_client()
+        pds_client = self._microservices_client_factory.get_preservation_data_service_client()
+        for device_id, paths in file_event_client.get_file_location_detail_by_sha256(sha256_hash):
+            yield pds_client.find_file_versions(md5_hash, sha256_hash, device_id, paths)
+
+    def _stream_file(self, file_generator, file):
+        for response in file_generator:
+            if response.status_code != 200:
```
alanag13, line 188: In addition to this, we will want to wrap the inner loop in `try/except` and catch `Py42HttpErrors` to continue when they occur. If we iterate over the whole loop and never successfully get the stream, raise a `Py42Error`.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +172,65 @@ def search_file_events(self, query):
...
+        raise FileNotFoundError("No file available for download.")
+
+    def stream_file_by_sha256(self, checksum):
+        """Stream file based on SHA256 checksum.
+
+        Args:
+            checksum (str): SHA256 encoded hash of the file.
+
+        Returns:
+            Downloads the requested file if found.
```
alanag13, line 210: `Returns a stream of the requested file`

src/py42/_internal/client_factories.py
```diff
@@ -114,6 +116,15 @@ def get_saved_search_client(self):
             )
         return self._saved_search_client
 
+    def get_preservation_data_service_client(self):
+        if not self._pds_client:
+            session = self._get_jwt_session(u"PRESERVATION-DATA-SERVICE_API-URL")
+            self._pds_client = PreservationDataServiceClient(session)
+        return self._pds_client
+
+    def get_storage_node_session(self, host_address):
```
alanag13, line 125: `get_storage_node_preservation_session`

src/py42/modules/securitydata.py
alanag13: Yes, agreed. The api methods on storage nodes are not really microservices, so a `create_storage_preservation_client` method there would be ideal.

src/py42/modules/securitydata.py
```diff
@@ -332,6 +337,15 @@ def _get_plans_in_node(destination, node):
     }
 
 
+def parse_file_location_response(response):
```
alanag13, line 340: This method can be made private

src/py42/modules/securitydata.py
```diff
@@ -180,56 +184,57 @@ def _search_by_hash(self, hash, type):
...
+            try:
+                storage_node_client = self._microservices_client_factory.create_storage_preservation_client(
+                    response[u"storageNodeURL"]
+                )
+                token = storage_node_client.get_download_token(
+                    response[u"archiveGuid"], response[u"fileId"], response[u"versionTimestamp"],
+                )
+                return storage_node_client.get_file(str(token))
+            except Py42HTTPError as err:
+                raise_py42_error(err)
```
alanag13, line 204: we should `continue` if an error is raised here, that way we only will raise an exception if none of the items in the generator work. If we raise here, this function will exit for the first item in the list that fails.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +176,66 @@ def search_file_events(self, query):
...
+    def _search_by_hash(self, hash, type):
+        query = FileEventQuery.all(type.eq(hash))
+        response = self.search_file_events(query)
+        return response["fileEvents"]
+
+    def _find_file_versions(self, md5_hash, sha256_hash):
+        file_event_client = self._microservices_client_factory.get_file_event_client()
+        pds_client = self._microservices_client_factory.get_preservation_data_service_client()
+        response = file_event_client.get_file_location_detail_by_sha256(sha256_hash)
+        for device_id, paths in parse_file_location_response(response):
```
alanag13, line 188: the inside of this for loop should should have a `try/except` block (where the except simply passes) so that if we hit an error when trying to find file versions for a specific file id/path, we will keep trying the others in the list.

alanag13:
GitHub doesn't seem to be showing this comment in the summary tab so linking it here for visibility: https://github.com/code42/py42/pull/129/files#r444226826

src/py42/modules/securitydata.py
kiran-chaudhary: I understand the intent behind why we wanted to do that. However, we would be hiding errors/api-errors or incorrect client request errors in doing so. If in future the api introduces a bug or breaking change we would just end up ignoring them.

Allowing errors to pass would end up having similar effect as drawbacks of catching generic exception.

If there are particular cases we foresee that these APIs return which indicate kind of "a success state but cannot fulfil the request now ", we should be catching those errors and allowing them to pass.

As in below, I would like to change line 193 as :
      `if response.status_code == 204:
           continue`
since that is a special case where the state is success but there is no file to download.

alanag13: > Allowing errors to pass would end up having similar effect as drawbacks of catching generic exception.
Not quite. We would only be allowing errors that we _expect_ could happen to pass. If there was any type of error besides `Py42HttpError`, we would still raise.

I understand your point about hiding api errors, we should be telling users this information -- so my suggestion would be that if a `Py42HttpError` occurs, we should log it with `py42.settings.debug.logger.warn()` (but still not throw it). We could include an advisement to turn on debug logging in the top-level error that occurs if none of the entries work.

kiran-chaudhary: Yes, makes sense, I will add the logging.

src/py42/_internal/clients/pds.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class PreservationDataServiceClient(BaseClient):
+    def find_file_versions(self, file_md5, file_sha256, device_id, paths):
+        """Fetch file version details.
+
+        Args:
+            file_md5 (str): MD5 encoded hash of the file.
+            file_sha256 (str): SHA256 encoded hash of the file.
+            device_id (str): Device id.
```
unparalleled-js, line 11: Could use more info for `device_id`, like `Id of the device where the files are from.` or something

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Args:
+            token (str): PDS Download token.
+
+        Returns:
+            Returns a stream of the requested token.
+        """
+        resource = u"GetFile"
+        uri = "{0}/{1}{2}".format(self._session.host_address, self._base_uri, resource)
+        if token.__contains__(u"PDSDownloadToken="):
```
unparalleled-js, line 43: Consider `if u"PDSDownloadToken=" in token`: rather than calling `__contains__` magic method.

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+            token = token.replace(u"PDSDownloadToken=", "")
```
unparalleled-js, line 44: Actually, you don't even need the above check, replace won't do anything if the string isn't present

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Changed.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Done.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Done.

src/py42/clients/file_event.py
kiran-chaudhary: Done.

src/py42/clients/file_event.py
kiran-chaudhary: Done.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Added.

src/py42/clients/file_event.py
kiran-chaudhary: Added.

src/py42/clients/pds.py
kiran-chaudhary: Moved.

src/py42/modules/securitydata.py
kiran-chaudhary: Removed.

src/py42/modules/securitydata.py
kiran-chaudhary: Removed.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Done.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Changed.

src/py42/modules/securitydata.py
kiran-chaudhary: Changed.

src/py42/modules/securitydata.py
kiran-chaudhary: Changed.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/_internal/clients/pds.py
kiran-chaudhary: Done.

src/py42/_internal/clients/storage/storagenode.py
kiran-chaudhary: Replaced `PDSDownloadToken` as its redundant.  Removed magic method.

src/py42/_internal/archive_access.py
```diff
@@ -77,7 +77,10 @@ def _get_file_via_walking_tree(self, file_path):
             if root[u"path"].lower() == path_root.lower():
                 return self._walk_tree(root, path_parts[1:])
 
-        raise Py42ArchiveFileNotFoundError(self._device_guid, file_path)
+        message = u"File not found in archive " u"for device {0} at path {1}".format(
```
unparalleled-js, line 80: This string is divided on the same line

src/py42/_internal/archive_access.py
```diff
@@ -90,8 +93,10 @@ def _walk_tree(self, current_node, remaining_path_components):
         for child in children:
             if child[u"path"].lower() == target_child_path.lower():
                 return self._walk_tree(child, remaining_path_components[1:])
-
-        raise Py42ArchiveFileNotFoundError(self._device_guid, target_child_path)
+        message = u"File not found in archive " u"for device {0} at path {1}".format(
```
unparalleled-js, line 96: String is divided on the same line

src/py42/_internal/clients/pds.py
```diff
@@ -0,0 +1,24 @@
+from py42.clients import BaseClient
+
+
+class PreservationDataServiceClient(BaseClient):
+    def find_file_versions(self, file_md5, file_sha256, device_id, paths):
```
unparalleled-js, line 5: It lets you pass both md5 and sha256?  Should we just make two methods here, `find_file_versions_by_sha256()` and `find_file_versions_by_md5()` ?

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
+from py42.clients import BaseClient
+import requests
+
+
+class StoragePreservationDataClient(BaseClient):
+
+    _base_uri = u"c42api/v3/"
+
+    def __init__(self, session):
+        super(StoragePreservationDataClient, self).__init__(session)
```
unparalleled-js, line 10: I don't think this `init()` is needed

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def get_download_token(self, archive_guid, file_id, timestamp):
+        """Get PDS download token of a file.
```
unparalleled-js, line 13: for a file, since I assume the token is not identical upon each request, or is it?

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Args:
+            archive_guid (str): Archive guid of the file
+            file_id (str): Id of the file.
+            timestamp (int): Last updated timestamp of the file.
```
unparalleled-js, line 18: Is this milliseconds or seconds?

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+        Returns:
+            :class:`py42.response.Py42Response`: A response containing download token of the file.
+        """
+        params = {
+            u"archiveGuid": archive_guid,
+            u"fileId": file_id,
+            u"versionTimestamp": timestamp,
+        }
+        resource = u"FileDownloadToken"
+        uri = "{0}{1}".format(self._base_uri, resource)
```
unparalleled-js, line 29: Missing `u` here on this URL string

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+        """Streams a file.
+
+        Args:
+            token (str): PDS Download token.
+
+        Returns:
+            Returns a stream of the requested token.
+        """
+        resource = u"GetFile"
+        uri = "{0}/{1}{2}".format(self._session.host_address, self._base_uri, resource)
```
unparalleled-js, line 42: Missing `u` on this string

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,46 @@
...
+    def get_download_token(self, archive_guid, file_id, timestamp):
+        """Get PDS download token of a file.
+
+        Args:
+            archive_guid (str): Archive guid of the file
+            file_id (str): Id of the file.
+            timestamp (int): Last updated timestamp of the file.
+
+        Returns:
+            :class:`py42.response.Py42Response`: A response containing download token of the file.
```
unparalleled-js, line 21: .. for the file?

src/py42/exceptions.py
```diff
@@ -8,10 +8,7 @@ class Py42Error(Exception):
 class Py42ArchiveFileNotFoundError(Py42Error):
     """An exception raised when a resource file is not found or the path is invalid."""
 
-    def __init__(self, device_guid, file_path):
-        message = u"File not found in archive " u"for device {0} at path {1}".format(
-            device_guid, file_path
-        )
+    def __init__(self, message):
```
unparalleled-js, line 11: This init does not need to be defined

src/py42/_internal/clients/pds.py
unparalleled-js: I guess since this is an internal client, it isn't that important

src/py42/modules/securitydata.py
```diff
@@ -169,6 +177,71 @@ def search_file_events(self, query):
...
+
+    def _find_file_versions(self, md5_hash, sha256_hash):
+        file_event_client = self._microservices_client_factory.get_file_event_client()
+        pds_client = self._microservices_client_factory.get_preservation_data_service_client()
+        response = file_event_client.get_file_location_detail_by_sha256(sha256_hash)
+
+        for device_id, paths in _parse_file_location_response(response):
+            try:
+                yield pds_client.find_file_versions(md5_hash, sha256_hash, device_id, paths)
+            except Py42HTTPError:
```
unparalleled-js, line 193: Could you add a comment here explaining why we have to ignore all HTTPErrors here? Seems odd, which if the params are bad?

src/py42/modules/securitydata.py
```diff
@@ -169,6 +177,71 @@ def search_file_events(self, query):
...
+                storage_node_client = self._microservices_client_factory.create_storage_preservation_client(
+                    response[u"storageNodeURL"]
+                )
+                token = storage_node_client.get_download_token(
+                    response[u"archiveGuid"], response[u"fileId"], response[u"versionTimestamp"],
+                )
+                return storage_node_client.get_file(str(token))
+            except HTTPError:
+                pass
+        raise Py42Error("No file available for download.")
```
unparalleled-js, line 210: Missing python 2 `u` here

src/py42/modules/securitydata.py
```diff
@@ -169,6 +177,71 @@ def search_file_events(self, query):
...
+
+        Args:
+            checksum (str): SHA256 hash of the file.
+
+        Returns:
+            Returns a stream of the requested file.
+        """
+        events = self._search_by_hash(checksum, SHA256)
+        if not len(events):
+            message = u"File not found in archive with md5 checksum {0}".format(checksum)
```
unparalleled-js, line 223: This message should say `sha256` and not `md5`

src/py42/modules/securitydata.py
```diff
@@ -270,6 +343,16 @@ def _get_plans_in_node(destination, node):
     }
 
 
+def _parse_file_location_response(response):
+
+    for location in response[u"locations"]:
```
unparalleled-js, line 348: Any chance of response not having `locations`? Is this is a good place to fail if that is the good or should we just treat it the file not having a location?

unparalleled-js: e.g. raise a FileNotFound exception or something

src/py42/modules/securitydata.py
```diff
@@ -270,6 +343,16 @@ def _get_plans_in_node(destination, node):
...
+        paths = []
+        file_name = location[u"fileName"]
+        device_id = location[u"deviceUid"]
+        paths.append("{0}{1}".format(location[u"filePath"], file_name))
```
unparalleled-js, line 352: Missing `u""` here

tests/_internal/clients/test_pds.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+from py42._internal.clients.pds import PreservationDataServiceClient
+
+
+class TestPreservationDataServiceClient(object):
+    @pytest.fixture
+    def mock_session(self, mock_session, successful_response):
```
unparalleled-js, line 8: might be worth putting this as a `mock_successful_session` in a conftest

tests/modules/test_securitydata.py
```diff
@@ -572,3 +610,311 @@ def test_saved_searches_returns_saved_search_client(
...
+        storage_node_client.get_download_token.return_value = file_download
+        storage_node_client.get_file.return_value = b"stream"
+        microservice_client_factory.create_storage_preservation_client.return_value = (
+            storage_node_client
+        )
+
+        response = security_module.stream_file_by_sha256("shahash")
+        assert response == b"stream"
+
+    def test_stream_file_by_sha256_raises_file_not_found_error(
```
unparalleled-js, line 688: It'd be helpful if you include the `condition` as part of the test signature... something like `test_stream_file_by_sha256_when_given_unknown_hash_raises_file_not_found_error()`

tests/modules/test_securitydata.py
```diff
@@ -572,3 +610,311 @@ def test_saved_searches_returns_saved_search_client(
...
+        mocker,
+        security_client,
+        storage_client_factory,
+        microservice_client_factory,
+        file_event_search,
+    ):
+        security_module = SecurityModule(
+            security_client, storage_client_factory, microservice_client_factory
+        )
+        file_event_search.text = "{}"
```
unparalleled-js, line 833: Consider putting what the actual response looks like here for when the file isn't found/

unparalleled-js: Or is it really an empty response?

unparalleled-js: I guess, similar to a comment I made above, I don't understand the condition for this test

src/py42/modules/securitydata.py
alanag13: @kiran-chaudhary, I thought we discussed this previously and agreed that we would log a message if an exception occurs?

alanag13: we should also do similar logging in `_stream_file()`.

src/py42/_internal/archive_access.py
kiran-chaudhary: Changed.

tests/modules/test_securitydata.py
kiran-chaudhary: This is a negative test, case when response content is empty.

tests/modules/test_securitydata.py
kiran-chaudhary: Changed method name.

tests/_internal/clients/test_pds.py
kiran-chaudhary: Moved.

src/py42/modules/securitydata.py
kiran-chaudhary: Added.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/_internal/archive_access.py
kiran-chaudhary: Changed.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/modules/securitydata.py
kiran-chaudhary: done.

src/py42/_internal/clients/storage/storagenode.py
```diff
@@ -0,0 +1,47 @@
...
+        resource = u"GetFile"
+        uri = u"{0}/{1}{2}".format(self._session.host_address, self._base_uri, resource)
+        if u"PDSDownloadToken=" in token:
+            replaced_token = token.replace(u"PDSDownloadToken=", "")
+        else:
+            replaced_token = token
+        params = {u"PDSDownloadToken": replaced_token}
+        debug.logger.info(u"{0}{1}".format(str("GET").ljust(8), uri))
+        debug.logger.debug(format_dict(params, u"  params"))
+        return requests.get(uri, params=params, stream=True)
```
kiran-chaudhary, line 47: As using `session.get` returned 406 API error, I had to use requests module here. Added logging to make logging consistent with `session` module and hence moved `format_dict` method to `util` module.

src/py42/_internal/archive_access.py
```diff
@@ -77,7 +77,10 @@ def _get_file_via_walking_tree(self, file_path):
             if root[u"path"].lower() == path_root.lower():
                 return self._walk_tree(root, path_parts[1:])
 
-        raise Py42ArchiveFileNotFoundError(self._device_guid, file_path)
+        message = u"File not found in archive for device {0} at path {1}".format(
```
unparalleled-js, line 80: This is code is shared with what is on line 96 and can be replaced with a helper function that creates the exception

unparalleled-js: Or, another custom exception would also work

src/py42/exceptions.py
unparalleled-js: I don't understand this change. When we raise the exception, we have to repeat the same exact message each time. Why not just share the code for creating the message and keep it here?  If you want to override the message, you can subclass this exception. That way we don't have to repeat code!

I don't know what you mean about it being consistent with the other exceptions. Other exceptions craft the message inside `super.__init__()`

src/py42/modules/securitydata.py
```diff
@@ -169,6 +178,82 @@ def search_file_events(self, query):
...
+        response = self.search_file_events(query)
+        return response[u"fileEvents"]
+
+    def _find_file_versions(self, md5_hash, sha256_hash):
+        file_event_client = self._microservices_client_factory.get_file_event_client()
+        pds_client = self._microservices_client_factory.get_preservation_data_service_client()
+        response = file_event_client.get_file_location_detail_by_sha256(sha256_hash)
+
+        if u"locations" not in response and not len(response[u"locations"]):
+            raise Py42ArchiveFileNotFoundError(u"PDS service can't find requested file.")
```
alanag13, line 192: This should probably be a different, new type of exception, this type of exception is intended for when the device and file path are known, which is not the case here.

A simple `Py42Error` would probably be fine here (no need for a new exception type, and the message should contain the hash that was being searched for.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +178,82 @@ def search_file_events(self, query):
...
+
+        for device_id, paths in _parse_file_location_response(response):
+            try:
+                yield pds_client.find_file_versions(md5_hash, sha256_hash, device_id, paths)
+            except Py42HTTPError as err:
+                # API searches multiple paths to find the file to be streamed, as returned by
+                # 'get_file_location_detail_by_sha256', hence we keep looking until we find a stream
+                # to return
+                debug.logger.warning(u"PDS fetch file version API failed. Error: ", err)
```
alanag13, line 201: this log message should contain the hash somewhere.

src/py42/modules/securitydata.py
```diff
@@ -169,6 +178,82 @@ def search_file_events(self, query):
...
+                    response[u"archiveGuid"], response[u"fileId"], response[u"versionTimestamp"],
+                )
+                return storage_node_client.get_file(str(token))
+            except HTTPError as err:
+                # API searches multiple paths to find the file to be streamed, as returned by
+                # 'get_file_location_detail_by_sha256', hence we keep looking until we find a stream
+                # to return
+                debug.logger.warning(u"PDS stream file token API failed, Error:", err)
+                pass
+        raise Py42Error(u"No file available for download.")
```
alanag13, line 222: `No file with has <hash> available for download on any storage node`.

src/py42/_internal/archive_access.py
alanag13: I would put `Py42ArchiveFileNotFoundError` back the way it was before and either make a new exception type or just use `Py42Error`.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/modules/securitydata.py
kiran-chaudhary: Added. Do note, had to add both the hashes in this particular case, otherwise I had to pass the hash-type to indicate which hash method is requested. I was aware that passing requested hash values is important  but left it out since I felt its an overkill to pass variable only to set an error message. 

I am open to rework again, If needed I can do the change accordingly.

src/py42/modules/securitydata.py
kiran-chaudhary: Done.

src/py42/_internal/archive_access.py
kiran-chaudhary: Done.

src/py42/exceptions.py
kiran-chaudhary: Agreed. Reverted the changes as requested.

 In context to consistency I was referring to classes `Py42SecurityPlanConnectionError` and `Py42StorageSessionInitializationError`, which just take an error message and were identical to this change.

PR #130, opened 2020-06-25
Feature/expose getting single page methods

unparalleled-js:
I need methods for exposing get detection list pages using page size and page nuns for Demisto.

I want the investigator to be able load in data slowly at a time, observe it, load more in, do something else, come back, load more in, etc.  This is coming from feedback from PAN.

Two new methods:
* get_high_risk_employees_page()
* get_departing_employees_page()

Also fixed an issue with get_all_pages which allows you to specify a page size there instead of it always using the global one no matter what.

### DIFF
--- a/src/py42/clients/detectionlists/departing_employee.py
+++ b/src/py42/clients/detectionlists/departing_employee.py
@@ -68,20 +68,20 @@ class DepartingEmployeeClient(BaseClient):
         data = {u"userId": user_id, u"tenantId": tenant_id}
         return self._session.post(uri, data=json.dumps(data))
 
-    def _get_departing_employees_page(
+    def get_departing_employees_page(
         self,
         tenant_id=None,
         filter_type=None,
         sort_key=u"CREATED_AT",
         sort_direction=u"DESC",
         page_num=None,
-        page_size=None,
+        page_size=100,
     ):
 
         uri = self._uri_prefix.format(u"search")
         data = {
             u"tenantId": tenant_id,
-            u"pgSize": 100,
+            u"pgSize": page_size,
             u"pgNum": page_num,
             u"filterType": filter_type,
             u"srtKey": sort_key,
@@ -105,12 +105,13 @@ class DepartingEmployeeClient(BaseClient):
             that each contain a page of departing employees.
         """
         return get_all_pages(
-            self._get_departing_employees_page,
+            self.get_departing_employees_page,
             u"items",
             tenant_id=self._user_context.get_current_tenant_id(),
             filter_type=filter_type,
             sort_key=sort_key,
             sort_direction=sort_direction,
+            page_size=100,
         )
 
     def set_alerts_enabled(self, alerts_enabled=True):
--- a/src/py42/clients/detectionlists/departing_employee.py
+++ b/src/py42/clients/detectionlists/departing_employee.py
@@ -80,8 +80,8 @@ class DepartingEmployeeClient(BaseClient):
 
         Args:
             filter_type (str, optional): Valid filter types. Defaults to None.
-            sort_key (str, optional): Sort results based by field. Defaults to None.
-            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to None.
+            sort_key (str, optional): Sort results based by field. Defaults to "CREATED_AT".
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to "DESC".
             page_num (str or int, optional): The page number to request. Defaults to None.
             page_size (str or int, optional): The items to have per page. Defaults to None.
 
@@ -102,8 +102,8 @@ class DepartingEmployeeClient(BaseClient):
 
         Args:
             filter_type (str, optional): Valid filter types. Defaults to None.
-            sort_key (str, optional): Sort results based by field. Defaults to None.
-            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to None.
+            sort_key (str, optional): Sort results based by field. Defaults to "CREATED_AT".
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to "DESC".
 
         Returns:
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
--- a/src/py42/clients/detectionlists/departing_employee.py
+++ b/src/py42/clients/detectionlists/departing_employee.py
@@ -74,7 +74,7 @@ class DepartingEmployeeClient(BaseClient):
         sort_key=u"CREATED_AT",
         sort_direction=u"DESC",
         page_num=None,
-        page_size=100,
+        page_size=None,
     ):
         """Get a single page of Departing Employees.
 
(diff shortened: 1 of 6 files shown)


src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -68,37 +68,42 @@ def remove(self, user_id):
...
-            u"pgNum": page_num,
-            u"filterType": filter_type,
-            u"srtKey": sort_key,
-            u"srtDirection": sort_direction,
-        }
-        return self._session.post(uri, data=json.dumps(data))
+        Args:
+            filter_type (str, optional): Valid filter types. Defaults to None.
+            sort_key (str, optional): Sort results based by field. Defaults to None.
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to None.
```
kiran-chaudhary, line 84: Defaults to DESC

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -68,37 +68,42 @@ def remove(self, user_id):
...
-            u"pgSize": 100,
-            u"pgNum": page_num,
-            u"filterType": filter_type,
-            u"srtKey": sort_key,
-            u"srtDirection": sort_direction,
-        }
-        return self._session.post(uri, data=json.dumps(data))
+        Args:
+            filter_type (str, optional): Valid filter types. Defaults to None.
+            sort_key (str, optional): Sort results based by field. Defaults to None.
```
kiran-chaudhary, line 83: Defaults to CREATED_AT

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -68,37 +68,42 @@ def remove(self, user_id):
...
         Args:
-            filter_type (str, optional): Filter results by status. Defaults to "OPEN".
-            sort_key (str, optional): Key to sort results on. Options: (``CREATED_AT``,
-                ``DEPARTURE_DATE``, ``DISPLAY_NAME``, ``NUM_EVENTS``, ``TOTAL_BYTES``). Defaults to
-                ``CREATED_AT``.
-            sort_direction (str, optional): Sort direction. Options: (``ASC``, ``DESC``). Defaults
-                to ``DESC``.
+            filter_type (str, optional): Valid filter types. Defaults to None.
+            sort_key (str, optional): Sort results based by field. Defaults to None.
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to None.
```
kiran-chaudhary, line 106: Defaults to DESC and sort_key too.

src/py42/clients/util.py
```diff
@@ -2,12 +2,13 @@
...
 
 def get_all_pages(func, key, *args, **kwargs):
-
-    item_count = settings.items_per_page
     page_num = 0
-    while item_count >= settings.items_per_page:
+    if kwargs.get("page_size") is None:
+        kwargs["page_size"] = settings.items_per_page
+    item_count = kwargs["page_size"]
+    while item_count >= kwargs["page_size"]:
```
kiran-chaudhary, line 9: Something is fishy here, can't understand the intention!!
 `item_count` is set to `page_size` on line 8 then on line 9 while condition would be satisfied on equality condition.  
I think the while loop will always get executed only once, then do we need the while loop?

alanag13: a while loop continues for each iteration that is satisfied, what you're saying would be true if `item_count` was _not_ equal to `kwargs["page_size"].`

`item_count` gets its value changed at the end of this loop to be equal to the number of items that the page returned. In other words, this will loop until we get to page that returns fewer items than `kwargs["page_size"]`, which is what we want.

src/py42/clients/util.py
```diff
@@ -2,12 +2,13 @@
 
 
 def get_all_pages(func, key, *args, **kwargs):
-
-    item_count = settings.items_per_page
     page_num = 0
-    while item_count >= settings.items_per_page:
+    if kwargs.get("page_size") is None:
```
alanag13, line 6:
```python
page_size = kwargs.get("page_size") or settings.items_per_page
item_count = page_size
while item_count >= page_size:
   ...
```

alanag13:
@unparalleled-js I understand that your need for this is probably limited in scope to the detection lists right now, but we will probably eventually want to have this consistently available across all `get_all` methods. Is there a story that captures that?

Otherwise this looks good, I just had one small readability suggestion.

src/py42/clients/util.py
unparalleled-js: The problem is the page size is already part of kwargs and then it ends up getting passed twice to the function below

unparalleled-js: Tests start failing when I make that change - I would have to re-architect other stuff to make that work. Is it ok the way it is?

unparalleled-js: I even tried with line 11 back to the way it was

unparalleled-js:
I can create a story for doing the other ones as well.

unparalleled-js:
integ-1117

src/py42/clients/util.py
alanag13: it's fine the way it is.

unparalleled-js:
@kiran-chaudhary Can you re-review this again?

src/py42/clients/util.py
kiran-chaudhary: Thank you.

src/py42/clients/detectionlists/high_risk_employee.py
```diff
@@ -127,4 +133,27 @@ def get_all(self, filter_type=u"OPEN", sort_key=None, sort_direction=None):
...
+    def _get_high_risk_employees_page(
+        self,
+        tenant_id,
+        filter_type=None,
+        sort_key=None,
+        sort_direction=None,
+        page_num=None,
+        page_size=None,
+    ):
+        # Overwriting page_size since default value 1000 returns error
```
kiran-chaudhary, line 148: This comment should be moved to `get_all`, as that is where we set value to 100.

kiran-chaudhary:
> @unparalleled-js I understand that your need for this is probably limited in scope to the detection lists right now, but we will probably eventually want to have this consistently available across all `get_all` methods. Is there a story that captures that?
> 
> Otherwise this looks good, I just had one small readability suggestion, in addtion to @kiran-chaudhary's docs comments.

Yes, both the private methods `_get_departing_employees_page ` and `_get_high_risk_employees_page ` are exactly identical, except that they are different session objects or different uri. So if possible they can be merged into a common utility function.

unparalleled-js:
Feeling like I was rushing this and that I don't really _need_ it... I see a lot of commands in Cortex XSOAR that offer `pgNum` and `pgSize` right in the command, and Nathan and PAN dev were a little surprised that py42 didn't allow requesting individual pages. That being said, I see a lot of commands also offer a `results` param that either that sets the page num OR pages at max until it reaches that num; it abstracts paging...  I can do that without setting the page num and get the same results for now! And then we can implement this better and fully. How does that sound?

PR #131, opened 2020-06-29
Getting an org by name

unparalleled-js:
Not sure what is going here, but `get_by_name` in orgs does not behave the same as other functions. Probably would be better to just wrap it in a `Py42Response` and give the list a key of `orgs` or something, but that would be a breaking change. At the very, least the documentation needs to be corrected and that is what this PR does, but I am also happy to go to the custom JSON route too.


src/py42/clients/orgs.py
```diff
@@ -63,25 +63,23 @@ def get_by_uid(self, org_uid, **kwargs):
...
+            that each contain a page of organizations with the given name.
         """
         found_orgs = []
         response = self.get_all(**kwargs)
         for page in response:
-            org_list = page[u"orgs"]
-            for org in org_list:
-                if org[u"orgName"] == org_name:
-                    found_orgs.append(org)
-        return found_orgs
```
unparalleled-js, line 84: Generator actually seems like way technically easier, but still not matter what we do here, it is weird.

For the generator method, we won't know the `totalCount` so we have to remove that from the response.
For the custom JSON method, we would have to somehow manually craft a `Py42Response` object or like modify the first page (`Py42Response`) from the `get_all()` call

unparalleled-js:
I don't know the best approach here. I left a comment explaining why this is a weird problem. I can fix the test and add a CL after

src/py42/clients/orgs.py
```diff
@@ -63,25 +63,24 @@ def get_by_uid(self, org_uid, **kwargs):
...
+            page[u"orgs"] = [o for o in page[u"orgs"] if o[u"orgName"] == org_name]
+            if page._data_root.get("totalCount"):
```
unparalleled-js, line 81: I'm not sure the best way to fix totalCount, since we won't know it from the response. This feels hacky though....

alanag13: Im beginning to feel like this method maybe shouldn't exist since it cant be faithfully represented as `Py42Response`, and that user should just be doing what we're doing here on their own if they need this.

unparalleled-js: Going to delete it

PR #132, opened 2020-06-29
Make email param required for `create_user()`

unparalleled-js:
I get 400 responses when leaving off this parameter, so it must be required.

### DIFF
--- a/src/py42/clients/users.py
+++ b/src/py42/clients/users.py
@@ -38,6 +38,8 @@ class UserClient(BaseClient):
         """
 
         uri = u"/api/User"
+        if email is None:
+            email = username
         data = {
             u"orgUid": org_uid,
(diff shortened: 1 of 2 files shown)


src/py42/clients/users.py
```diff
@@ -27,7 +27,7 @@ def create_user(
         Args:
             org_uid (str): The org UID for the organization the new user belongs to.
             username (str): The username for the new user.
-            email (str, optional): The email for the new user. Defaults to None.
+            email (str): The email for the new user.
```
alanag13, line 30: remove `=None` from email in the args list, doing this won't be a breaking change if not including `email` already doesn't work.

unparalleled-js:
We could even make username optional and just set to be the email if they don't provide one

PR #134, opened 2020-07-08
Bugfix/fix errored pds download

alanag13:
See internal detail at https://jira.corp.code42.com/browse/INTEG-1124.

If the `GetFile` request to the storage node fails, today this is resulting a stream that contains an error message from the server being returned instead of properly failing. This appears to have been caused by some differences in http headers.

Also fixes an issue that @timabrmsn found where streams were being loaded completely into memory due to `.text` being called in the `Py42Response` constructor.

To test:
```python
# this hash produces the error in our "partner" account, should result in a raised`Py42Error`.
sdk.securitydata.stream_file_by_sha256("8dc4e9ac5355cb2a87447cdaa9a15badefa027af70c3c8200a2e776b37f993ea")

# also ensure that the normal use case still works.
# Inspect the contents of written file to verify it does not simply contain an error message.
# from the "demo" account
stream = sdk.securitydata.stream_file_by_sha256("098db7aff65b03a60eb784d4005f08325243d8e747f52a98244ff6116171f56d")
with open("testfile.txt", "wb") as test:
    for chunk in stream.iter_content():
        test.write(chunk)
```

### DIFF
--- a/src/py42/modules/securitydata.py
+++ b/src/py42/modules/securitydata.py
@@ -211,10 +211,9 @@ class SecurityModule(object):
         for response in file_generator:
             if response.status_code == 204:
                 continue
-            url = response[u"storageNodeURL"]
             try:
                 storage_node_client = self._microservices_client_factory.create_storage_preservation_client(
-                    url
+                    response[u"storageNodeURL"]
                 )
                 token = storage_node_client.get_download_token(
                     response[u"archiveGuid"], response[u"fileId"], response[u"versionTimestamp"]
@@ -225,7 +224,9 @@ class SecurityModule(object):
                 # 'get_file_location_detail_by_sha256', hence we keep looking until we find a stream
                 # to return
                 debug.logger.warning(
-                    u"Failed to stream file with hash {0} from {1}.".format(checksum, url)
+                    u"Failed to stream file with hash {0}, info: {1}.".format(
+                        checksum, response.text
+                    )
                 )
         raise Py42Error(
             u"No file with hash {0} available for download on any storage node.".format(checksum)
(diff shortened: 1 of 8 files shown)


unparalleled-js:
What about the other stream_from_backup method? Does that have this issue?

alanag13:
@unparalleled-js good question, I could try to web restore the same file that the hash is trying to restore and see.

alanag13:
hmm, the error that was happening on the server went away now and these files are both successfully being downloaded. For what it's worth, I tested this while the error was still reproducible, and it did the right thing.

I think we'll have to hold off on any changes to `stream_file_from_backup` until we have evidence / reason to believe that it is not working properly.

alanag13:
I found a different "bad"' hash, updated the description.

src/py42/modules/securitydata.py
```diff
@@ -202,32 +202,31 @@ def _find_file_versions(self, md5_hash, sha256_hash):
...
-                    u"PDS fetch file version API failed for md5 hash {0} and sha256 hash {1}. "
+                    u"Failed to find any file version for md5 hash {0} / sha256 hash {1}. "
                     u"Error: ".format(md5_hash, sha256_hash),
                     err,
                 )
-                pass
 
     def _stream_file(self, file_generator, checksum):
         for response in file_generator:
+            url = response[u"storageNodeURL"]
```
timabrmsn, line 212: This breaks if the response data isn't a dict, like a `<Py42Response [status=204, data='']>

alanag13: good find, thanks

PR #136, opened 2020-07-10
Add "content" and "data" properties to Py42Response

alanag13:
This addresses two usability problems we have found with `Py42Response`:
- you cannot use functions that are normally found on dicts, such as `get()`.
- you _must_ stream the response using `iter_content()` when downloading a file.

This is also fixes an internal regression that was caused by our other tools refrencing the internal property `_data_root`, which recently had some behavioral changes.


src/py42/response.py
```diff
@@ -80,24 +80,32 @@ def status_code(self):
         return self._response.status_code
 
     def __str__(self):
-        return str(self._data)
+        return str(self._data_root)
 
     def __repr__(self):
-        data = self._data
+        data = self._data_root
```
timabrmsn, line 86: I'm not sure if this would ever really be a problem, but we could put a check here to make sure we're not loading all the streaming data due to this being called:

```
if not self._response._content_consumed:
    data = "<streamed>"
else:
    data = self._data_root
```

alanag13: hmm, I like the idea, but if we have the response but just haven't read `text` or `content` yet, this would always say `streamed`, wouldn't it?

alanag13: I feel like the only way to prevent something like that would to remove `data` from `__repr__`. I would actually be okay with that because it aligns with `requests`, what do you think?

timabrmsn: only if the initial request has `streamed=True` passed will `._content_consumed` be False. So this would only prevent the repr method forcing the full read of the data, but once it is read, then it would display the contents.

alanag13: oh cool, thanks for the clarification.

timabrmsn: The only real use-case this would benefit is if you're playing around with streaming files from py42 in the repl. This would allow you to inspect the response a bit without it blocking on the download.

PR #138, opened 2020-07-16
Improve tox setup and developer onboarding

alanag13:
Ironically, I ultimately decided against using Pipenv.

The main reason for that is that it is aimed at reproducing application builds by using lock files (making sure that two separate builds alway use the same versions of the same dependencies, and 2nd, 3rd, etc generation dependencies). Since this is a library and not an application running in a standalone environment, we have to be flexible and allow for the fact that those dependencies could be installed along side us. More here: https://pipenv.pypa.io/en/latest/advanced/#pipfile-vs-setup-py

This instead contains some other useful changes:
- vscode and pycharm config files so that testing tools, etc are set up right out of the box
- dev dependencies do not require `pre-commit` to be installed, it is not set up yet but this and linters will instead be run by `tox`, so that you can run them whenever you want, and they will be checked during ci. You will of course still have the option to install them the old way.
- dev dependencies do not install things needed for docs and this is now handled by tox (you can install those dependencies manually if you really want to run the docs locally but this isn't the typical use case)
- tox now will test that the docs build successfully on every PR
- reduced line length for black to default of 88
- improved environment setup guide to include steps for creating a virtualenv


.github/workflows/build.yml
```diff
@@ -24,9 +24,11 @@ jobs:
...
       - name: Install tox
         run: pip install tox==3.14.3
-      - name: Run Tox
+      - name: Run Unit tests
         run: tox -e py  # Run tox using the version of Python in `PATH`
       - name: Submit coverage report
         uses: codecov/codecov-action@v1.0.7
         with:
           file: ./coverage.xml
+      - name: Build docs
```
alanag13, line 33: PRs will now make sure the docs build don't fail

.idea/py42.iml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alanag13, line 1: This config file helps make sure google commands and pytest are setup for pycharm out of the box

.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
alanag13, line 1: pytest will be set up out of the box in vscode

CONTRIBUTING.md
```diff
@@ -1,25 +1,49 @@
...
+```
+
+After installing `pyenv` and `pyenv-virtualenv`, be sure to add the following entries to your `.zshrc` (or `.bashrc` if you are using bash) and restart your shell:
+
+```bash
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
 ```
 
-If you are using `zsh`, you may need to escape the brackets.
```
alanag13, line 12: I just haven't found this to be true

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -15,15 +15,15 @@ def __init__(self, session, user_context, detection_list_user_client):
         self._detection_list_user_client = detection_list_user_client
 
     def add(self, user_id, departure_date=None):
-        """Adds a user to the Departing Employees list. Creates a detection list user profile if one
-         didn't already exist.
+        """Adds a user to the Departing Employees list. Creates a detection list user profile if one \
```
alanag13, line 18: I made it so that documentation build warnings are now errors, which prompted these changes.

tox.ini
```diff
@@ -1,49 +1,32 @@
...
-deps = coverage
-skip_install = true
-commands = coverage erase
+commands = sphinx-build -W -b html -d {envtmpdir}/doctress docs {envtmpdir}/html
 
-[testenv:lint27]
-basepython = python2.7
-deps = pylint==1.9.5
-commands = pylint -E py42
+#[testenv:style]
```
alanag13, line 29: this is where black and flake8 checks will be added

setup.py
```diff
@@ -15,32 +15,27 @@
...
-    python_requires=">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, <4",
-    install_requires=["requests>=2.3"],
-    license="MIT",
     url="https://github.com/code42/py42",
     project_urls={
         "Issue Tracker": "https://github.com/code42/py42/issues",
         "Documentation": "https://py42docs.code42.com/",
         "Source Code": "https://github.com/code42/py42",
     },
+    description="The Official Code42 Python API Client",
```
unparalleled-js, line 24: The word "official" may be superfluous. I believe it was removed from
the py42docs site too.

PR #139, opened 2020-07-17
Chore/linter

alanag13:
some notes here:
- no logic has changed,  the formatter and flake8 rules touched a lot of files
- I mostly would like this review to be of the `CONTRIBUTING` document to make sure it all makes sense
- one of the formatters that auto-sorts imports also breaks up `from` imports into multiple lines, and I think I would rather have this be `import module as m` and then use `m.do_the_thing(), m.do_the_other_thing()` instead of `from module import do_the_thing, do_the_other_thing`, but i'll worry about that in a separate PR.


setup.cfg
```diff
@@ -0,0 +1,34 @@
...
+    # line length, handled by black
+    B950
+    E501
+    # bare except, handled by bugbear B001
+    E722
+    # binary operation line break, different opinion from black
+    W503
+# up to 88 allowed by bugbear B950
+max-line-length = 80
+per-file-ignores =
```
alanag13, line 30: this where you can add exemptions to flake8 rules, the filter ones i'd like to address longer term but it might involve a breaking change

unparalleled-js:
I'd like a Table of Contents in CONTRIBUTING in case I come back and need to remember how to run the docs or something - to make it a faster lookup

CONTRIBUTING.md
```diff
@@ -39,42 +39,62 @@ Next, with your virtual environment activated, install py42 and its development
...
+pip install -e .[dev]
 ```
 
 Open the project in your IDE of choice and change the python environment to
 point to your virtual environment, and you should be ready to go!
 
-## General
+## Run a full build
+
+We use [tox](https://tox.readthedocs.io/en/latest/#) to run our build against Python 2.7, 3.5, 3.6, 3.7, and 3.8. When run locally, `tox` will run everything against the version of python that your virtual envrionment is running. When you open a PR, this process will be run against _all_ supported python versions.
```
unparalleled-js, line 50: It might be worth separating out what happens locally versus what happens during a PR. We could have a section about PRs / the Review process which we will probably add to later

unparalleled-js:
It might be worth to mention licensing somehow - like an implicit CLA.

CONTRIBUTING.md
unparalleled-js: On that note, also mentioning the branch name expectancies... even if we don't need a branch name style, it makes us seem more accepting to PRs. It could be like " We commend putting prefixing the PR branch name with chore, feature, or  bugfix."

timabrmsn:
I know this hasn't changed in this PR, but our setup instructions (the `pyenv` part) are Mac-only. We should probably have a generic "create a virtual env and activate it" section that Windows/Linux users can use too. 

Something like: 

```
install python3.6+

run `python3 -m venv py42`

activate your env
```

alanag13:
> It might be worth to mention licensing somehow - like an implicit CLA.

We should eventually aim to do this in a more official sense, I don't think we should add anything without getting legal, etc involved.

alanag13:
@timabrmsn @unparalleled-js thanks for the feedback, this is ready again

CONTRIBUTING.md
```diff
@@ -27,54 +46,91 @@ eval "$(pyenv virtualenv-init -)"
...
+
+Install a version of python 3.6 or higher from [python.org](https://python.org).
+Next, in a directory somewhere outside the project, create and activate your virtual environment:
+
+```bash
+python -m venv py42
+# macOS/Linux
+source py42/bin/activate
+# Windows
+.\env\Scripts\Activate
```
timabrmsn, line 66: should be `.\py42\Scripts\Activate`

timabrmsn:
Looks good apart from the Windows env folder name.

CONTRIBUTING.md
alanag13: ah thanks

PR #143, opened 2020-07-20
Feature/roles functions

### LINKED ISSUE #142: [Enhancement] Add functions to enable role assignment
## Summary
Currently, py42 is able to get a list of all users with a given role id using `sdk.users.get_all(role_id=<num>)`. 
However, it does not currently have a way to actually _get_ the role ids, or do other useful functions related to roles.

Allow py42 to:
- retrieve of all roles in the system
- add a given role to a user
- remove a given role from a user

## Proposed API

```python
sdk.users.get_available_roles()
sdk.users.get_roles(user_id)
sdk.users.add_role(user_id, role_name)
sdk.users.remove_role(user_id, role_name)
```

## Intended Use Case
Programmatic user role management. Many outside factors can determine whether user should be assigned to or removed from a role, and the ability to do this via automation can save a lot of manual effort.

alanag13:
### Description of Change ###
Adds functions to enable managing user roles

### Issues Resolved ###

- closes #142 

### Testing Procedure ###
- call the functions as described in #142.
- use adminstration -> users -> selected a user ->edit to view a user's roles and ensure that they are added and removed as expected.

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- Functions for managing role assignment:
+    - `sdk.users.get_available_roles()`
+    - `sdk.users.get_roles()`
+    - `sdk.users.add_role()`
+    - `sdk.users.remove_role()`
+
 ## 1.6.2 - 2020-07-10
 
 ### Added
--- a/src/py42/clients/users.py
+++ b/src/py42/clients/users.py
@@ -248,7 +248,7 @@ class UserClient(BaseClient):
     def get_available_roles(self):
         """Report the list of roles that are available for the authenticated user to
         assign to other users.
-        `REST Documentation <https://console.us.code42.com/apidocviewer/#UserMoveProcess-post>`__
+        `V4 REST Documentation <https://console.us.code42.com/swagger/#/role/Role_View>`__
 
         Returns:
             :class:`py42.response.Py42Response`
--- a/src/py42/clients/users.py
+++ b/src/py42/clients/users.py
@@ -244,3 +244,58 @@ class UserClient(BaseClient):
         uri = u"/api/UserMoveProcess"
         data = {u"userId": user_id, u"parentOrgId": org_id}
         return self._session.post(uri, data=json.dumps(data))
+
+    def get_available_roles(self):
+        """Report the list of roles that are available for the authenticated user to
+        assign to other users.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#UserMoveProcess-post>`__
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        uri = u"/api/v4/role/view"
+        return self._session.get(uri)
+
+    def get_roles(self, user_id):
+        """Return the list of roles that are currently assigned to the given user.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#UserRole-get>`__
+
+        Args:
+            user_id (int): An ID for a user.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        uri = u"/api/UserRole/{}".format(user_id)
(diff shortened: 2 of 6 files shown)


src/py42/clients/users.py
```diff
@@ -244,3 +244,57 @@ def change_org_assignment(self, user_id, org_id):
...
+        REST Documentation <https://console.us.code42.com/apidocviewer/#UserRole-delete>`__
+
+        Args:
+            user_id (int): An ID for a user.
+            role_name (str): The name of the role to unassign from the user.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        role_name = role_name.replace(" ", "%20")
```
alanag13, line 298: I would normally use `params` here, but that replaces whitespace with `+` instead of `%20`, and would give errors when passed a role that contained a space.

CHANGELOG.md
```diff
@@ -8,6 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- Functions for managing role assignment:
```
unparalleled-js, line 15: Is there a benefit to doing `sdk.users.roles.get()`, `sdk.users.roles.add()` etc instead? Just making sure you considered it is all.

unparalleled-js: or event `sdk.roles`

src/py42/clients/users.py
unparalleled-js: Interesting. Should we just call a `urlencode` method instead?

unparalleled-js:
```
>>> urllib.parse.quote("hi there")
'hi%20there'

```

unparalleled-js: Also, a comment explaining might prevent future devs from changing it without knowing why

src/py42/clients/users.py
```diff
@@ -244,3 +244,57 @@ def change_org_assignment(self, user_id, org_id):
...
+        """
+        uri = u"/api/v4/role/view"
+        return self._session.get(uri)
+
+    def get_roles(self, user_id):
+        """Return the list of roles that are currently assigned to the given user.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#UserRole-get>`__
+
+        Args:
+            user_id (int): An ID for a user.
```
unparalleled-js, line 264: I think at some point we should make these all be `int | str` since I think it works with both and this way you don't have to do unnecessary transformations if they arise

src/py42/clients/users.py
```diff
@@ -244,3 +244,57 @@ def change_org_assignment(self, user_id, org_id):
         uri = u"/api/UserMoveProcess"
         data = {u"userId": user_id, u"parentOrgId": org_id}
         return self._session.post(uri, data=json.dumps(data))
+
+    def get_available_roles(self):
+        """Report the list of roles that are available for the authenticated user to
+        assign to other users.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#UserMoveProcess-post>`__
```
timabrmsn, line 251: Probably don't want to link to UserMoveProcess api.

src/py42/clients/users.py
alanag13: this is a far better option for sure

src/py42/clients/users.py
alanag13: woop

CHANGELOG.md
alanag13: conceptually, at least in the UI, user are added to roles and not the other way around, which is why I felt these functions should be associated with users. I think we also discussed this as a team during a grooming meeting.

I can see the argument for `users.roles`, the main reason I didn't go this way is simply because I didn't feel that roles would ever have enough functions to warrant having its own "bucket". @timabrmsn do you have thoughts here?

timabrmsn: I know there are planned changes to roles/permissions coming (PRM-2940), though I haven't looked through them closely enough to know how impactful that would be to this design.

I think for what we have currently, it makes sense to keep it directly under users.

alanag13:
@timabrmsn @unparalleled-js your comments should be addressed!

PR #145, opened 2020-07-21
Feature/add eq and contains methods to filter groups

timabrmsn:
### Description of Change ###

This change adds `__eq__` and `__hash__` methods to the `QueryFilter` class and `__eq__` and `__contains__` methods to the `FilterGroup` class, which simplifies comparing filters for equality and membership. It also changes the `__str__` and `__iter__` implementations to return deterministic results for equivalent queries by using the new `FilterGroup._filter_set` property, which removes duplicate filters and sorts them. 

For `QueryFilter` the comparison checks work between any combination of the classes themselves, the string representation of the filter, or the `__iter__` representation of the filter if they are converted to a tuple or list.

For `FilterGroup` you can only compare equality between two classes themselves, but you can check to see if a given `QueryFilter` is in the `FilterGroup`:

### PR Checklist ###


timabrmsn:
Some examples:

Arg order doesn't matter:
```
>>> FileName.is_in(["test.pdf", "test.txt"]) == FileName.is_in(["test.txt", "test.pdf"])
True
```
Duplicates don't break equality:
```
>>> FileName.is_in(["test.pdf", "test.txt"]) == FileName.is_in(["test.txt", "test.pdf", "test.pdf"])
True
```
Can compare string representation with class directly:
```
>>> str(FileName.is_in(["test.pdf", "test.txt"])) == FileName.is_in(["test.txt", "test.pdf", "test.pdf"])
True
```
Can check if a given filter is in a FilterGroup:
```
>>> QueryFilter("fileName", "IS", "test.txt") in FileName.is_in(["test.txt", "test.pdf"])
True
```

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.users.add_role()`
     - `sdk.users.remove_role()`
 
+- `__eq__` and `__hash__` methods to the `py42.sdk.queries.query_filter.QueryFilter` class to enable easier comparison of filters
```
unparalleled-js, line 21: We should also mention the sorted change here, no?  Results that were not sorted before will now be sorted. This could break some unit tests. Also, would let users know they don't have to manually sort them if they were before.

alanag13: yes, in fact this change caused our own unit tests to need to be changed, so it should be called out.

alanag13:
> removes duplicate filters and sorts them

@timabrmsn can there be a test for this? Also to make sure my understanding is correct -- filters are only "duplicate" if their term, operation, and value are all identical, correct?

unparalleled-js:
I love this!

> Some examples:
> 
> Arg order doesn't matter:
> 
> ```
> >>> FileName.is_in(["test.pdf", "test.txt"]) == FileName.is_in(["test.txt", "test.pdf"])
> True
> ```
> 
> Duplicates don't break equality:
> 
> ```
> >>> FileName.is_in(["test.pdf", "test.txt"]) == FileName.is_in(["test.txt", "test.pdf", "test.pdf"])
> True
> ```
> 
> Can compare string representation with class directly:
> 
> ```
> >>> str(FileName.is_in(["test.pdf", "test.txt"])) == FileName.is_in(["test.txt", "test.pdf", "test.pdf"])
> True
> ```
> 
> Can check if a given filter is in a FilterGroup:
> 
> ```
> >>> QueryFilter("fileName", "IS", "test.txt") in FileName.is_in(["test.txt", "test.pdf"])
> True
> ```

Love this

CHANGELOG.md
timabrmsn: good call, I'll note it

timabrmsn:
> > removes duplicate filters and sorts them
> 
> @timabrmsn can there be a test for this? Also to make sure my understanding is correct -- filters are only "duplicate" if their term, operation, and value are all identical, correct?

Yep, I did add tests for these. And correct, the `__hash__` method operates on the str representation of the filter, so everything would have to be the same for them to be identified as duplicates.

alanag13:
@timabrmsn LGTM

PR #150, opened 2020-07-23
Chore/doc filter create methods

unparalleled-js:
### Description of Change ###

* Adds docs to method for creating filter groups.
* Corrects errors from some other docs.
* Updates a guide on how to craft filters from raw JSON

### Issues Resolved ###
fixes INTEG-1134

### Testing Procedure ###

Run the docs; verify they look as expected and the links works.

### PR Checklist ###


.gitignore
```diff
@@ -9,6 +9,9 @@ __pycache__/
 # C extensions
 *.so
 
+# IDE files
```
unparalleled-js, line 12: I could not get the .gitignore to work from inside `.idea`... Not sure why, but other examples on I have seen just put it here.

src/py42/sdk/queries/fileevents/file_event_query.py
```diff
@@ -8,18 +8,18 @@
 class FileEventQuery(BaseQuery):
     """Helper class for building Code42 Forensic Search queries.
 
-    A FileEventQuery instance's ``all()`` and ``any()`` take one or more :class:`FilterGroup`
-    objects to construct a query that can be passed to the :meth:`FileEventClient.search()` method.
-    ``all()`` returns results that match all of the provided filter criteria, ``any()`` will return
-    results that match any of the filters.
+    A FileEventQuery instance's ``all()`` and ``any()`` take one or more
+    :class:`~py42.sdk.queries.query_filter.FilterGroup` objects to construct a query that
+    can be passed to the :meth:`FileEventClient.search()` method. ``all()`` returns results
```
unparalleled-js, line 13: Note that this link does not actually work because it is referencing an internal client. Still, I like that it's there because it tells me, the developer, that it's a method at least, and in a form that I am used to.

PR #151, opened 2020-07-23
Bugfix/fix md5 download error

alanag13:
### Description of Change ###

Corrects `sdk.securitydata.stream_file_by_md5` to raise the correct error when no file with the given checksum is found.

Also slightly adjusts the verbiage for `stream_file_by_sha256` when the hash is not found.

### Issues Resolved ###

- [INTEG-1138](https://jira.corp.code42.com/browse/INTEG-1138)

### Testing Procedure ###
call `sdk.securitydata.stream_file_by_md5` and `sdk.securitydata.stream_file_by_sha256` with an invalid hash (any random string will do).

The exception raised should be a `Py42Error` with these messages, respectively:
- `No files found with md5 checksum <your input here>`
- `No files found with sha256 checksum <your input here>`

### PR Checklist ###


tests/modules/test_securitydata.py
```diff
@@ -823,7 +826,8 @@ def test_stream_file_by_sha256_raises_py42_error_when_find_file_versions_returns
 
         with pytest.raises(Py42Error) as e:
             security_module.stream_file_by_sha256("shahash")
-            assert e.value.args[0] == PDS_EXCEPTION_MESSAGE.format("shahash")
+
+        assert e.value.args[0] == PDS_EXCEPTION_MESSAGE.format("shahash")
```
alanag13, line 830: I fixed some unrelated unit tests -- `with pytest.raises()` exits the `with` block as soon as an exception is raised so some assertions were not actually being checked

alanag13:
test failures are due to tests I "fixed", but I will address them

alanag13:
one of those tests had to be removed because it was invalid -- was trying to test that a property that did not exist was set.

src/py42/modules/securitydata.py
```diff
@@ -276,7 +273,8 @@ def stream_file_by_md5(self, checksum):
         """
         events = self._search_by_hash(checksum, MD5)
         if not len(events):
-            raise Py42ArchiveFileNotFoundError(checksum, "")
+            message = u"No files found with md5 checksum {}".format(checksum)
+            raise Py42Error(message)
```
unparalleled-js, line 277: Would it be better to make a custom Exception class for this so we can easily catch it in certain circumstances?  Like a `Py42UnavailableFileError` that takes a hash type that was used to search for, and it's value

tests/modules/test_securitydata.py
unparalleled-js: Doesn't it still exit before execution this way? I thought that we have to do this in try/except to get it to work

src/py42/modules/securitydata.py
alanag13: `Py42HashUnavailableError`?

tests/modules/test_securitydata.py
alanag13: Nope, these lines definitely get hit, I ensured this with breakpoints and also by changing them to intentionally cause test failures.

src/py42/modules/securitydata.py
unparalleled-js: Makes sense to me!

unparalleled-js: `Py42HashNotFoundError` would mirror `Py42ArchiveFileNotFoundError`

alanag13: done!

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Fixed
+
+- `sdk.securitydata.stream_file_by_md5()` now correctly raises `Py42ChecksumNotFoundError` when no matching md5 is found (previously was `Py42ArchiveFileNotFoundError`).
```
unparalleled-js, line 15: This doesn't include the change in SHA256, how that now returns this new error type rather than just a Py42Error

alanag13: done, thanks!

src/py42/modules/securitydata.py
```diff
@@ -257,10 +257,7 @@ def stream_file_by_sha256(self, checksum):
         """
         events = self._search_by_hash(checksum, SHA256)
         if not len(events):
-            message = u"File not found in archive with sha256 checksum {}".format(
-                checksum
-            )
-            raise Py42Error(message)
+            raise Py42ChecksumNotFoundError("SHA256", checksum)
```
unparalleled-js, line 260: I know it doesn't matter here - but these strings don't have `u` prefixes

tests/modules/test_securitydata.py
```diff
@@ -750,7 +751,7 @@ def test_stream_file_by_sha256_returns_stream_of_file(
         response = security_module.stream_file_by_sha256("shahash")
         assert response == b"stream"
 
-    def test_stream_file_by_sha256_raises_file_not_found_error_when_search_returns_empty_response(
+    def test_stream_file_by_sha256_raises_py42_error_when_search_returns_empty_response(
```
unparalleled-js, line 754: Should this say `raises_py42_checksum_not_found_error` ?

tests/modules/test_securitydata.py
```diff
@@ -919,9 +925,11 @@ def test_stream_file_by_md5_raises_file_not_found_error_when_search_returns_empt
         file_event_search.text = "{}"
         file_event_client = mocker.MagicMock(spec=FileEventClient)
         file_event_client.search.return_value = file_event_search
-        with pytest.raises(Py42Error):
```
unparalleled-js, line 922: This test name hasn't changed in this PR, still says `file_not_found_error`

alanag13: thanks, done.

tests/modules/test_securitydata.py
```diff
@@ -919,9 +925,11 @@ def test_stream_file_by_md5_raises_file_not_found_error_when_search_returns_empt
...
+        with pytest.raises(Py42ChecksumNotFoundError) as e:
             security_module.stream_file_by_md5("md5hash")
 
+        assert u"No files found with MD5 checksum" in e.value.args[0]
+
     def test_stream_file_by_md5_raises_file_not_found_error_when_file_location_returns_empty_response(
```
unparalleled-js, line 933: hmmm... looking closer, are the resto f the tests in this module also supposed to testing for `Py42ChecksumNotFoundError`? They are all still checking the base class

alanag13: that's a different error that is raised under a different circumstance which actually _is_ just a `Py42Error`.
It's debatable whether or not that could also be made into a more specific error too, but I'd rather not have that scope creep into this PR.

PR #152, opened 2020-07-24
Return py42 response on all writes

unparalleled-js:
### Description of Change ###

- functions now return `Py42Response` objects as expected:
    - `sdk.detectionlists.update_user_notes()`
    - `sdk.detectionlists.add_user_risk_tags()`
    - `sdk.detectionlists.remove_user_risk_tags()`
    - `sdk.detectionlists.add_user_cloud_alias()`
    - `sdk.detectionlists.remove_user_cloud_alias()`
- `sdk.archive.get_all_org_cold_storage_archives()` now actually uses parameters `include_child_orgs`, `sort_key` and `sort_dir`

Correct other, minor documentation strings too.

### Issues Resolved ###

- closes #INTEG-1091

### Testing Procedure ###
* Call any of the methods (listed in CHANGELOG) and make sure you get a Py42Response object as a response. Before, it was just None.

### PR Checklist ###


src/py42/modules/archive.py
```diff
@@ -146,4 +146,9 @@ def get_all_org_cold_storage_archives(
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
             that each contain a page of cold storage archive information.
         """
-        return self._archive_client.get_all_org_cold_storage_archives(org_id)
+        return self._archive_client.get_all_org_cold_storage_archives(
```
unparalleled-js, line 149: The params did nothing before

tests/modules/test_detectionlists.py
```diff
@@ -75,7 +75,17 @@ def test_update_user_notes_calls_user_client_with_expected_values(
...
-    def test_add_user_risk_tag_calls_user_client_with_expected_values(
+    def test_update_user_notes_returns_response(
+        self, mock_microservice_client_factory, mock_detection_list_user_client
+    ):
+        mock_microservice_client_factory.get_detection_list_user_client.return_value = (
+            mock_detection_list_user_client
+        )
+        module = DetectionListsModule(mock_microservice_client_factory)
+        response = module.update_user_notes(TEST_USER_ID, "newnotes")
+        assert response
```
unparalleled-js, line 86: type is MagicMock in this case, but before was None, so these tests catch the error

unparalleled-js:
Also, I looked for other methods like this but couldn't find any

tests/modules/test_detectionlists.py
```diff
@@ -123,6 +163,16 @@ def test_remove_user_cloud_alias_calls_user_client_with_expected_values(
...
 
+    def test_remove_user_cloud_alias_returns_response(
+        self, mock_microservice_client_factory, mock_detection_list_user_client
+    ):
+        mock_microservice_client_factory.get_detection_list_user_client.return_value = (
+            mock_detection_list_user_client
+        )
+        module = DetectionListsModule(mock_microservice_client_factory)
+        response = module.remove_user_cloud_alias(TEST_USER_ID, "oldalias")
+        assert response
```
alanag13, line 174: adopting this approach for testing each call we make in general will probably be a good idea going forward.

unparalleled-js: If you want, I add them all now! But that would be a lot

src/py42/modules/archive.py
alanag13: good find, this was definitely my fault :-/

tests/modules/test_detectionlists.py
alanag13: no no, wasn't suggesting that, just putting thoughts out there

alanag13:
I want to retest `get_all_org_cold_storage_archives()` for safety's sake, but this otherwise looks good, will approve once that's confirmed as good.

PR #156, opened 2020-07-28
Feature/expose getting single page methods

unparalleled-js:
### Description of Change ###

- Methods for getting individual response pages:
    - `sdk.detectionlists.departing_employee.get_page()`
    - `sdk.detectionlists.high_risk.get_page()`
    - `sdk.users.get_page()`
    - `sdk.devices.get_page()`
    - `sdk.legalhold.get_matters_page()`
    - `sdk.legalhold.get_custodians_page()`
    - `sdk.alerts.get_page()`

### Issues Resolved ###
INTEG-1117

### Testing Procedure ###
Call the methods above

### PR Checklist ###


src/py42/modules/alertrules.py
```diff
@@ -70,12 +73,34 @@ def remove_all_users(self, rule_id):
         rules_client = self.microservice_client_factory.get_alert_rules_client()
         return rules_client.remove_all_users(rule_id)
 
+    def get_page(
+        self, sort_key=u"CreatedAt", sort_direction=u"DESC", page_num=1, page_size=None
+    ):
+        """Gets a page of alerts.
```
kiran-chaudhary, line 79: Should it be "Gets a page of alert rules."?

tests/clients/test_alerts.py
```diff
@@ -257,19 +257,6 @@ def test_reopen_when_given_single_alert_id_posts_expected_data(
             and post_data["alertIds"][0] == "ALERT_ID_1"
         )
 
-    def test_reopen_when_given_tenant_id_posts_expected_data(
```
unparalleled-js, line 260: duplicate test

src/py42/_internal/clients/alerts.py
```diff
@@ -68,12 +82,17 @@ def _get_alert_rules(
         page_num=None,
         page_size=None,
     ):
+        # This method exists separately from `get_page()` because
+        # of the tenant ID parameter - trying to avoid it and avoid duplicate calls
+        # to retrieve it.
+
+        # This API expects the first page to start with zero.
+        page_num = page_num - 1
```
timabrmsn, line 90: We should probably document that this is being done under the hood. If someone uses the now public method and knows the API starts at 0, their first request will be for page -1.

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -72,53 +73,83 @@ def remove(self, user_id):
...
-            u"tenantId": tenant_id,
-            u"pgSize": 100,
-            u"pgNum": page_num,
-            u"filterType": filter_type,
-            u"srtKey": sort_key,
-            u"srtDirection": sort_direction,
-        }
-        return self._session.post(uri, data=json.dumps(data))
+        Args:
+            filter_type (str, optional): Valid filter types. Defaults to None.
```
alanag13, line 87: What are the valid filter types?

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -72,53 +73,83 @@ def remove(self, user_id):
...
+            sort_key (str, optional): Sort results based by field. Defaults to "CREATED_AT".
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to "DESC".
+            page_num (str or int, optional): The page number to request. Defaults to 1.
+            page_size (str or int, optional): The items to have per page. Defaults to 100.
```
alanag13, line 91: `The number of departing employees to return per page`

src/py42/clients/detectionlists/departing_employee.py
```diff
@@ -72,53 +73,83 @@ def remove(self, user_id):
...
+            sort_key (str, optional): Sort results based by field. Defaults to "CREATED_AT".
+            sort_direction (str. optional): ``ASC`` or ``DESC``. Defaults to "DESC".
 
         Returns:
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
             that each contain a page of departing employees.
         """
         return get_all_pages(
-            self._get_departing_employees_page,
+            self._get_page,
```
alanag13, line 121: you could use `self.get_page` here, and then you wouldn't have to include the code for gettting the tenant id.

src/py42/clients/legalhold.py
```diff
@@ -105,16 +106,37 @@ def get_matter_by_uid(self, legal_hold_uid):
...
+        """Gets a page of existing Legal Hold Matters.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#LegalHold-get>`__
+
+        Args:
+            creator_user_uid (str, optional): Find Matters by the identifier of the user who created
+                them. Defaults to None.
+            active (bool or None, optional): Find Matters by their active state. True returns
+                active Matters, False returns inactive Matters, None returns all Matters regardless
+                of state. Defaults to True.
+            name (str, optional): Find Matters with a 'name' that either equals or partially
```
alanag13, line 127: probably just `contains`, rather than `partially contains`.

src/py42/clients/legalhold.py
```diff
@@ -105,16 +106,37 @@ def get_matter_by_uid(self, legal_hold_uid):
...
+                contains this value. Defaults to None.
+            hold_ext_ref (str, optional): Find Matters having a matching external reference field.
+                Defaults to None.
+            page_num (int, optional): The page number to request. Defaults to 0.
```
alanag13, line 131: `Defaults to 1.`

src/py42/clients/legalhold.py
```diff
@@ -105,16 +106,37 @@ def get_matter_by_uid(self, legal_hold_uid):
...
+            page_size (int, optional): The size of the page. Defaults to `py42.settings.items_per_page`.
```
alanag13, line 132: `The number of legal hold matters to return per page`

src/py42/clients/util.py
```diff
@@ -2,14 +2,14 @@
...
-        response = func(
-            *args, page_num=page_num, page_size=settings.items_per_page, **kwargs
-        )
+    if kwargs.get("page_num") is None:
+        kwargs["page_num"] = 0
+    if kwargs.get("page_size") is None:
+        kwargs["page_size"] = settings.items_per_page
+    item_count = kwargs["page_size"]
+    while item_count >= kwargs["page_size"]:
+        kwargs["page_num"] += 1
```
alanag13, line 11: is this right? if `get_all_pages` receives a `page_num` of 1, it will increase it to 2 before it makes its first request.

src/py42/_internal/clients/alerts.py
```diff
@@ -68,12 +82,17 @@ def _get_alert_rules(
         page_num=None,
         page_size=None,
     ):
+        # This method exists separately from `get_page()` because
+        # of the tenant ID parameter - trying to avoid it and avoid duplicate calls
```
timabrmsn, line 86: `get_current_tenant_id()` will only make a single request, then store the tenant_id on the class for future lookups. So I don't think the duplication of these methods is worth saving a few attribute lookups.

src/py42/clients/util.py
alanag13: I'm not sure I understand why this function was changed at all, can you explain?

src/py42/_internal/clients/alerts.py
alanag13: I forgot that implementation detail, this is a good callout.

src/py42/clients/detectionlists/departing_employee.py
alanag13: see @timabrmsn's note re: how extra calls to get the tenant id don't really incur a penalty

alanag13:
@unparalleled-js  @timabrmsn one thought I had: rather than adding separate methods to get single pages, would it be better to have the existing generators have optional `start_page` and `end_page` args (which you could set to be the same if you only wanted a single page)?

src/py42/clients/util.py
unparalleled-js: Before, in detection lists' paging, the actual parameter on the `func` arg for `page_size` did nothing. This is because it is hardcoded to `100` inside the function. If you _did_, however, try to use the `page_size` parameter on `func` from detectionlists, it still wouldn't matter. This is because this util method here, `get_all_pages()`, uses a `page_size` of `settings.items_per_page` and I would see cases where it would use that over anything from `kwargs`.  The same goes for `page_num`. It would _always_ start at zero unless hardcoded in the actual `func`, like it is in alerts.  BUT, to create a `get_page()` method, we need it to actually be a parameter. Else, it always gets the first page, or in the case of alerts, gets the second page.

unparalleled-js: You are right about your first comment though, weird because I was testing this before and didn't notice it but now I do

unparalleled-js:
> @unparalleled-js @timabrmsn one thought I had: rather than adding separate methods to get single pages, would it be better to have the existing generators have optional `start_page` and `end_page` args (which you could set to be the same if you only wanted a single page)?
> 
> `start_page=2, end_page=5` - generator for pages 2-5, inclusive
> `start_page=None, end_page=5` - generator for first page through page 5, inclusive
> `start_page=2, end_page=None` - generator for page 2 to last page, inclusive

So, to do paging, I would have to iterate using two iterators, like `start_page=1, end_page=1`, then to get the next page, I'd have to do `start_page=2, end_page=2`. Do-able I guess, but I just want to be able to request an individual page at a time, preferably not as a generator

src/py42/clients/util.py
unparalleled-js: Trying to make the `page_size` passed twice error not occur... it is passed in from `detectionlists.get_all()` in the kwargs as well as in `get_all_pages()` hardcoded

src/py42/_internal/clients/alerts.py
unparalleled-js: I will put it in the moduel

src/py42/clients/util.py
alanag13: @unparalleled-js something seems still not right here -- we are alway increasing `page_num` by one before the first request gets made, won't this result in the first page always being skipped? Shouldn't we change this so that `response = func(*args, **kwargs)` happens first and _then_ the page num gets increased?

alanag13:
@unparalleled-js is there a reason orgs isn't getting this functionality as well?

unparalleled-js:
> @unparalleled-js is there a reason orgs isn't getting this functionality as well?

No, I would like orgs for sure, so I will put that in, thanks

Sorry! I had accidentally edited your comment when I meant to reply before

src/py42/clients/util.py
unparalleled-js: Maybe you are not seeing it, but it has changed back to it's original impl as far as page num goes. Page size is that only change to this function now, and I have tested all the methods

unparalleled-js: Note that technically, we have _always_ been increasing page num by one before the first request. Look at the main branch.

timabrmsn:
I think I'd prefer keeping two methods `get_page` and `get_all` to keep the api simple (calling `get_all` with a start and end of the same page, to get a generator that you then have to call `next()` on is not how I'd want to get one page's results).

I do like the idea of adding a `start_page` param to the `get_all` methods, that allows us to move the logic for whether to start at 0 or 1 to each caller.

I don't think `end_page` would be necessary though, since the result is a generator, if you want a subset of results you can just stop iterating after your desired number.

alanag13:
> I think I'd prefer keeping two methods `get_page` and `get_all` to keep the api simple (calling `get_all` with a start and end of the same page, to get a generator that you then have to call `next()` on is not how I'd want to get one page's results).
> 
> I do like the idea of adding a `start_page` param to the `get_all` methods, that allows us to move the logic for whether to start at 0 or 1 to each caller.
> 
> I don't think `end_page` would be necessary though, since the result is a generator, if you want a subset of results you can just stop iterating after your desired number.

we discussed offline and came to similar conclusions, though we haven't added a `start_page` param at this time.

PR #159, opened 2020-07-31
Feature/more exceptions

unparalleled-js:
### Description of Change ###

    - `Py42UserAlreadyAddedError`
    - `Py42UserNotInLegalHoldError`
    - `Py42LegalHoldNotFoundOrPermissionDeniedError`
    - `Py42UserDoesNotExistError`
    - `Py42InvalidRuleTypeError`

And their logic mirrored from what is in the CLI.
The CLI tests still pass as well.

### Issues Resolved ###
INTEG-1072

### Testing Procedure ###
Verify these statements:
- `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when the user is already on the list.
- `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when the user already on the list.
- `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when the user is already on the matter.
- `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldNotFoundOrPermissionDeniedError` when the matter is unavailable.
- `py42.users.get_by_username()` now raises `Py42UserDoesNotExistError`.
- `py42.alerts.rules.add_user()` now raises `Py42InvalidRuleTypeError` when trying to add a user to a system rule.
- `py42.alerts.rules.remove_user()` now raises `Py42InvalidRuleTypeError` when trying to a remove user from a system rule.

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `Py42ResponseError`
     - `Py42UserAlreadyAddedError`
     - `Py42UserNotInLegalHoldError`
-    - `Py42LegalHoldNotFoundError`
-    - `Py42PermissionDeniedError`
+    - `Py42LegalHoldNotFoundOrPermissionDeniedError`
     - `Py42UserDoesNotExistError`
     - `Py42InvalidRuleOperationError`
 
@@ -71,8 +70,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 - `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when the user is already on the list.
 - `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when the user already on the list.
 - `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when the user is already on the matter.
-- `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldNotFoundError` on bad requests.
-- `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldPermissionDeniedError` when the user does not have access.
+- `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldNotFoundOrPermissionDeniedError` when the user does not have
+    access or the ID does not exist.
 - `py42.users.get_by_username()` now raises `Py42UserDoesNotExistError` when the user is not found.
 - `py42.alerts.rules.add_user()` now raises `Py42InvalidRuleTypeError` when trying to add a user to a system rule.
 - `py42.alerts.rules.remove_user()` now raises `Py42InvalidRuleTypeError` when trying to remove user from a system rule.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Added
+
 - Methods for calling the agent-state APIs:
     - `sdk.devices.get_agent_state()`
     - `sdk.devices.get_agent_full_disk_access_state()`
     - `sdk.orgs.get_agent_state()`
     - `sdk.orgs.get_agent_full_disk_access_states()`
-
-### Added
 
 - Exception classes (`py42.execeptions`
     - `Py42ResponseError`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.orgs.get_agent_state()`
     - `sdk.orgs.get_agent_full_disk_access_states()`
 
-- Exception classes (`py42.execeptions`
+- Exception classes (`py42.execeptions`)
     - `Py42ResponseError`
     - `Py42UserAlreadyAddedError`
     - `Py42UserNotInLegalHoldError`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `Py42UserNotInLegalHoldError`
     - `Py42LegalHoldNotFoundOrPermissionDeniedError`
     - `Py42UserDoesNotExistError`
+    - `Py42InvalidRuleTypeError`
 
 - Methods for getting individual response pages:
     - `sdk.detectionlists.departing_employee.get_page()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when user already present on list.
 - `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when user already present on list.
-- `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when use already on matter.
+- `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when user already on matter.
 - `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldNotFoundOrPermissionDeniedError` when matter unavailable.
 - `py42.users.get_by_username()` now raises `Py42UserDoesNotExistError`
 - `py42.alerts.rules.add_user()` now raises `Py42InvalidRuleTypeError` when trying to add a user to a system rule.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,13 +60,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when user already present on list.
-- `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when user already present on list.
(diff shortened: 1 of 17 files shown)


src/py42/exceptions.py
```diff
@@ -63,6 +63,44 @@ def __init__(self, error_message):
...
+        )
+
+
+class Py42InvalidRuleTypeError(Py42Error):
+    """An exception raised when trying to add or remove users to a system rule."""
+
+    def __init__(self, rule_id, source):
+        msg = u"Only alert rules with a source of 'Alerting' can be targeted by this command. "
+        msg += u"Rule {0} has a source of '{1}'."
+        super(Py42InvalidRuleTypeError, self).__init__(msg.format(rule_id, source))
```
timabrmsn, line 101: Can we have these be subclasses of Py42HTTPError and keep the original exception's response data on them?

src/py42/exceptions.py
```diff
@@ -63,6 +63,44 @@ def __init__(self, error_message):
...
+
+
```
timabrmsn, line 103: Can we have these be subclasses of Py42HTTPError and keep the original exception's response data on them? 

Edit: changed the selection range, not sure why it picked that part initially.

unparalleled-js: Couldn't just about every exception be a Py42HTTPError, even ones prior to this PR such as `Py42ChecksumNotFoundError` or `Py42SessionInitializationError` ?

unparalleled-js: Also, `Py42UserDoesNotExist` does not actually arise from an exception. It happens in `get_by_username()` which returns an empty list when the user is not found...

I guess that is a change I should probably document.

timabrmsn: Can we wrap the empty list response in one? I think anytime we're raising an exception as a result of making an http request, we should have some way of inspecting the request/response object itself.

timabrmsn: Sorry, somehow missed your first comment. 

Yeah, from a debugability perspective, I would prefer any error that results from an http request to contain the request data with it and not eat it. Because if, for some reason, there's a bug and we're raising the exception when we shouldn't, then we would need to modify py42 first (or use other tools to make the same request and inspect it) to be able to figure out why.

unparalleled-js: Makes sense to me... I got it working currently but am not too satisfied with how the exceptions module behaves, so will look into refactoring

src/py42/exceptions.py
```diff
@@ -5,36 +5,85 @@ class Py42Error(Exception):
     """A generic, Py42 custom base exception."""
 
 
-class Py42ArchiveFileNotFoundError(Py42Error):
+class Py42ResponseError(Py42Error):
```
unparalleled-js, line 8: @timabrmsn I made some adjustments based on what you said -- had to make another base-y exception type here. This scenario is good for exceptions that are raised from responses that did not result in HTTPError, such as `UserDoesNotExistError` where the response is an empty list and we raise an error. Now, we can return the response for debugging purposes.

We still need `Py42Error` because there are some errors which there is not 1 single response responsible for causing the error, such the errors that get raised during `__setitem__` in the `Py42Response` or the error that gets raised when not finding a file (after paging over many responses) in the securitydata module.

unparalleled-js:
@timabrmsn I made quite a bit of changes since your approval

CHANGELOG.md
```diff
@@ -10,13 +10,21 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+### Added
+
 - Methods for calling the agent-state APIs:
     - `sdk.devices.get_agent_state()`
     - `sdk.devices.get_agent_full_disk_access_state()`
     - `sdk.orgs.get_agent_state()`
     - `sdk.orgs.get_agent_full_disk_access_states()`
 
-### Added
+- Exception classes (`py42.execeptions`
```
kiran-chaudhary, line 21: Is this opening brace intentional?

CHANGELOG.md
```diff
@@ -57,6 +65,21 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
 
+### Changed
+
+- `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when the user is already on the list.
+- `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when the user already on the list.
+- `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when the user is already on the matter.
+- `py42.legalhold.get_matter_by_uid()` now raises `Py42LegalHoldNotFoundOrPermissionDeniedError` when the matter is unavailable.
+- `py42.users.get_by_username()` now raises `Py42UserDoesNotExistError` when the user is not found.
+- `py42.alerts.rules.add_user()` now raises `Py42InvalidRuleTypeError` when trying to add a user to a system rule.
+- `py42.alerts.rules.remove_user()` now raises `Py42InvalidRuleTypeError` when trying to a remove user from a system rule.
```
kiran-chaudhary, line 76: ...trying to remove a user...

src/py42/clients/legalhold.py
```diff
@@ -104,7 +108,10 @@ def get_matter_by_uid(self, legal_hold_uid):
             :class:`py42.response.Py42Response`: A response containing the Matter.
         """
         uri = u"/api/LegalHold/{}".format(legal_hold_uid)
-        return self._session.get(uri)
+        try:
+            return self._session.get(uri)
+        except (Py42BadRequestError, Py42ForbiddenError) as err:
+            raise Py42LegalHoldNotFoundOrPermissionDeniedError(err, legal_hold_uid)
```
kiran-chaudhary, line 114: Wouldn't it be nice if we separate this into two exceptions. Something like:
```
except Py42BadRequestError ....:
   raise Py42LegalHoldNofFound ...
except Py42ForbiddenError ...:
  raise Py42PermissionDeniedError ...
```

 having a `Py42PermissionDeniedError` error might be useful in future. Since our sdk starts with authentication this case doesn't apply much but if there is some API having a role based feature, having "permission denied error" can be useful. 
Quoting "Zen of Python": Explicit is better than implicit.

src/py42/modules/alertrules.py
```diff
@@ -139,3 +151,11 @@ def get_by_observer_id(self, observer_id):
         """
         alerts_client = self.microservice_client_factory.get_alerts_client()
         return alerts_client.get_rule_by_observer_id(observer_id)
+
+
+def _check_if_system_rule(base_err, rules):
+    """You cannot add or remove users from system rules this way."""
```
kiran-chaudhary, line 157: this way!?

src/py42/clients/legalhold.py
unparalleled-js: I think `Py42ForbiddenError` is `Py42PermissionDeniedError`, but I agree these should be split. I assume this s a common use case for users not to have access to legal hold, and therefore we made this exception to increase usability there. @timabrmsn Am I miss anything there? Do you foresee any issue? In the CLI, we will just catch both legal hold-specific exceptions.  Is this 1-1 meaning `Py42BadRequest` actually maps to `Py42LegalholdNotFound`?

unparalleled-js:
@kiran-chaudhary  I addressed all your feedback

src/py42/exceptions.py
timabrmsn: nice, I like it.

src/py42/clients/legalhold.py
timabrmsn: I did a bit of testing on this, and it looks like the Py42BadRequest exception is raised when the hold_uid is either not an integer, or else too large (look like 20 digits long is the cutoff). Otherwise it raises the Py42ForbiddenError if the ID doesn't exist or we don't have permission to view it.

unparalleled-js: Ok... Well, that explains why it was the way it was. I will put it back - but I will add a comment explaining this so that if we cross this bridge again, we will remember these findings. thanks

unparalleled-js:
> Can `Py42InvalidRuleTypeError` be renamed to `Py42InvalidOperation` or `Py42InvalidAction` or something on those lines, as that would indicate to user more clearly why it failed over saying its an Invalid Rule.

I see what you are saying I think - that the rule itself is not invalid but the operation of adding a user to the rule is invalid

src/py42/clients/legalhold.py
unparalleled-js: Or would it be better to leave bad request as it is and this catch `Py42PermissionDeniedError` and have a similar message to be, like "The ID may not exist or you may not have permission to view the matter" ?

unparalleled-js: https://github.com/code42/py42/pull/159/commits/5ad1f3f74300709ab912645f6d944a5f836c1295

kiran-chaudhary: Given that we have documented as explained by @timabrmsn above, as long as the user can look into the exception and figure out  which error occurred its fine if we keep it as is.

unparalleled-js: I made it such that only Forbidden raises the error

unparalleled-js:
> Just shared how I see [this](https://github.com/code42/py42/pull/159#discussion_r466212006) case, otherwise LGTM.

I think it is a little better now - @timabrmsn said that only permission denied was ambiguous, so I removed bad request from that `except` block.  If you get permission denied, it can be either because the ID does not exist or because you don't have access. If you get Bad Request, the ID is not a valid ID, like it is not a number... Does that make sense?

kiran-chaudhary:
> > Just shared how I see [this](https://github.com/code42/py42/pull/159#discussion_r466212006) case, otherwise LGTM.
> 
> I think it is a little better now - @timabrmsn said that only permission denied was ambiguous, so I removed bad request from that `except` block. If you get permission denied, it can be either because the ID does not exist or because you don't have access. If you get Bad Request, the ID is not a valid ID, like it is not a number... Does that make sense?

Yes.

PR #160, opened 2020-08-03
Feature/show-more-directory-support

unparalleled-js:
### Description of Change ###

Kind of a prerequisite for enabling restoring multiple files or directories. INTEG-206.
Wanted to document a little more on restoring directories and also use the `zip_result` param because it does not hurt and looks more intentional. I also cleaned up some testing around this, using kwargs because it is more clear to test parameters I think this way.

### Issues Resolved ###

Does not resolve any issue but is part of INTEG-206 work

### Testing Procedure ###

Make sure you can still restore files via `archive.stream_from_backup()`.
Run in debug mode to set `zip_result` set to `True` when restoring a directory.

### PR Checklist ###

### Additional Notes

I don't think a CHANGELOG entry is needed here, but correct me if I'm wrong. A request parameter is changing under the hood.

We could also make `zip_result` a parameter if you request

### DIFF
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -43,16 +43,16 @@ class ArchiveAccessorManager(object):
     def _get_decryption_keys(self, device_guid, private_password, encryption_key):
         decryption_keys = {}
         if encryption_key:
-            decryption_keys["encryption_key"] = encryption_key
+            decryption_keys[u"encryption_key"] = encryption_key
         else:
             data_key_token = (
                 self._get_data_key_token(device_guid) if not encryption_key else None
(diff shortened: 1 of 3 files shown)


src/py42/_internal/archive_access.py
```diff
@@ -160,13 +160,17 @@ def is_job_complete(self, job_id):
         return self._get_completion_status(response)
 
     def _start_restore(self, file_selection):
+        zip_result = (
```
unparalleled-js, line 163: I just want to my example in the docs to always be True, that is why I thought it might safer to just to explicitly request a zipped result when restoring directories.

PR #162, opened 2020-08-03
add get_agent_state methods

### LINKED ISSUE #161: [Enhancement] Add support for querying agent state by org and device
## Summary
Code42 has APIs that allow the querying of agent state by guid and by org ID. Py42 should allow these to be called and returned in the way we do for other APIs.

Swagger documentation can be found here: 

* https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid
* https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByOrganizationId

## Proposed API

```
sdk.devices.get_agent_state("guid",["example key 1", "example key 2"])
sdk.orgs.get_agent_state("orgId",["example key 1", "example key 2"])
```

## Intended Use Case
Administrators can quickly and easily determine the health or other parameters of agents in their environments.

ceciliastevens:
### Description of Change ###

Adds four methods:

```
sdk.devices.get_agent_state()
sdk.orgs.get_agent_state()
sdk.devices.get_agent_full_disk_access_state()
sdk.orgs.get_agent_full_disk_access_states()
```

### Issues Resolved ###
fixes #161 

### Testing Procedure ###
1. Add a Mac OS endpoint on a recent client (8.2 or later) to a Code42 server
2. Call the new endpoints for its guid and orgId
3. Verify that the status is returned as expected

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- Methods for calling the agent-state APIs"
+    - `sdk.devices.get_agent_state()`
+    - `sdk.devices.get_agent_full_disk_access_state()`
+    - `sdk.orgs.get_agent_state()`
+    - `sdk.orgs.get_agent_full_disk_access_states()`
+
 ### Added
 
 - Methods for getting individual response pages:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
-- Methods for calling the agent-state APIs"
+- Methods for calling the agent-state APIs:
     - `sdk.devices.get_agent_state()`
     - `sdk.devices.get_agent_full_disk_access_state()`
     - `sdk.orgs.get_agent_state()`
--- a/src/py42/clients/devices.py
+++ b/src/py42/clients/devices.py
@@ -245,7 +245,7 @@ class DeviceClient(BaseClient):
 
             Args:
                 guid (str): The globally unique identifier of the device.
-                propertyName (str): The name of the property to retrieve (e.g. `fullDiskAccess`).
+                property_name (str): The name of the property to retrieve (e.g. `fullDiskAccess`).
 
             Returns:
                 :class:`py42.response.Py42Response`: A response containing settings information.
--- a/src/py42/clients/devices.py
+++ b/src/py42/clients/devices.py
@@ -239,19 +239,19 @@ class DeviceClient(BaseClient):
         params = {u"guid": guid, u"keys": keys}
         return self._session.get(uri, params=params)
 
-    def get_agent_state(self, guid, propertyName):
+    def get_agent_state(self, guid, property_name):
         """Gets the agent state of the device.
             `REST Documentation <https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid>`__
 
             Args:
                 guid (str): The globally unique identifier of the device.
-                propertyName (str): The name of the property to retrieve.
+                propertyName (str): The name of the property to retrieve (e.g. `fullDiskAccess`).
 
             Returns:
                 :class:`py42.response.Py42Response`: A response containing settings information.
             """
         uri = u"/api/v14/agent-state/view-by-device-guid"
-        params = {u"deviceGuid": guid, u"propertyName": propertyName}
+        params = {u"deviceGuid": guid, u"propertyName": property_name}
         return self._session.get(uri, params=params)
 
     def get_agent_full_disk_access_state(self, guid):
@@ -264,5 +264,4 @@ class DeviceClient(BaseClient):
             Returns:
                 :class:`py42.response.Py42Response`: A response containing settings information.
             """
-        FULLDISKACCESS = "fullDiskAccess"
-        return self.get_agent_state(guid, FULLDISKACCESS)
+        return self.get_agent_state(guid, "fullDiskAccess")
--- a/src/py42/clients/devices.py
+++ b/src/py42/clients/devices.py
@@ -238,3 +238,31 @@ class DeviceClient(BaseClient):
         uri = u"/api/v4/device-setting/view"
         params = {u"guid": guid, u"keys": keys}
         return self._session.get(uri, params=params)
+
+    def get_agent_state(self, guid, propertyName):
+        """Gets the agent state of the device.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- Methods for calling the agent-state APIs"
```
unparalleled-js, line 13: random quote at the end of this line

src/py42/clients/devices.py
```diff
@@ -238,3 +238,31 @@ def get_settings(self, guid, keys=None):
         uri = u"/api/v4/device-setting/view"
         params = {u"guid": guid, u"keys": keys}
         return self._session.get(uri, params=params)
+
+    def get_agent_state(self, guid, propertyName):
```
unparalleled-js, line 242: We use Hungarian case, `property_name`

src/py42/clients/devices.py
```diff
@@ -238,3 +238,31 @@ def get_settings(self, guid, keys=None):
...
+        """Gets the agent state of the device.
+            `REST Documentation <https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid>`__
+
+            Args:
+                guid (str): The globally unique identifier of the device.
+                propertyName (str): The name of the property to retrieve.
```
unparalleled-js, line 248: Is there any more info for the `property_name` parameter? Like some example values would be nice

src/py42/clients/devices.py
```diff
@@ -238,3 +238,31 @@ def get_settings(self, guid, keys=None):
...
+        """Gets the full disk access status of a device.
+            `REST Documentation <https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid>`__
+
+            Args:
+                guid (str): The globally unique identifier of the device.
+
+            Returns:
+                :class:`py42.response.Py42Response`: A response containing settings information.
+            """
+        FULLDISKACCESS = "fullDiskAccess"
```
unparalleled-js, line 267: Missing `u` prefix on this string, could also just do away with the const variable and pass the string value directly

src/py42/clients/devices.py
```diff
@@ -238,3 +238,31 @@ def get_settings(self, guid, keys=None):
...
+                propertyName (str): The name of the property to retrieve.
+
+            Returns:
+                :class:`py42.response.Py42Response`: A response containing settings information.
+            """
+        uri = u"/api/v14/agent-state/view-by-device-guid"
+        params = {u"deviceGuid": guid, u"propertyName": propertyName}
+        return self._session.get(uri, params=params)
+
+    def get_agent_full_disk_access_state(self, guid):
```
unparalleled-js, line 257: Instead of this being a separate method, should we just make the constants available at the modular level and then have the users use which constant they want plus the `get_agent_state()` method?

unparalleled-js:
Thank you @ceciliastevens ! Most of my thoughts on devices apply to orgs too.

src/py42/clients/devices.py
unparalleled-js: I guess I'd like to know the other types of agent state to request... Is this an ever-growing list? Is there going to be a new method for each one of them? That is why I thought the importable constants might be better. I don't have good domain knowledge on these APIs though.

ceciliastevens: This was discussed in https://github.com/code42/py42/issues/161, I'm following @alanag13's recommendation from there. 

At this time the only property name that seems to be implemented is the fullDiskAccess, although there may be more in the future.

CHANGELOG.md
ceciliastevens: I've updated this in ebd235272b9b85156655f837475a2ec872406253

src/py42/clients/devices.py
ceciliastevens: I've updated this in ebd235272b9b85156655f837475a2ec872406253

src/py42/clients/devices.py
ceciliastevens: I've updated this in ebd235272b9b85156655f837475a2ec872406253

src/py42/clients/devices.py
ceciliastevens: I've updated this in ebd235272b9b85156655f837475a2ec872406253

src/py42/clients/devices.py
unparalleled-js: Oh okay, I apologize, I didn't see that.

After seeing that, perhaps `FULL_DISK_ACCESS` should be a constant at the modular _and_ get used in this function like @alanag13 suggested.

Not a biggie right now though since there is only one.

src/py42/clients/devices.py
```diff
@@ -238,3 +238,30 @@ def get_settings(self, guid, keys=None):
...
         params = {u"guid": guid, u"keys": keys}
         return self._session.get(uri, params=params)
+
+    def get_agent_state(self, guid, property_name):
+        """Gets the agent state of the device.
+            `REST Documentation <https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid>`__
+
+            Args:
+                guid (str): The globally unique identifier of the device.
+                propertyName (str): The name of the property to retrieve (e.g. `fullDiskAccess`).
```
unparalleled-js, line 248: Casing in doc str for this param still camel

unparalleled-js: I can approve after this!

src/py42/clients/orgs.py
```diff
@@ -160,3 +160,30 @@ def get_current(self, **kwargs):
...
         uri = u"/api/Org/my"
         return self._session.get(uri, params=kwargs)
+
+    def get_agent_state(self, orgId, property_name):
+        """Gets the agent state of the devices in the org.
+            `REST Documentation <https://console.us.code42.com/swagger/index.html?urls.primaryName=v14#/agent-state/AgentState_ViewByDeviceGuid>`__
+
+            Args:
+                orgId (str): The org's identifier.
+                propertyName (str): The name of the property to retrieve (e.g. `fullDiskAccess`).
```
unparalleled-js, line 170: Here as well

src/py42/clients/devices.py
ceciliastevens: Should now be updated!

src/py42/clients/orgs.py
ceciliastevens: Updated here as well.

PR #164, opened 2020-08-05
Feature/stream selection lists

unparalleled-js:
### Description of Change ###

Permits downloading from multiple paths using `archive.stream_from_backup()`.

### Issues Resolved ###

- closes # INTEG-206

### Testing Procedure ###

Here are some example method calls:

```
sdk.archive.stream_from_backup([DIR_TO_RESTORE, DIR_TO_RESTORE_2], MY_WINDOWS)
sdk.archive.stream_from_backup(DIR_TO_RESTORE, MY_WINDOWS)
sdk.archive.stream_from_backup([DIR_TO_RESTORE_2], MY_WINDOWS)
sdk.archive.stream_from_backup(FILE_TO_RESTORE, MY_WINDOWS)
sdk.archive.stream_from_backup([FILE_TO_RESTORE], MY_WINDOWS)
```

### PR Checklist ###

### DIFF
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -282,6 +282,7 @@ class RestoreJobManager(_RestorePoller):
         is_done = response[u"done"]
         percentage_dict = {
             u"jobId": job_id,
+            u"status": response[u"status"],
             u"percentComplete": response[u"percentComplete"] if not is_done else 100,
         }
         debug.logger.debug(format_dict(percentage_dict))
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -43,16 +43,16 @@ class ArchiveAccessorManager(object):
     def _get_decryption_keys(self, device_guid, private_password, encryption_key):
         decryption_keys = {}
         if encryption_key:
-            decryption_keys["encryption_key"] = encryption_key
+            decryption_keys[u"encryption_key"] = encryption_key
         else:
             data_key_token = (
                 self._get_data_key_token(device_guid) if not encryption_key else None
             )
             if data_key_token:
-                decryption_keys["data_key_token"] = data_key_token
+                decryption_keys[u"data_key_token"] = data_key_token
 
             if private_password:
-                decryption_keys["private_password"] = private_password
+                decryption_keys[u"private_password"] = private_password
         return decryption_keys
 
     def _get_data_key_token(self, device_guid):
@@ -166,18 +166,19 @@ class RestoreJobManager(object):
         response = self._storage_archive_client.get_restore_status(job_id)
         return self._get_completion_status(response)
 
-    def _start_restore(self, file_selections, exceptions):
-        num_files = sum([fs.num_files for fs in file_selections])
-        num_dirs = sum([fs.num_dirs for fs in file_selections])
-        size = sum([fs.size for fs in file_selections])
+    def _start_restore(self, file_selection, exceptions):
+        num_files = sum([fs.num_files for fs in file_selection])
+        num_dirs = sum([fs.num_dirs for fs in file_selection])
+        size = sum([fs.size for fs in file_selection])
+        zip_result = _check_for_multiple_files(file_selection) or None
         return self._storage_archive_client.start_restore(
-            self._device_guid,
-            self._archive_session_id,
-            [fs.path_set for fs in file_selections],
-            num_files,
-            num_dirs,
-            size,
-            zip_result=len(file_selections) > 1 or None,
+            guid=self._device_guid,
+            web_restore_session_id=self._archive_session_id,
+            path_set=[fs.path_set for fs in file_selection],
+            num_files=num_files,
+            num_dirs=num_dirs,
+            size=size,
+            zip_result=zip_result,
             show_deleted=True,
             exceptions=exceptions,
         )
@@ -193,3 +194,14 @@ class RestoreJobManager(object):
 
 def create_restore_job_manager(storage_archive_client, device_guid, archive_session_id):
     return RestoreJobManager(storage_archive_client, device_guid, archive_session_id)
+
+
+def _check_for_multiple_files(file_selection):
+    if not file_selection:
+        return False
+    if len(file_selection) > 1:
+        return True
+
+    # Only one file selected
+    selection = file_selection[0]
+    return selection.path_set[u"type"].lower() == u"directory"
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -1,8 +1,11 @@
 import posixpath
 import time
+import json
 from collections import namedtuple
 
 from py42.exceptions import Py42ArchiveFileNotFoundError
+from py42.settings import debug
+from py42.util import format_dict
 
 
 FileSelection = namedtuple(u"FileSelection", u"path_set, num_files, num_dirs, size")
@@ -184,12 +187,14 @@ class _RestorePoller(object):
         )
 
 
-def _create_size_dict(size_response):
-    return {
-        u"numFiles": size_response[u"numFiles"],
-        u"numDirs": size_response[u"numDirs"],
-        u"size": size_response[u"size"],
-    }
+def _create_size_dict(job_id, size_response):
+    size_dict = json.loads(size_response.text)
+    size_dict[u"jobId"] = job_id
+    return size_dict
+
+
+def _print_file_size(size_dict):
+    debug.logger.debug(format_dict(size_dict))
 
 
 class FileSizePoller(_RestorePoller):
@@ -226,9 +231,11 @@ class FileSizePoller(_RestorePoller):
         while job_ids:
             for job_id in job_ids:
                 response = self._get_job_status(job_id)
+                size_dict = _create_size_dict(job_id, response)
+                _print_file_size(size_dict)
                 if response[u"status"] == u"DONE":
                     job_ids.remove(job_id)
-                    sizes.append(_create_size_dict(response))
+                    sizes.append(size_dict)
 
             # File size calculation is taking too long.
             if time.time() - t0 > timeout:
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -107,7 +107,7 @@ class ArchiveAccessor(object):
     def stream_from_backup(self, file_paths, ignore_size_calc=False):
         if not isinstance(file_paths, (list, tuple)):
             file_paths = [file_paths]
-        file_paths = [fp.replace("\\", "/") for fp in file_paths]
+        file_paths = [fp.replace(u"\\", u"/") for fp in file_paths]
         metadata_list = self._get_restore_metadata(file_paths)
         file_ids = [md[u"id"] for md in metadata_list]
         file_sizes = (
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -4,6 +4,7 @@ from collections import namedtuple
 
 from py42.exceptions import Py42ArchiveFileNotFoundError
 
+
 FileSelection = namedtuple(u"FileSelection", u"path_set, num_files, num_dirs, size")
 
 
@@ -82,6 +83,10 @@ def _create_file_selections(file_paths, metadata_list, file_sizes):
     return file_selections
 
 
+def _get_default_file_size_info(file_ids):
+    return [{u"numFiles": 1, u"numDirs": 1, u"size": 1} for _ in file_ids]
+
+
 class ArchiveAccessor(object):
 
     DEFAULT_DIRECTORY_DOWNLOAD_NAME = u"download"
@@ -99,13 +104,17 @@ class ArchiveAccessor(object):
         self._storage_archive_client = storage_archive_client
         self._restore_job_manager = restore_job_manager
 
-    def stream_from_backup(self, file_paths):
+    def stream_from_backup(self, file_paths, ignore_size_calc=False):
         if not isinstance(file_paths, (list, tuple)):
             file_paths = [file_paths]
 
         metadata_list = self._get_restore_metadata(file_paths)
         file_ids = [md[u"id"] for md in metadata_list]
-        file_sizes = self._get_file_size_info(file_ids)
+        file_sizes = (
+            self._get_file_size_info(file_ids)
+            if not ignore_size_calc
+            else _get_default_file_size_info(file_ids)
+        )
         file_selections = _create_file_selections(file_paths, metadata_list, file_sizes)
         return self._restore_job_manager.get_stream(file_selections)
 
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -274,6 +274,11 @@ class RestoreJobManager(_RestorePoller):
 
     def _is_job_complete(self, job_id):
         response = self._storage_archive_client.get_restore_status(job_id)
+        percentage_dict = {
+            u"jobId": job_id,
+            u"percentComplete": response[u"percentComplete"],
+        }
+        debug.logger.debug(format_dict(percentage_dict))
         return response[u"done"]
 
     def _start_restore(self, file_selection):
--- a/src/py42/_internal/archive_access.py
+++ b/src/py42/_internal/archive_access.py
@@ -82,11 +82,18 @@ class ArchiveAccessor(object):
         self._restore_job_manager = restore_job_manager
 
     def stream_from_backup(self, file_path):
-        metadata = self._get_file_via_walking_tree(file_path)
-        file_selection = self._build_file_selection(
-            metadata[u"path"], metadata[u"type"]
-        )
-        return self._restore_job_manager.get_stream(file_selection)
+        if not isinstance(file_path, (list, tuple)):
+            file_path = [file_path]
+
+        selections = []
+        for path in file_path:
+            metadata = self._get_file_via_walking_tree(path)
+            file_selection = self._build_file_selection(
+                metadata[u"path"], metadata[u"type"]
+            )
+            selections.append(file_selection)
+
+        return self._restore_job_manager.get_stream(selections)
 
     def _get_file_via_walking_tree(self, file_path):
         path_parts = file_path.split(u"/")
@@ -125,7 +132,7 @@ class ArchiveAccessor(object):
 
     @staticmethod
     def _build_file_selection(file_path, file_type):
-        path_set = [{u"type": file_type, u"path": file_path, u"selected": True}]
+        path_set = {u"type": file_type, u"path": file_path, u"selected": True}
         # pass in dummy values; only used with progress indication, which we don't currently use
         num_files = num_dirs = size = 1
         return FileSelection(path_set, num_files, num_dirs, size)
@@ -146,8 +153,8 @@ class RestoreJobManager(object):
         self._archive_session_id = archive_session_id
         self._job_polling_interval = job_polling_interval
 
-    def get_stream(self, file_selection):
-        response = self._start_restore(file_selection)
+    def get_stream(self, file_selections):
+        response = self._start_restore(file_selections)
         job_id = response["jobId"]
 
         while not self.is_job_complete(job_id):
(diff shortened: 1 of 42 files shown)


unparalleled-js:
> Referred [endpoint doc], are we excluding the timestamp parameter intentionally.(https://console.us.code42.com/apidocviewer/#WebRestoreJob-post) .
> 
> Also as per above doc the request format is list of dictionary with key "path"
> 
> ```
> [{
> 	"path":"/home/user/file1.txt", "timestamp":1392130247
> }] 
> ```
> 
> However if it works then do ignore.

I think there is a separate story for that. The timestamp is used for getting file versions.

src/py42/_internal/archive_access.py
```diff
@@ -81,12 +81,17 @@ def __init__(
...
-        file_selection = self._build_file_selection(
-            metadata[u"path"], metadata[u"type"]
-        )
-        return self._restore_job_manager.get_stream(file_selection)
+    def stream_from_backup(self, file_path, exceptions):
+        if not isinstance(file_path, (list, tuple)):
+            file_path = [file_path]
+
+        file_selections = []
+        for path in file_path:
```
alanag13, line 89: can this be `file_paths`? The name `file_path` sounds like a single string.

src/py42/_internal/archive_access.py
```diff
@@ -159,15 +164,21 @@ def is_job_complete(self, job_id):
...
-            file_selection.size,
+            guid=self._device_guid,
+            web_restore_session_id=self._archive_session_id,
+            path_set=[fs.path_set for fs in file_selection],
+            num_files=num_files,
+            num_dirs=num_dirs,
+            size=size,
+            zip_result=zip_result,
             show_deleted=True,
+            exceptions=exceptions,
```
alanag13, line 181: I would call this something like `exclusions` to avoid it being associated with the programming concept of an exception (error).

alanag13, line 181: will this exclude a directory if one is passed here, or do you need to list out every individual file?

src/py42/_internal/archive_access.py
```diff
@@ -159,15 +164,21 @@ def is_job_complete(self, job_id):
...
-            self._device_guid,
-            self._archive_session_id,
-            file_selection.path_set,
-            file_selection.num_files,
-            file_selection.num_dirs,
-            file_selection.size,
+            guid=self._device_guid,
+            web_restore_session_id=self._archive_session_id,
+            path_set=[fs.path_set for fs in file_selection],
+            num_files=num_files,
```
alanag13, line 176: is `num_files` supposed to include the files that are in the selected directories?

src/py42/modules/archive.py
```diff
@@ -14,41 +14,51 @@ def stream_from_backup(
...
+        exceptions=None,
     ):
-        """Streams a file from a backup archive to memory.
+        """Streams a file from a backup archive to memory. If streaming a directory, the
+        results will be zipped.
         `REST Documentation <https://console.us.code42.com/apidocviewer/#WebRestoreJobResult-get>`__
 
         Args:
-            file_path (str): The path to the file in your archive.
+            file_path (str or list): The path or list of paths to the files or directories in your archive.
```
alanag13, line 24: `list` should include the type of object in the list

src/py42/modules/archive.py
```diff
@@ -14,41 +14,51 @@ def stream_from_backup(
...
-                security. Defaults to None.
-            encryption_key (str, None): A custom encryption key for decryption an archive's file
-                contents, necessary for restoring files. This is only relevant to users with custom
+            archive_password (str or None, optional): The password for the archive, if password-
+                protected. This is only relevant to users with archive key password security. Defaults
+                to None.
+            encryption_key (str or None, optional): A custom encryption key for decrypting an archive's
+                file contents, necessary for restoring files. This is only relevant to users with custom
                 key archive security. Defaults to None.
+            exceptions (list, optional): A list of file paths to exclude from the restore. Only
```
alanag13, line 36: `list` should include the type of object in the list

src/py42/_internal/archive_access.py
unparalleled-js: Yes, I can do that. This would be a little breaking change, for key-value parameters.

src/py42/_internal/archive_access.py
unparalleled-js: `supposed to` ? No, I don't think it really matters.

But does it in the UI? Yes.

In any case, we are setting them to `1` always when crafting `FileSelection` objects. This is because these params don't really matter.  However, when you get to _this_ point in a file restore execution, I'd expect this is where you would amass them the counts for the request.  During the creaetion of `FileSelection` objects is where you would figure out all the counts for `num_files`, `num_dirs`, and `size`.... not here.

src/py42/_internal/archive_access.py
unparalleled-js: Having issues here with exclusions... will take a minute to figure it out; I will let you know. I think I addressed everything else though

src/py42/_internal/archive_access.py
alanag13: if they don't matter, why do anything different with them than what we were doing previously?

unparalleled-js: Previously there was only one value each.. Now, there is a list. Do I just use the values of the first element of the list? That seems odd... So I amassed them.

src/py42/_internal/clients/storage/archive.py
```diff
@@ -100,7 +100,7 @@ def start_restore(
         show_deleted=None,
         restore_full_path=None,
         timestamp=None,
-        exceptions=None,
+        exclusions=None,
```
alanag13, line 103: are these parameters still needed? If they aren't going to be used anymore, i'd rather just remove them from here.

unparalleled-js: All of the unused ones? I removed `exclusions` from the higher level calls. There are many examples in py42 where modules don't use all the request params of client methods.

I will remove `exclusions` here since it is faulty, but it still is technically a request parameter.

If there are more parameters you are referring to (on any client), let me know and I can remove those too.

src/py42/_internal/archive_access.py
unparalleled-js: I am considering that it might be best to just implement the logic for figuring out the dir and file counts (during the part where we create `FileSelection` tuples). Reasons:

1.  They are required parameters even though they shouldn't be, but why not lie.
2. It could be helpful to know these counts during debugging and looking at sent request params.  (maybe?)
3. It could afford progress bars in people's scripts.

timabrmsn: It would be nice to have a method available to generate the file count, especially for when we add push restore functionality, as that's where the real benefit is. The desktop client then can show an accurate progress bar and estimated time remaining, etc. 

However, knowing that in some extremely large-scale archives, calculating file counts can take an extremely long time (if a folder has a million files in it, for example). I'd suggest we add a `calculation_timeout` param that would skip waiting for results past a certain time and just start the restore job.

src/py42/modules/archive.py
```diff
@@ -9,46 +9,54 @@ def __init__(self, archive_accessor_manager, archive_client):
...
         self,
-        file_path,
+        file_paths,
         device_guid,
         destination_guid=None,
         archive_password=None,
         encryption_key=None,
     ):
-        """Streams a file from a backup archive to memory.
+        """Streams a file from a backup archive to memory. If streaming a multiple files, the
```
timabrmsn, line 18: "a multiple files"

src/py42/_internal/archive_access.py
unparalleled-js: I had issues trying to implement a timeout decorator... I could get it to terminate the whole process but couldn't get it to return defaults values without terminating the whole process.

Would it be easier to just have a parameter for ignoring calculating file sizes and then to recommend large restores to use this parameter?

^ Did this on the last commit

timabrmsn: Wouldn't you use the https://www.crashplan.com/apidocviewer/#WebRestoreFileSizePolling endpoint to poll for file counts/size (like console UI does) and you could just end the polling loop if it goes past the timeout window?

src/py42/_internal/archive_access.py
```diff
@@ -81,12 +104,31 @@ def __init__(
         self._storage_archive_client = storage_archive_client
         self._restore_job_manager = restore_job_manager
 
-    def stream_from_backup(self, file_path):
-        metadata = self._get_file_via_walking_tree(file_path)
-        file_selection = self._build_file_selection(
-            metadata[u"path"], metadata[u"type"]
+    def stream_from_backup(self, file_paths, ignore_size_calc=False):
```
timabrmsn, line 107: I know this isn't new to this PR, but I encountered it while testing just now. The api expects all path separators (even for Windows paths) to be `/` so if a user enters a Windows path as they would on Windows (`C:\Users\admin\Desktop`), they'll get a File Not Found error. We should do a `.replace("\\", "/")` on the input strings.

unparalleled-js: Good find!

src/py42/_internal/archive_access.py
unparalleled-js: Oh and just check the time in the loop? Ya I guess so. I was trying to using signals and what-not to interrupt the thread after a timeout

unparalleled-js: I was trying to just use "/api/WebRestoreFileSize".

unparalleled-js: It is going to take a while to get this in, but I am working on it

unparalleled-js:
> # [Codecov](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=h1) Report
> > Merging [#164](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=desc) into [master](https://codecov.io/gh/code42/py42/commit/6904abab4244ec7269679aa5e645834fdc733aeb&el=desc) will **decrease** coverage by `0.81%`.
> > The diff coverage is `73.19%`.
> 
> [Impacted file tree graph](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=tree)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           master     #164      +/-   ##
> ==========================================
> - Coverage   92.50%   91.68%   -0.82%     
> ==========================================
>   Files          64       64              
>   Lines        2387     2465      +78     
> ==========================================
> + Hits         2208     2260      +52     
> - Misses        179      205      +26     
> ```
> 
> [Impacted Files](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=tree)	Coverage Δ	
> [src/py42/_internal/compat.py](https://codecov.io/gh/code42/py42/pull/164/diff?src=pr&el=tree#diff-c3JjL3B5NDIvX2ludGVybmFsL2NvbXBhdC5weQ==)	`100.00% <ø> (ø)`	
> [src/py42/_internal/archive_access.py](https://codecov.io/gh/code42/py42/pull/164/diff?src=pr&el=tree#diff-c3JjL3B5NDIvX2ludGVybmFsL2FyY2hpdmVfYWNjZXNzLnB5)	`84.43% <70.45%> (-15.57%)`	⬇️
> [src/py42/_internal/clients/storage/archive.py](https://codecov.io/gh/code42/py42/pull/164/diff?src=pr&el=tree#diff-c3JjL3B5NDIvX2ludGVybmFsL2NsaWVudHMvc3RvcmFnZS9hcmNoaXZlLnB5)	`100.00% <100.00%> (ø)`	
> [src/py42/modules/archive.py](https://codecov.io/gh/code42/py42/pull/164/diff?src=pr&el=tree#diff-c3JjL3B5NDIvbW9kdWxlcy9hcmNoaXZlLnB5)	`100.00% <100.00%> (ø)`	
> [Continue to review full report at Codecov](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=continue).
> 
> > **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> > `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> > Powered by [Codecov](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=footer). Last update [6904aba...831c2cd](https://codecov.io/gh/code42/py42/pull/164?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

Does this not update itself?  When I run code coverage locally, I get the 100% on `archive_access` after my recent commits

unparalleled-js:
Does CodeCov not update itself? When I run code coverage locally, I get the 100% on archive_access.py after my recent commits

kiran-chaudhary:
> Does CodeCov not update itself? When I run code coverage locally, I get the 100% on `archive_access.py` after my recent commits

It does, but it doesn't show report on every commit as per my observation.

unparalleled-js:
@kiran-chaudhary @alanag13 @timabrmsn I can't really proceed with push restore until we can agree on some of these changes here. Let me know if this warrants another design discussion; I am happy to adjust...

src/py42/_internal/archive_access.py
```diff
@@ -167,48 +168,107 @@ def _get_children(self, node_id=None):
...
+            return None
+
+        sizes = []
+        t0 = time.time()
         for file_id in file_ids:
-            size_data = self._storage_archive_client.get_file_size(
-                self._device_guid, file_id
+            job_id = self._start_poll(file_id)
+            response = self._wait_for_job(job_id)
+            sizes.append(
```
timabrmsn, line 210: Can we kick off all the jobs at once before waiting for them so they can run in parallel?

src/py42/modules/archive.py
```diff
@@ -34,9 +34,9 @@ def stream_from_backup(
             encryption_key (str or None, optional): A custom encryption key for decrypting an archive's
                 file contents, necessary for restoring files. This is only relevant to users with custom
                 key archive security. Defaults to None.
-            ignore_size_calc (bool, optional): When True, ignores figuring our accurate file counts and
-                sizes. Recommended to set to True for large restores. File count and sizes are used
-                for indicating progress.
+            file_size_calc_timeout (int, optional): Set to limit the amount of seconds spent calculating
+                file sizes when crafting the request. Set to 0 to ignore file sizes altogether. Defaults
+                to None.
```
timabrmsn, line 39: Should clarify here that defaulting to None means it uses a timeout of 10 seconds. Also, would it make sense to have `None` be "no timeout" so it just runs until the job completes, however long that might take?

src/py42/_internal/archive_access.py
```diff
@@ -98,22 +99,22 @@ def __init__(
...
             file_paths = [file_paths]
-
+        file_paths = [fp.replace(u"\\", u"/") for fp in file_paths]
         metadata_list = self._get_restore_metadata(file_paths)
         file_ids = [md[u"id"] for md in metadata_list]
-        file_sizes = (
-            self._get_file_size_info(file_ids)
-            if not ignore_size_calc
-            else _get_default_file_size_info(file_ids)
+        file_sizes = self._file_size_poller.get_file_sizes(
```
timabrmsn, line 116: Are the calculated file sizes accessible by the user at any point in the process of creating/streaming a restore job? I wonder if it would be useful to somehow expose the job size, so if it's very large there will be some expectation that it'll take longer. Maybe at least log the results at debug level? The file size polling could probably also use some progress logging now that I'm thinking of it.

unparalleled-js: Interesting. I think debug works now, but here is an idea for how we can achieve this:

Create a new logger specifically for Restores (and size polling)... since python loggers are singletons, we can allow this logger to be exportable. The end user would then add a handler that sends the logs to wherever, such as updating UIs.

src/py42/modules/archive.py
unparalleled-js: Yes, I agree with both of that and will put that in.

src/py42/_internal/archive_access.py
unparalleled-js: I added the debugging logging and tested it out. It works! I have it like JSON so it looks like the other logging we do. Also, File size calculation does not take nearly as long as the actual restore.

unparalleled-js: Like, the restore is magnitudes longer... so perhaps that is where the real progress bar should happen. I restored my entire C drive (which was still pretty small), and I did the size calculation in "WORKING" state for a very short amount of time.

unparalleled-js: Should we log progress during restore?

timabrmsn: Yeah, most calculations happen pretty quick, but if you've got an archive with a huge number of files, it can spin for quite awhile, so it would be useful in those cases to be able to see that it's doing _something_ and whether it is making progress or not. 

Yeah, we should log anytime there's an operation happening behind the scenes that the user might want to know about.

unparalleled-js: Well, we are logging requests. So, we can tell that it _is_ doing something. This was true for sile polling too. But logging progress is just the next step.

src/py42/_internal/archive_access.py
```diff
@@ -125,61 +171,144 @@ def _get_children(self, node_id=None):
...
+            for job_id in job_ids:
+                response = self._get_job_status(job_id)
+                size_dict = _create_size_dict(job_id, response)
+                _print_file_size(size_dict)
+                if response[u"status"] == u"DONE":
+                    job_ids.remove(job_id)
+                    sizes.append(size_dict)
+
+            # File size calculation is taking too long.
+            if time.time() - t0 > timeout:
```
unparalleled-js, line 241: Note that this is not a True timeout... Trying to use signal to achieve that but would need this to be on a separate process.  Maybe something to consider.  Could also check this in the `for` loop

src/py42/modules/archive.py
```diff
@@ -1,3 +1,6 @@
+_FILE_SIZE_CALC_TIMEOUT = 10
```
unparalleled-js, line 1: `10` enough?

src/py42/_internal/archive_access.py
```diff
@@ -125,61 +171,151 @@ def _get_children(self, node_id=None):
...
-
-    def is_job_complete(self, job_id):
+    def _is_job_complete(self, job_id):
         response = self._storage_archive_client.get_restore_status(job_id)
-        return self._get_completion_status(response)
+        percentage_dict = {
+            u"jobId": job_id,
+            u"percentComplete": response[u"percentComplete"],
+        }
+        debug.logger.debug(format_dict(percentage_dict))
```
unparalleled-js, line 281: logging percent complete now! @timabrmsn

src/py42/_internal/archive_access.py
unparalleled-js: Ok, logging percent complete now! I agree this is much better for the logs

src/py42/modules/archive.py
timabrmsn: Yeah, I think that's fine.

src/py42/_internal/archive_access.py
```diff
@@ -125,61 +176,151 @@ def _get_children(self, node_id=None):
...
 
-        return self._get_stream(job_id)
-
-    def is_job_complete(self, job_id):
+    def _is_job_complete(self, job_id):
         response = self._storage_archive_client.get_restore_status(job_id)
-        return self._get_completion_status(response)
+        percentage_dict = {
+            u"jobId": job_id,
+            u"percentComplete": response[u"percentComplete"],
```
timabrmsn, line 284: I tested this out and it looks like when the job finishes the response doesn't have `percentComplete` so an exception is thrown trying to build the `percentage_dict`. 

```
{'jobId': '967372837549429135', 'sourceId': '947634481628936837', 'name': 'WebRestore_3', 'userId': 13483629, 'done': True, 'canceled': False, 'creationDate': 1597346743610, 'expirationDate': 1597433143610, 'bytesZipped': 668621884, 'zipResult': True}
```

unparalleled-js: Whoops, I never let it finish running long enough to see this. I have been restoring my whole C drive to test this out. Will fix now, thanks!

unparalleled-js:
Try it again! @timabrmsn

src/py42/_internal/archive_access.py
timabrmsn: It looks like there are two parts to the web restore job that have different percentages. The status "preparing" and the status "compressing". We should probably add status to the percentage dict so it doesn't look like the job just starts over at 0% and runs through again.

unparalleled-js:
> Apart from logging the "status" portion of the web restore job percentages, and maybe adding a note in the changelog that progress can be tracked with debug logging, this looks good to me!

Very cool! I did implement those latest suggestions, as they did seem pretty important

PR #165, opened 2020-08-06
Fix tenant ID

unparalleled-js:
### Description of Change ###

Removed tenant ID from `alerts.get_details()`.  This is because it was removed from an internal client and it is not needed.

### Testing Procedure ###
Run this CLI command `code42 alerts search -f JSON -b 2020-05-20` and make sure it does indicate this error at the very top of the output.

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Removed
+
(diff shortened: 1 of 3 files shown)


alanag13:
@unparalleled-js if we're going to remove this here, we might as well remove it from the other methods where it is currently optional, too.

unparalleled-js:
> @unparalleled-js if we're going to remove this here, we might as well remove it from the other methods where it is currently optional, too.

Right, is done now

PR #166, opened 2020-08-07
Implement Org and Device Settings

### LINKED ISSUE #154: [Enhancement] Add support for reading/writing settings APIs
## Summary
Add some helper functions that can handle the different settings APIs (outlined in Jira INTEG-904)

## Intended Use Case
PS would like to start adding methods to py42 which can update specific settings for devices and orgs, and in order to do so in a scalable manner, we should abstract the common actions into some helper functions.

timabrmsn:
### Description of Change ###

Implements a new classes `OrgSettings` and `DeviceSettings` that can be used to manage settings for an Organization or Device. 

The Org class abstracts the management of all settings for an org, regardless of whether the setting uses the `/api/Org` endpoint or `/api/OrgSetting` endpoint. It stores the org's existing settings internally and allows viewing/modifying settings via properties. Once all desired modifications are made, to actually update the org the manager can be passed to `sdk.orgs.update_org_settings()` for posting updates to the server. Since there could be potentially two API calls made to two different endpoints, `update_org_settings()` returns a namedtuple with the result of both calls. If any exceptions were raised in making either call, the result's `.error` attribute will be set to `True` and the exception will be stored on the result's appropriate attribute instead of the response.

Example usage:

```
>>> settings_manager = sdk.orgs.get_settings(<org_id>)
>>> 
>>> settings_manager.org_name
'Endpoint Monitoring Org'
>>> 
>>> settings_manager.endpoint_monitoring_enabled
True
>>> 
>>> settings_manager.org_name = "Endpoint Monitoring Disabled"
>>> settings_manager.endpoint_monitoring_enabled = False
>>> 
>>> settings_manager.changes
{'settings': {'orgName': 'Endpoint Monitoring Disabled'}, 't_settings': [{'key': 'org-securityTools-enable', 'value': 'false', 'locked': False}, {'key': 'device_advancedExfiltrationDetection_enabled', 'value': 'false', 'locked': False
}, {'key': 'org-securityTools-device-detection-enable', 'value': 'false', 'locked': False}, {'key': 'org-securityTools-open-file-detection-enable', 'value': 'false', 'locked': False}, {'key': 'org-securityTools-cloud-detection-enable'
, 'value': 'false', 'locked': False}]}
>>> 
>>> result = sdk.orgs.update_settings(settings_manager)
>>> 
>>> result.error
False
>>> result.settings_response
<Py42Response [status=200, data={'active': True, 'blocked': False, 'classification': 'BASIC', 'configInheritanceCounts': {}, ...}]>
>>> result.org_settings_response
<Py42Response [status=204, data='']>
```

Both Org and Device managers are `UserDict` subclasses, and use some data descriptor classes to help manage changes to the underlying values of the dict, while also allowing for value conversion and for all changes to be tracked.

Because an Org's "device defaults" config is essentially the same as a single device's, the `DeviceSettings` is also used to manage an Org's device defaults, so the user API is consistent. 

To manage a device's primary backup set and add the "D:\" drive to the file selection:

```
>>> d = sdk.devices.get_settings(<guid>)
>>> d.backup_sets
[<BackupSet: id: 1, name: 'Test Environment - Backup Backup Set'>]
>>>
>>> d.backup_sets[0].included_files
['C:/Users/admin/']
>>> d.backup_sets[0].included_files.append("D:/")
>>> d.changes
{'included_files': "['C:/Users/admin/'] -> ['C:/Users/admin/', 'D:/']"}
>>>
>>> sdk.devices.update_settings(d)
<Py42Response [status=200, data={'active': True, 'address': '192.168.74.138:4247', 'alertState': 0, 'alertStates': ['OK'], ...}]>
```

To add the "D:\" drive to an org's default primary backup set, the equivalent `DeviceSettings` is stored on the org manager's `.device_defaults` attribute:

```
>>> o = sdk.orgs.get_settings(<org_id>)
>>> o.device_defaults.backup_sets[0]
<BackupSet: id: 1, name: 'Next-Gen Platinum Support Test Environment - Backup Backup Set'>
>>> o.device_defaults.backup_sets[0].included_files
[':user']
>>> o.device_defaults.backup_sets[0].included_files.append("D:/")
>>> o.changes
{'included_files': "[':user'] -> [':user', 'D:/']"}
>>> sdk.orgs.update_settings(o)
SettingsManagerResponse(error=False, settings_response=<Py42Response [status=200, data={'active': True, 'blocked': False, 'classification': 'BASIC', 'configInheritanceCounts': {}, ...}]>, org_settings_response=None)
```

### Issues Resolved ###
Resolves #154


src/py42/util.py
```diff
@@ -72,3 +72,30 @@ def format_dict(dict_, label=None):
...
+
+def bool_to_str(value):
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    else:
+        raise KeyError("Value must be True or False")
+
+
+def str_to_bool(value):
+    if value == "true":
```
unparalleled-js, line 85: These should call `lowercase` first

src/py42/clients/orgs.py
```diff
@@ -187,3 +408,14 @@ def get_agent_full_disk_access_states(self, guid):
                 :class:`py42.response.Py42Response`: A response containing settings information.
             """
         return self.get_agent_state(guid, u"fullDiskAccess")
+
+    def put_to_org_endpoint(self, org_id, data):
+        uri = "/api/Org/{}".format(org_id)
```
unparalleled-js, line 413: Missing `u` here

src/py42/clients/orgs.py
```diff
@@ -187,3 +408,14 @@ def get_agent_full_disk_access_states(self, guid):
...
+        return self._session.put(uri, data=json.dumps(data))
+
+    def put_to_org_setting_endpoint(self, org_id, data):
```
unparalleled-js, line 416: Are these `put` methods only used for settings? They seem quite generic. Should they maybe be internalized with underscores?

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+            self.org_response = self._org_client.put_to_org_endpoint(
+                self.org_id, data=settings_payload
+            )
+        except Py42Error as e:
+            self.errored = True
+            self.org_response = e
+
+    def _update_org_settings(self):
+        packet_list = list(self._t_settings.maps[0].values())
+        org_settings_payload = {"packets": packet_list}
```
unparalleled-js, line 215: Missing `u` here

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+            if value != orig[key]:
+                diff[key] = value
+        return diff
+
+    def _prepare_settings_payload(self):
+        org_updates, org_original = self._org.maps
+        settings_updates, settings_orig = self._settings.maps
+        payload = org_original.copy()
+        payload.update(org_updates)
+        payload["settings"] = settings_orig.copy()
```
unparalleled-js, line 199: Missing `u`'s herre on `settings`

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+        if value:
+            self.endpoint_monitoring_enabled = True
+        self._t_settings["device_fileForensics_enabled"] = {
+            "key": "device_fileForensics_enabled",
+            "value": bool_to_str(value),
+            "locked": False,
+        }
+
+    def update(self):
+        msg = "Updating org_id={}, org_name={}, with changes: {}".format(
```
unparalleled-js, line 175: Missing `u`s a few times on this assignment statement

unparalleled-js: Well, just all over this file. Just make sure to audit these

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+from py42.exceptions import Py42Error
+from py42.settings import debug
+from py42.util import bool_required
+from py42.util import bool_to_str
+from py42.util import str_to_bool
+
+
+class OrgSettingsManager(object):
+    def __init__(self, org_client, org_id):
+        org_response = org_client.get_by_id(org_id, incSettings=True)
```
unparalleled-js, line 16: Is there a way to achieve this without making networking calls in the ctor?

src/py42/util.py
```diff
@@ -72,3 +72,30 @@ def format_dict(dict_, label=None):
...
+        return True
+    if value == "false":
+        return False
+    else:
+        return value
+
+
+def bool_required(f):
```
unparalleled-js, line 93: Could allow this is work for str's where `str_to_bool` works

src/py42/util.py
```diff
@@ -72,3 +72,30 @@ def format_dict(dict_, label=None):
...
+        raise KeyError("Value must be True or False")
+
+
+def str_to_bool(value):
+    if value == "true":
+        return True
+    if value == "false":
+        return False
+    else:
+        return value
```
unparalleled-js, line 90: Why not just return `None` here?

unparalleled-js: or raise `KeyError`

src/py42/util.py
```diff
@@ -72,3 +72,30 @@ def format_dict(dict_, label=None):
     if label:
         return u"{} {}".format(label, indented_dict)
     return indented_dict
+
+
+def bool_to_str(value):
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    else:
+        raise KeyError("Value must be True or False")
```
unparalleled-js, line 81: Is `KeyError` the right error here? I'm not used to thinking of function parameters as keys.

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+        msg = "Updating org_id={}, org_name={}, with changes: {}".format(
+            self.org_id, self._org.maps[1]["orgName"], self.changes
+        )
+        debug.logger.debug(msg)
+        if self.changes["settings"]:
+            self._update_settings()
+        if self.changes["t_settings"]:
+            self._update_org_settings()
+
+        return "Success" if not self.errored else "Error(s) occurred."
```
unparalleled-js, line 184: Could we return the `response` from the `put` instead of `success` ?

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+            "key": "org-securityTools-enable",
+            "value": bool_to_str(value),
+            "locked": False,
+        }
+        self._t_settings["device_advancedExfiltrationDetection_enabled"] = {
+            "key": "device_advancedExfiltrationDetection_enabled",
+            "value": bool_to_str(value),
+            "locked": False,
+        }
+        if not value:  # disable everything but FMC, like the UI does
```
unparalleled-js, line 103: Would it make sense to remove this `if` condition and just have the `self.endpoint_monitoring_removable_media_enabled = value` etc ?

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,225 @@
...
+        self._org_client = org_client
+        self.errored = None
+        self.org_response = None
+        self.org_settings_response = None
+
+    @property
+    def changes(self):
+        changes_dict = {}
+        changes_dict["settings"] = self._diff_chainmap(self._org) or {}
+        changes_dict["settings"].update(self._diff_chainmap(self._settings))
```
unparalleled-js, line 34: Is this a network call here?  Perhaps getting the changes should be a method instead of a property if this is the case. It seems to do a lot, which might be surprising if accessing as a property.

src/py42/util.py
unparalleled-js: How about `TypeError` ?

src/py42/util.py
timabrmsn: These bool converters are for the funky-ness of the `OrgSettings` API. The `value` field of the t_settings packets that are submitted for changes (e.g. `"key": "org-securityTools-enable", "value": "false", "locked": false}`). So rather than make the end-user of py42 set the attribute as the strings `true` or `false` we're auto-converting them to booleans when returning the attribute and auto-converting them back to strings when setting them.

I suppose since the method is so specific to these packet values, I could make them methods on the class so they won't be used elsewhere and we can just assume that the inputs are always going to be what we expect.

src/py42/clients/orgs.py
timabrmsn: Yeah, they're only used for modifying org settings. 

I guess the internalization question depends on how restrictive we want to be for end-users of py42 making any lower-level calls themselves. Do we only let them change what settings we've implemented (and if they _really_ need to they can use the internal `._session` object)? I'm more of the mind to say "here's the easy way to do it (using the OrgSettingsManager) or if you have something more you need to do manually, here's the basic low-level method".

src/py42/util.py
unparalleled-js: It doesn't hurt! It just always seems like the right thing to do. I figured it wasn't necessary in this case.

src/py42/clients/orgs.py
timabrmsn: Yeah, that's how I initially did it (had `sdk.orgs.get_settings_manager()` make the calls then `OrgSettingsManager` just took the two settings dicts as arguments). I thought it might be useful for people to be able to construct the object themselves without having to make custom GET requests, but it's a bit awkward since you still need to pass in the org_client.. Maybe I'll change it back so there's only one way to get OrgSettingsManagers and that's through `.get_settings_manager()`

src/py42/clients/orgs.py
unparalleled-js: Maybe just renaming them to like `update_settings()`?

src/py42/util.py
timabrmsn: I think I was letting the `else` capture if a boolean was passed that we'd just let it through, but it might be better to raise since this func is making an explicit conversion assumption and we wouldn't want to hide a bug by it being used incorrectly.

src/py42/util.py
timabrmsn: yeah `TypeError` is better.

src/py42/clients/orgs.py
timabrmsn: Well, there could be two responses (one for `/api/Org` and one for `/api/OrgSetting`), or two exceptions, or one successful response and one exception..

unparalleled-js: I see.

What about a named tuple containing both responses?

unparalleled-js: I am not completely opposed to this, just generally prefer things to be functional if possible, rather than state objects. The end users could decide if they need to hold on to responses.

unparalleled-js: If we stick with how it is now:

"Success" is good in the python interpreter but a bool would be better for better for scripts I think

unparalleled-js: Or a generator? And then to use the Settings Manager, you would have to loop over the update call. That way works nicely with either 1 or 2 responses.

unparalleled-js:
Did Cecilia and her team request this feature? We should get their input as well, as they will be using this.

src/py42/clients/orgs.py
timabrmsn: These are a bit of an oddity, since it's trying to somewhat mimic the UI checkboxes for these settings. In the console UI, if you click the "enable endpoint monitoring" box, it auto-checks all the other boxes, same if you uncheck it. 

Since those flags are tied together in the UI, there's no way to have "endpoint monitoring" disabled, but also have "removable media" enabled. But you can craft an API call where they are that way. So I have it where if you set the `removeable_media_enabled` attribute to `True`, it sets `endpoint_monitoring_enabled` to true (since it has to be). But if `endpoint_monitoring_enabled = True` means _everything_ gets set to true then we'd have to have another method of setting _just_ the values (there are two) which coincide with the endpoint monitoring checkbox itself in the UI. 

So it's a slight departure from the UI, in that `endpoint_monitoring_enabled = True` only sets the minimum set of keys, where `endpoint_monitoring_enabled = False` sets everything to `False`, but I think it is a good balance between prevent incorrect states while also making the user explicitly set the things they want to be on.

ceciliastevens:
> Did Cecilia and her team request this feature? We should get their input as well, as they will be using this.

I do believe this request originated from PS, yes. I like the approach taken here, to the extent I understand it. It looks like it will be intuitive to the end-user while being extensible in the way we want.

src/py42/clients/orgs.py
timabrmsn: No, this is more of a helper to be able to inspect what is actually being updated. It gets the diffs of the two dicts in each ChainMap object and only shows the keys that actually were modified. This is used both in logging the changes and in deciding whether to actually make any API calls in the `.update()` method.

src/py42/clients/orgs.py
timabrmsn: Yeah, there's not really a "good" solution here. I do like the namedtuple idea, it could have a flag for if there was an error, and then the two response objects on it for inspection.

alanag13: I also like the idea of a namedtuple here.

src/py42/clients/orgs.py
alanag13: I'm fine with these methods being exposed, but I think they need a more intuitive name. What does calling each of these endpoints _do_?

src/py42/clients/orgs.py
```diff
@@ -1,8 +1,224 @@
...
+    def endpoint_monitoring_file_metadata_enabled(self, value):
+        if value:
+            self.endpoint_monitoring_enabled = True
+        self._t_settings[u"device_fileForensics_enabled"] = {
+            u"key": u"device_fileForensics_enabled",
+            u"value": bool_to_str(value),
+            u"locked": False,
+        }
+
+    def update(self):
```
alanag13, line 173: should _org, _settings, and _t_settings be cleared after an update command resolves?

alanag13, line 173: To avoid state-related issues, I wonder if we should separate the update command from the structure that is managing all of the state of the settings. Sort of like how we work with queries for alerts/files.
```python
class SettingsManager:
     def propx
     def propy
     def etc

class OrgService:
     def update_settings(settings_manager):
```

src/py42/clients/orgs.py
timabrmsn: The problem with `update_settings()` is that there are two endpoints that can do the "updating of settings", and there's no categorical distinction between which setting goes to which endpoint, apart from how they're stored in the server backend. 

So I kept it very low-level, where the function name is literally what it does, "make a put request to the OrgSetting endpoint", so that if that's what you want/need to do because the settings manager doesn't provide the functionality you need, you can craft your request data manually and make the put without having to dig into the internal session. 

I suppose we could have a single `update_settings()` function that takes two potential `data` arguments, like:

``
update_settings(
    org_id, 
    org_data={"some":"data"}, 
    org_settings_data={"some":"otherdata"}
)
```

Where if both types are passed in it makes requests to both endpoints?

File: src/py42/clients/orgs.py (hunk 14 above)
Comment (timabrmsn) [reply src/py42/clients/orgs.py:173]: I'm not opposed to making the update function external to the manager and having it take a manager as an argument, but I'm not sure what state-related problems that might solve.

Comment (alanag13) [reply src/py42/clients/orgs.py:173]: The "problem" is more of a usability one -- if `update` and the fields are attached to the same object, should I expect that when update is called that all of the properties that I have set have been cleared or not? If the two are separated I think the expectation that the properties do _not_ get cleared is much higher.

File: src/py42/clients/orgs.py (hunk 3 above)
Comment (alanag13) [reply src/py42/clients/orgs.py:415]: The problem with the current names is that it necessitates a knowledge of the HTTP verbs and url names and what their function is here. If I'm looking for a function that allow me to free-form update these settings and I don't know what endpoints are responsible for that (with py42 you shouldnt have to), these don't hit to me that they are what should be used.

I like your proposal since the name does indicate that. It's unfortunate that we will always have the problem of not knowing what the correct setting "destination" is, but that's still an improvement.

Comment (timabrmsn) [reply src/py42/clients/orgs.py:415]: Yeah, but the complexity of the data that you need to post to both of those endpoints is so high that knowing the verb and url name is inconsequential. 

I think maybe we should just restrict all py42 settings changes to the manager(s) and keep these private.

File: src/py42/clients/orgs.py (hunk 14 above)
Comment (timabrmsn) [reply src/py42/clients/orgs.py:173]: I'm not quite sure what you're meaning by "cleared" here. 

Changes to the properties ultimately change the underlying dict that will be the body to be posted, which already exists with the _current_ org state before the user makes any changes. So I can check `orgsetting.org_name` and see that it's "MyOrg", then update the org name by setting the property `orgsetting.org_name = "MyNewOrg"`, after calling `update()` you'd expect it to go back to "MyOrg"? Or are you just talking about the manager class's state knowledge of what's been changed? So after update, all the properties reflect what the org currently is now, but now there's no values in the `.changes` property?

Comment (alanag13) [reply src/py42/clients/orgs.py:173]: The latter is closer to what I referring to. Imagine this scenario:
```python
# make a bunch of changes to the settings manager

settings_mgr.update()

# if it worked, do some other logic that helps me decide if I should change
# some additional settings

settings_mgr.some_other_settings = true

settings_mgr.upate()

# were all of the previous properties that I sent the first time I called `update()` resubmitted or not?
# did `update()` "unstage" the changes that I made successfully earlier?
```

The outcome becomes a little more clear (at least to in my opinion) when we separate things a bit and just treat the settings manager as a typical dict / struct:
```python
# make a bunch of changes to the settings manager

org_client.update_settings(settings_mgr)

# if it worked, do some other logic that helps me decide if I should change
# additional settings

settings_mgr.some_other_setting = true

# oops, I'm using the same object as I was before here, so this will _probably_ result in me
# re-sending the previous properties...
org_client.upate_settings(settings_mgr)

settings_mgr = SettingsMgr()
settings.new_setting = True

# now I know for sure that only new_setting will be sent
org_client.update_settings(settings_mgr)
```

alanag13: I understand that both approaches will have the same result regardless of whether or not things are "unstaged" in the first scenario. My point is that users do not understand that this an abstraction over submitting the entire current org state, and could be less confident in what the outcome of what theyre running will be in that case.

alanag13: I'm probably splitting hairs a little bit here and I'm not going to take an aggressive stand on this -- it's a thought, not a requirement, and this implementation does what it's supposed to do well.

src/py42/clients/orgs.py
timabrmsn: I decided to remove the public PUT methods and just make all device/org settings changes go through the manager classes being passed to `update_org_settings()` and `update_device_settings()`.

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+def get_val(d, keys):
+    """Helper for getting nested values from a dict based on a list of keys."""
+    for key in keys:
+        d = d[key]
+    return d
+
+
+def bool_to_str(value):
+    if isinstance(value, bool):
+        return "true" if value else "false"
```
unparalleled-js, line 22: Could do `str(value).lower()`

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+def get_val(d, keys):
+    """Helper for getting nested values from a dict based on a list of keys."""
+    for key in keys:
+        d = d[key]
+    return d
+
+
+def bool_to_str(value):
```
unparalleled-js, line 20: Are ints supported for boolean settings?

src/py42/clients/devices.py
```diff
@@ -265,3 +268,30 @@ def get_agent_full_disk_access_state(self, guid):
...
+        Args:
+            guid (int,str): The globally unique identifier of the device.
+
+        Returns:
+            :class:`py42.clients._settings_managers.DeviceSettingsManager`: A class to help manage device settings.
+        """
+        settings = self.get_by_guid(guid, incSettings=True)
+        return DeviceSettingsManager(settings.data)
+
+    def update_device_settings(self, settings_manager):
```
unparalleled-js, line 284: Instead of passing the whole manager, is there an object that would make more sense? Like a `settings_update` object that a manager can create that has everything it needs?  This works fine, but just feels strange from an API perspective.

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    elif value in ("true", "false"):
+        return value
+    else:
+        raise ValueError("Value must be True or False")
+
+
+def str_to_bool(value):
```
unparalleled-js, line 29: Perhaps these should be internal prefixed? Else I am tempted to re-comment that you should call `lower()`

unparalleled-js: Otherwise, these might be more generally useful and could be add to `util`

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+    if value == "true":
+        return True
+    if value == "false":
+        return False
+    else:
+        return value
+
+
+def days_to_minutes(days):
+    return str(int(days) * 1440)
```
unparalleled-js, line 39: Is the expected behavior to always round floats down here?

```int(1.9)
>>> 1
```

unparalleled-js: Might it make more sense to do `str(int(days * 1440.0))` or `str(int(round(days * 1440.0)))`

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+
+class SettingProperty(object):
+    """Descriptor class to help manage changes to nested dict values. Assumes attributes
+    being managed are on a UserDict/UserList subclass.
+
+    Args:
+        name (str): name of attribute this class manages (changes will be registered with this name).
+        location (list): list of keys defining the location of the value being managed in the managed class.
+        get_converter (func, optional): function to convert retrieved values to preferred format.
+        set_converter (func, optional): function to convert values being set to preferred format.
```
unparalleled-js, line 63: Missing `. Default to ...` statements here

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+            init_val = get_val(instance.data, self.location)
+            self.init_val = self.get_converter(init_val)
+        if self.init_val == new_val:
+            if self.name in instance.changes:
+                instance.changes.pop(self.name)
+        else:
+            instance.changes[self.name] = show_change(self.init_val, new_val)
+
+
+class TSettingProperty(object):
```
unparalleled-js, line 96: What does t_settings mean? Type settings?

src/py42/clients/_settings_managers.py
```diff
@@ -0,0 +1,520 @@
...
+        value = self.data.pop(index)
+        self.register_change()
+        return value
+
+    def remove(self, value):
+        self.data.remove(value)
+        self.register_change()
+
+
+class OrgSettingsManager(UserDict):
```
unparalleled-js, line 174: this plays into my other comment about the method param on DeviceClient etc...

Maybe these objects could actually be classified as mere `OrgSettings` rather than `OrgSettingsManager`.

src/py42/clients/_settings_managers.py
timabrmsn: Not anywhere that I know of in our settings APIs.

src/py42/clients/devices.py
timabrmsn: The manager is really just the settings response dict at its core with helper attributes and a few other things to make modifying the nested values easier, and make it so the user doesn't need to know the structure of the dict/json at all.

Technically this function could take a regular dict of device settings, the only change needed would be to call `dict()` on the input arg instead of passing `settings_manager.data` to the PUT request.

src/py42/clients/_settings_managers.py
timabrmsn: Initially I was assuming the web UI only allowed you to enter whole digit days, but I guess you can enter floats. So I've updated it. Good catch.

src/py42/clients/_settings_managers.py
timabrmsn: It's for settings that get stored in the `t_settings` table of the database. The reason that there are two endpoints each for modifying devices/orgs is because there's a fundamental difference to the backend in how they're stored, so they have very different APIs for accessing and updating. I don't think there's an "official" name for these settings, they're commonly just referred to as t_settings.

The `/api/Org` and `/api/Device` endpoints can basically take the same json that you get from a GET request, and send that back with modifications in a PUT and it just works. However, the `/api/DeviceSettings` and `/api/OrgSettings` endpoints return results as a dict of keys like this:

```
  "data": {
    "device plan key": {
      "scope": "DEVICE",
      "value": "foo",
      "locked": false,
      "id": 2
    },
    "system key": {
      "scope": "SYSTEM",
      "value": "foo",
      "locked": false
    }
```
But require updates to be sent in a list of "packets" like this:

```
{
  "packets": [
    {
      "key": <key1>,
      "value": <value1>,
      "locked": <true/false>
    },
    {
      "key": <key2>,
      "value": <value2>,
      "locked": <true/false>
    },
    ...
  ]
}
```

So the `TSettingsProperty` is a helper that does the tracking and translation from the response data into packets suitable for posting. 

Since both `/api/OrgSettings` and `/api/DeviceSetting` both use the same table and API, I figured I'd name it after the technical backend they share... although I didn't implement any of the device t_settings in the device manager (since there's only a small handful that don't actually get changed much), so that shared-ness isn't clear from the existing code.

src/py42/clients/_settings_managers.py
timabrmsn: Yeah, I suppose the "manager" name made more sense when it had its own `.update()` method that actually made the http requests. I don't have a problem dropping it.

kiran-chaudhary (commented):
Shared some observations around `conversions`. 

Some technical documentation around design decisions and assumptions would be helpful.

Also the explanation made in PR description should be added to Docs.

.vscode/settings.json
```diff
@@ -7,5 +7,6 @@
     "python.testing.nosetestsEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.linting.flake8Enabled": true,
-    "python.linting.enabled": true
+    "python.linting.enabled": true,
+    "python.pythonPath": "/Users/tim.abramson/PycharmProjects/venvs/py42/bin/python"
```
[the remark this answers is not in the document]
timabrmsn: whoops, that file shouldn't have been committed.

src/py42/clients/settings/converters.py
```diff
@@ -0,0 +1,48 @@
+from py42._compat import str
+
+
+def bool_to_str(value):
+    if isinstance(value, bool) or value in ("true", "false"):
+        return str(value).lower()
+    else:
+        raise ValueError("Value must be True or False")
```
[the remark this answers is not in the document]
timabrmsn: So the goal with these converters is to smooth out the oddities of the settings APIs. In this case the fact that the required json values need to be the strings `"true"` and `"false"` instead of json booleans. We don't want to have end-users of py42 have to remember which values need to be strings and which need to be booleans, so we do the conversion for them, so they can just do: 

```
if org_setting.some_property:
     org_setting.some_property = False
```

But if they _do_ happen to assign the string value, we just use it. 

And for why this one raises an AttributeError, is because it's being used as a `set_converter` on many `SettingProperty` classes. The ones that don't raise AttributeError are only being used as `get_converter`s to transform the current values (which we expect to never be invalid since we're getting them from the server). 

Basically, the goal is to make sure the end user doesn't have to know the ins and outs of the underlying nested settings dict and all it's quirks, just that each "setting" that's available in the console UI has a corresponding property (or sometimes method) on the `OrgSettings`/`DeviceSettings` object. The property classes then do the work to convert values _from_ the underlying dicts to something user-friendly, as well as convert the user-friendly representation back to what the settings dicts require.

src/py42/clients/settings/converters.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def str_to_bool(value):
+    if value.lower() == "true":
+        return True
+    if value.lower() == "false":
+        return False
+    else:
+        return value
```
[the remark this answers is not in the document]
timabrmsn: Initially this was used in a way where the value could have been "true", "false" or something else, and I wanted it to just convert the string-ified booleans to true booleans and let through anything else untouched, but now it is only used when we expect to do the conversions, so I suppose it should probably raise an exception so it doesn't get stuck somewhere unexpected..

src/py42/clients/settings/converters.py
```diff
@@ -0,0 +1,48 @@
...
+def days_to_minutes(days):
+    return str(int(float(days) * 1440))
+
+
+def minutes_to_days(minutes):
+    minutes = int(minutes)
+    return int(minutes / 1440)
+
+
+def no_conversion(x):
```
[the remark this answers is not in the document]
timabrmsn: It's used as the default converter for `SettingProperty` instances if none are passed.

src/py42/clients/settings/converters.py
```diff
@@ -0,0 +1,48 @@
...
+    return value.split(",")
+
+
+def days_to_minutes(days):
+    return str(int(float(days) * 1440))
+
+
+def minutes_to_days(minutes):
+    minutes = int(minutes)
+    return int(minutes / 1440)
```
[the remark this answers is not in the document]
timabrmsn: Yep, because the underlying setting dict stores the values as strings, but we want py42 end-users to be able to interact with them as numbers and not have to convert them manually themselves.

src/py42/clients/settings/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+
+def set_val(d, keys, value):
+    """Helper for setting nested values from a dict based on a list of keys."""
+    d = get_val(d, keys[:-1])
+    d[keys[-1]] = value
+
+
+def get_val(d, keys):
+    """Helper for getting nested values from a dict based on a list of keys."""
+    for key in keys:
```
unparalleled-js, line 15: What is going on here?  Why not just set `d = d[keys[-1]]` ?

src/py42/clients/settings/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Args:
+        name (str): name of attribute this class manages (changes will be registered with this name).
+        location (list): list of keys defining the location of the value being managed in the managed class.
+    """
+
+    def __get__(self, instance, owner):
+        val = get_val(instance.data, self.location)
+        if isinstance(val, dict):
+            return str_to_bool(val["@locked"])
```
unparalleled-js, line 96: Could consider using constants instead of repeating string values

src/py42/clients/settings/__init__.py
timabrmsn: Because there can be more than one level of nesting. This allows us to define each setting property with just the list of keys as the "location" of the value we're looking to get/set: 

```
backup_status_email_frequency_days = SettingProperty(
        name="backup_status_email_frequency_days",
        location=["settings", "serviceBackupConfig", "backupStatusEmailFreqInMinutes"],
        get_converter=minutes_to_days,
        set_converter=days_to_minutes,
    )
```

unparalleled-js: Oh I see now

unparalleled-js: It reminds me of the Ruby `dig` method: https://tiagoamaro.com.br/2016/08/27/ruby-2-3-dig/

src/py42/clients/settings/converters.py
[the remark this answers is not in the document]
unparalleled-js: Right, @kiran-chaudhary `True` just means the boolean here, not the string

docs/userguides/devicesettings.md
```diff
@@ -0,0 +1,63 @@
...
+12345
+>>> device_settings.guid
+908765043021
+>>> device_settings.org_id
+42
+>>> device_settings.user_id
+494842
+>>> device_settings.version
+1525200006800
+>>> device_settings.available_destinations
```
unparalleled-js, line 27: Ha cool

src/py42/clients/settings/__init__.py
```diff
@@ -0,0 +1,199 @@
...
+
+    Args:
+        name (str): name of attribute this class manages (changes will be registered with this name).
+        location (list): list of keys defining the location of the value being managed in the managed class.
+        get_converter (func, optional): function to convert retrieved values to preferred format. Defaults to no conversion.
+        set_converter (func, optional): function to convert values being set to preferred format. Defaults to no conversion.
+    """
+
+    def __init__(
+        self, name, location, get_converter=no_conversion, set_converter=no_conversion
```
unparalleled-js, line 57: Instead of `no_conversion()`, since it is only used as a default, these could just be empty lambdas... `..., get_converter=lambda x: x, set_converter=lambda x: x)`

src/py42/clients/settings/_converters.py
```diff
@@ -0,0 +1,48 @@
...
+        return ",".join(value)
+    else:
+        return value
+
+
+def comma_separated_to_list(value):
+    return value.split(",")
+
+
+def days_to_minutes(days):
```
unparalleled-js, line 38: Is days always an int?

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,349 @@
...
+        get_converter=str_to_bool,
+        set_converter=bool_to_str,
+    )
+    warning_alert_days = SettingProperty(
+        name="warning_alert_days",
+        location=["settings", "serviceBackupConfig", "minutesUntilWarning"],
+        get_converter=minutes_to_days,
+        set_converter=days_to_minutes,
+    )
+    critical_alert_days = SettingProperty(
```
unparalleled-js, line 54: It must `critical` in the UI I am guessing? But `severe` on the API

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,349 @@
...
+        self._destinations = device_dict["availableDestinations"]
+        self.backup_sets = [
+            BackupSet(self, self.changes, backup_set)
+            for backup_set in self.data["settings"]["serviceBackupConfig"][
+                "backupConfig"
+            ]["backupSets"]
+        ]
+
+    @property
+    def computer_id(self):
```
unparalleled-js, line 92: This should probably be `device_id`. We call `computer_id` `device_id` in other places in py42.

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+class BackupSet(UserDict, object):
+    """Helper class for managing device backup sets and Org device default backup sets."""
+
+    def __init__(self, settings_manager, changes_dict, backup_set_dict):
+        self._manager = settings_manager
+        self._changes = changes_dict
+        self.data = backup_set_dict
+        includes, excludes = self._extract_file_selection_lists()
```
unparalleled-js, line 130: there are a lot of methods in the class... It'd be easier to review if the internal ones were ordered in the order of the usage so I can follow along easier

src/py42/clients/settings/__init__.py
timabrmsn: Maybe I'll just set the default to `None` and then only call the converter if it's not `None`.

docs/methoddocs/backupset.md
```diff
@@ -0,0 +1,7 @@
+# Backup Sets
```
unparalleled-js, line 1: `Backup Set` or `Backup Sets`, I am seeing both ways here

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+        ]
+
+    @property
+    def available_destinations(self):
+        """Returns a dict of destinations available to be used by devices. Dict keys are
+        destination guids and values are destination names.
+        """
+        return {d["guid"]: d["destinationName"] for d in self._destinations}
+
+    warning_email_enabled = SettingProperty(
```
unparalleled-js, line 39: Do these properties show up in the doc strings?

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+            raise Py42Error(
+                "Invalid destination guid or destination not offered to device's Org."
+            )
+
+    def remove_destination(self, destination_guid):
+        """Removes a destination from use by this backup set.
+
+        Args:
+            destination_guid (str, int): The globally unique identifier of the
+            destination to be removed.
```
unparalleled-js, line 260: If I can recall, this will not look good in the docs w/o indentation for the second line

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+            for d in self.data["destinations"]:
+                if d["@id"] == destination_guid:
+                    self.data["destinations"].remove(d)
+            self._changes["destinations"] = show_change(
+                self._orig_destinations, self.destinations
+            )
+
+    def lock_destination(self, destination_guid):
+        """Locks an in-use destination, disallowing the device owner from removing this
+        destination from their backup. Raises an exception if the supplied destination
```
unparalleled-js, line 274: Perhaps we should say raises `Py42Error` so that users can catch it if they need to and they know which one to catch from looking at the docs

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+        for d in self.data["destinations"]:
+            guid = d["@id"]
+            dest_name = self._manager.available_destinations[guid]
+            if "@locked" in d:
+                dest_name = dest_name + " <LOCKED>"
+            destination_dict[guid] = dest_name
+        return destination_dict
+
+    def add_destination(self, destination_guid):
+        """Adds a destination to be used by this backup set. Raises an exception if
```
unparalleled-js, line 236: would be better to specify the exception class here

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+        destination_guid = str(destination_guid)
+        if destination_guid in self._manager.available_destinations:
+            if destination_guid not in self.destinations:
+                self.data["destinations"].append({"@id": destination_guid})
+                self._changes["destinations"] = show_change(
+                    self._orig_destinations, self.destinations
+                )
+        else:
+            raise Py42Error(
```
unparalleled-js, line 251: I see this same exception being re-used... If we need to change the text, we will have to do in multiple places.  This should probably be a custom exception.

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+
+    def lock_destination(self, destination_guid):
+        """Locks an in-use destination, disallowing the device owner from removing this
+        destination from their backup. Raises an exception if the supplied destination
+        guid is not in use on this backup set, or not available to the parent device/org.
+        """
+        destination_guid = str(destination_guid)
+        if destination_guid in self._manager.available_destinations:
+            if destination_guid not in self.destinations:
+                raise Py42Error("Destination is not added to device, unable to lock.")
```
unparalleled-js, line 280: The exception is als re-used a bunch, which could make it complicated to maintain if we need to change some of it's text, so it should be re-usable somehow, either via custom exception or a local method to raise this error

src/py42/clients/settings/device_settings.py
```diff
@@ -0,0 +1,415 @@
...
+        if isinstance(pathset, dict):
+            pathset = [pathset]
+        includes = [p["@include"] for p in pathset if "@include" in p]
+        excludes = [p["@exclude"] for p in pathset if "@exclude" in p]
+        return includes, excludes
+
+    def _build_file_selection(self):
+        if not self._included_files:
+            self.data["backupPaths"]["pathset"] = {
+                "paths": {"@cleared": "true", "@os": "Linux", "path": []}
```
unparalleled-js, line 330: Why is the Linux the default?

unparalleled-js:
Consider going making all strings unicode prefixed!  Looks really really good, I don't see anything that is too bad, some small areas to improve code sharing re: exceptions, but that is pretty much it. I'll approve, consider all feedback.

unparalleled-js:
Actually, I just noticed there is no CHANGELOG entry for this

src/py42/clients/settings/_converters.py
timabrmsn: Days here will be user supplied when they're setting the attribute, so we don't know. I call `float()` on the input here to be able to handle an input string like `"1.5"`, but also since we want to raise if it's not numeric, the `float()` call will do that for us. 

But I did update `minutes_to_days` to return a float value rather than rounding down, since you _can_ save partial days in the UI.

src/py42/clients/settings/device_settings.py
timabrmsn: Yep, there are a lot of mismatches between what the endpoint expects and what the UI calls it, so we're always mimicking the UI.

src/py42/clients/settings/device_settings.py
timabrmsn: The docstrings below each property do get rendered correctly in the sphinx docs. But I just checked running `help()` on a `DeviceSettings` instance in the repl and it shows the docstring of the `SettingProperty` class there instead.. I wonder if there's any way to change that.

src/py42/clients/settings/device_settings.py
timabrmsn: done

src/py42/clients/settings/device_settings.py
timabrmsn: done

src/py42/clients/settings/device_settings.py
timabrmsn: done

src/py42/clients/settings/device_settings.py
timabrmsn: done

src/py42/clients/settings/device_settings.py
timabrmsn: I'm not sure, but that's what the UI uses regardless of what OS the client actually is installed on. I'm guessing it was initially designed to allow for the server to handle the paths differently based on OS, but they eventually had the client do all conversions (i.e. `C:/Users` > `C:\Users` on Windows).

src/py42/clients/settings/org_settings.py
```diff
@@ -297,10 +297,20 @@ def endpoint_monitoring_file_metadata_collection_enabled(self, value):
         "file_metadata_collection_exclusions",
         "device_fileForensics_fileExclusions_org",
     )
-    """List of file types and file paths to exclude from file metadata collection.
+    """File types and file paths to exclude from file metadata collection.
 
     See `Support Documentation <https://support.code42.com/Administrator/Cloud/Configuring/File_Metadata_Collection_exclusions>`__
-    for more details.
+    for more details on the expected shape of the body this setting expects.
```
unparalleled-js, line 303: `expected` and `expects` together here feel redundant

src/py42/clients/settings/org_settings.py
```diff
@@ -297,10 +297,20 @@ def endpoint_monitoring_file_metadata_collection_enabled(self, value):
...
+    """
+
+    endpoint_monitoring_file_exfiltration_detection_exclusions = TSettingProperty(
+        "file_exfiltration_detection_exclusions",
+        "org_securityTools_detection_monitoring_exclusions",
+    )
+    """File types and file paths to exclude from file exfiltration detection.
+
+    See `Support Documentation <https://support.code42.com/Administrator/Cloud/Configuring/Endpoint_monitoring#ExcludePaths>`__
+    for more details on the expected shape of the body this setting expects.
```
unparalleled-js, line 313: Same here re: expected, expects

src/py42/clients/settings/device_settings.py
unparalleled-js: @timabrmsn thoughts here?

timabrmsn: I'm a bit reluctant, since `computerId` is the actual key in the underlying setting dict, and since that's the technical term used in our API docs and in code. But you're right, if someone's knowledge of our stuff is only from py42 they'd be inclined to look for `device_id`... 

I suppose we could just expose another property named `device_id` so you can find it regardless of what term you're looking for. Just label one as an alias of the other?

src/py42/clients/settings/device_settings.py
timabrmsn: Yeah, I'm working on a bit of cleanup on that front and adding some docstrings to make it clear why these exist now.

unparalleled-js (approved):

PR #173, opened 2020-08-18
Archive client

DiscoRiver:
### Description of Change ###

I've updated ArchiveClient and ArchiveModule with functions to obtain archive information from query values accepted by the Archive API (https://www.crashplan.com/apidocviewer/#Archive)

User-facing functions are;
- `sdk.archive.get_by_archive_guid`
- `sdk.archive.get_by_device_guid`
- `sdk.archive.get_by_user_uid`
- `sdk.archive.get_by_store_point_id`
- `sdk.archive.get_by_destination_guid`
- `sdk.archive.get_by_plan_uid`
- `sdk.archive.get_by_user_uid_list`

For updates to ArchiveClient, there were some considerations;

- Can't specify an archiveGuid value through parameters
- Query values are numbers except the list of userUid values, which are strings.

Because of the above, the following methods have been added;
- `archive_client.get_single_archive` 
  - Queries a single archiveGuid
- `archive_client.get_archives_from_value`
  - Handles the vast majority of query values, and sets params dynamically from `id_type`.
- `archive_client.get_archives_from_user_uid_list`
  - Handles the comma-separated list of userUid strings

### Testing Procedure ###

Test have been written against the 3 new methods in ArchiveClient. Each test provide it's own parameters.


DiscoRiver:
@unparalleled-js - I didn't add a `get_all()` method because the Archive API doesn't provide a generic "get everything" query. One has to specify a query parameter.

I did check the v3+ APIs to see if something had been updated/added for this, but couldn't find anything. 

If I've missed something, let me know!

unparalleled-js:
> @unparalleled-js - I didn't add a `get_all()` method because the Archive API doesn't provide a generic "get everything" query. One has to specify a query parameter.
> 
> I did check the v3+ APIs to see if something had been updated/added for this, but couldn't find anything.
> 
> If I've missed something, let me know!

We can always add it later! But, for your information, when we do `get_all()` methods in py42, we are looping over the pages for the user ourselves via a python generator.  Check this out when you get a chance: https://py42docs.code42.com/en/stable/userguides/basics.html#paging

^ That is a py42 user guide explaining how to do paging using our `get_all()` methods.  I actually fought to get the `get_page()` methods publicly exposed because sometimes it can be helpful to manually request pages.

src/py42/_internal/clients/archive.py
```diff
@@ -10,6 +11,76 @@ def get_data_key_token(self, device_guid):
...
+        uri = u"/api/Archive/{}".format(archive_guid)
+        return self._session.get(uri)
+
+    def get_page(self, page_num, page_size=None, **kwargs):
+        """Gets an individual page of archives.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#Archive-get>`__
+
+        Args:
+            page_num (int): The page number to request.
+            page_size (int, optional): The number of archives to return per page.
```
unparalleled-js, line 33: Make sure to end optional parameter doc strings with `Defaults to None` (or whatever the default value is),

src/py42/_internal/clients/archive.py
```diff
@@ -10,6 +11,76 @@ def get_data_key_token(self, device_guid):
...
+        Args:
+            archive_guid (int): The GUID for the archive.
+
+        Returns:
+            :class:`py42.response.Py42Response`: A response containing archive information.
+        """
+        uri = u"/api/Archive/{}".format(archive_guid)
+        return self._session.get(uri)
+
+    def get_page(self, page_num, page_size=None, **kwargs):
```
unparalleled-js, line 27: Consider enumerating more arguments here. Perhaps not the deprecated ones.

src/py42/_internal/clients/archive.py
```diff
@@ -10,6 +11,76 @@ def get_data_key_token(self, device_guid):
...
+        uri = u"/api/Archive"
+        page_size = page_size or settings.items_per_page
+        params = dict(
+            pgNum=page_num,
+            pgSize=page_size,
+            **kwargs
+        )
+        return self._session.get(uri, params=params)
+
+    def get_archives_from_value(self, id_value, id_type):
```
unparalleled-js, line 47: If this is returning a generator, it needs to follow our standard of being prefixes with `get_all_`.

Perhaps `get_all_archives_from_value()`.

src/py42/_internal/clients/archive.py
```diff
@@ -10,6 +11,76 @@ def get_data_key_token(self, device_guid):
...
+        """Gets archive information from a query value.
```
unparalleled-js, line 48: How about `Gets archives information from an ID, such as a User UID, a Device GUID, or the archive GUID.`

unparalleled-js: Or, in the `args` section, enumerate the supported types

src/py42/_internal/clients/archive.py
```diff
@@ -10,6 +11,76 @@ def get_data_key_token(self, device_guid):
...
+            that each contain a page of archives.
+                """
+        params = {u"{}".format(id_type): u"{}".format(id_value)}
+        return get_all_pages(
+            self.get_page,
+            u"archives",
+            **params,
+        )
+
+    def get_archives_from_user_uid_list(self, user_uid_list):
```
unparalleled-js, line 67: Could this method be combined with the one above somehow?

src/py42/_internal/clients/archive.py
unparalleled-js: Right now, these docs are not publicly facing but they actually might be soon, else I would have just suggested removing them

src/py42/modules/archive.py
```diff
@@ -7,6 +7,114 @@ def __init__(self, archive_accessor_manager, archive_client):
...
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#Archive-get>`__
+
+        Args:
+            store_point_id (int): The storePointId for a storepoint.
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response`
+            objects that each contain a page of archives.
+        """
+        return self._archive_client.get_archives_from_value(store_point_id,
```
unparalleled-js, line 61: Some of the line wrapping on this file differs from how our formatters do it.

src/py42/modules/archive.py
```diff
@@ -7,6 +7,114 @@ def __init__(self, archive_accessor_manager, archive_client):
...
+            plan_uid (int): The planUid for a device.
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response`
+            objects that each contain a page of archives.
+        """
+        return self._archive_client.get_archives_from_value(plan_uid,
+                                                            u"planUid")
+
+    def get_by_uid_list(self, user_uid_list):
```
unparalleled-js, line 105: `get_by_user_uids()` ?

unparalleled-js: I don't really like that there is a separate method expecting a list.  We should just check the if it not a list using the `isinstance` builtin method, and if it is not a list, that wrapping it between `[]`

```
if not isinstance(user_uids, (list, tuple)):
    user_uids = [user_uids]
```

unparalleled-js:
2 Main things here:

1.  Let's make the method expecting a list and single value the same method and using `isinstance` to wrap the results in a list if it is single value. See above comments!

2. For methods returning `generator` type, they should be prefixed with `get_all()`.  I am going to update CONTRIBUTING.md to say this.

alanag13:
Hey @DiscoRiver, some thoughts here -- I how do you feel about reducingsthe methods that you have above to just the following?
```
sdk.archive.get_by_archive_guid
sdk.archive.get_by_device_guid
sdk.archive.get_by_user_uid
sdk.archive.get_by_destination_guid
```
The reason for this would be to eliminate the exposure of methods/variables that users are unlikely to recognize or use since they would probably serve mostly just to confuse. 

`user_list` seems useful but has an api that is not in line with the rest of this sdk (or the rest of the apis). I would rather have users simply request each user necessary in a loop.

One of the main goals of this project is to create a more simple, predictable interface for users rather than to mirror the API's interface, which we've all come to know as somewhat cumbersome.

src/py42/modules/archive.py
alanag13: See my note in the main comment section, I would actually prefer that this method not exist at all, though i'm of course open to listening if theres a particularly strong use case for it.

src/py42/modules/archive.py
alanag13: draft PRs dont go through ci/cd. @DiscoRiver, be sure to read through `CONTRIBUTING.md` to get an environment set up that will perform all of your style checks, auto formatting, etc.

DiscoRiver:
@alanag13 - I don't have a strong case for keeping `user_list`, or any of the other methods you suggested removing. As you said, users can just as easily loop through the users they want to query, and I'm all for keeping things simple. 

I'll work on that, and the other points raised tomorrow.

src/py42/_internal/clients/archive.py
DiscoRiver: I considered updating it to `get_page(self, page_num, device_guid=None, user_uid=None, destination_guid=None, page_size=None, **kwargs):`, but the new arguments wouldn't get used because they're passed as `**kwargs`. 

Since `get_page` should only ever get one of the values `device_guid`, `user_uid`, or `destination_guid`, I'm not sure of the benefit (other than perhaps documentation), unless I've misunderstood something?

src/py42/_internal/clients/archive.py
DiscoRiver: Updated optional parameters to include that.

src/py42/_internal/clients/archive.py
DiscoRiver: Is this simply for documentation purposes? The reason I ask is because currently, the query parameters in `get_all_archives_from_value` are generated from `id_value` and `id_type`, like this;

```
params = {"{}".format(id_type): "{}".format(id_value)}
return get_all_pages(self.get_page, "archives", **params,)
```

If I add the expected parameters as arguments to `get_page` (similar to `UserClient`), `get_all_pages` will complain about receiving duplicate keyword parameters (since it's already being passed as a keyword parameter). Also `get_page` for `ArchiveClient` will only expect one query parameter since the API only supports a single ID.

What I thought was that I would just enumerate the expected arguments in `get_page`, add docstrings for them, but then omit them from `params` in `get_page`, but that doesn't seem like the right way. 

Do you have any suggestions? Or maybe I'm misunderstanding?

src/py42/_internal/clients/archive.py
DiscoRiver: Updated.

src/py42/_internal/clients/archive.py
DiscoRiver: I updated it, but only the initial description as you suggested.

src/py42/_internal/clients/archive.py
DiscoRiver: This method was removed.

src/py42/modules/archive.py
DiscoRiver: Thanks! I forgot to run it! Everything passed now on my latest commits.

src/py42/modules/archive.py
DiscoRiver: This method has been removed.

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+from py42 import settings
 from py42.clients import BaseClient
 from py42.clients.util import get_all_pages
 
 
 class ArchiveClient(BaseClient):
     def get_data_key_token(self, device_guid):
-        uri = u"/api/DataKeyToken"
-        data = {u"computerGuid": device_guid}
+        uri = "/api/DataKeyToken"
```
unparalleled-js, line 10: Why were these `u`'s deleted?

src/py42/_internal/clients/archive.py
unparalleled-js: Ah, I just meant other **kwargs, not the IDs

unparalleled-js: Looks like there is a `srtKey` and a `srtDir` at least

unparalleled-js: Perhaps `export`, which outputs a CSV. That may be useful down the line.

Let me be clear that this was just a suggestion and is not required

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#Archive-get>`__
+
+        Args:
+            page_num (int): The page number to request.
+            page_size (int, optional): The number of archives to return per page. Defaults to `py42.settings.items_per_page`.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        uri = "/api/Archive"
```
unparalleled-js, line 38: Missing `u` prefix on this string

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+        return self._session.get(uri, params=params)
+
+    def get_all_archives_from_value(self, id_value, id_type):
+        """Gets archive information from an ID, such as a User UID, Device GUID, or Destination GUID.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#Archive-get>`__
+
+        Args:
+            id_value (int): Query value for archive.
+            id_type (str): API query value description (e.g backupSourceGuid,
+            userUid, destiationGuid)
```
unparalleled-js, line 50: You need to indent or add a couple of spaces for continued lines on args or else they might not compile correctly

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response` objects
+            that each contain a page of archives.
+                """
```
unparalleled-js, line 55: These quotes are in the wrong spot

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+        params = {"{}".format(id_type): "{}".format(id_value)}
```
unparalleled-js, line 56: Missing `u` prefix here

src/py42/_internal/clients/archive.py
```diff
@@ -1,40 +1,86 @@
...
+        return get_all_pages(self.get_page, "archives", **params,)
+
     def get_backup_sets(self, device_guid, destination_guid):
-        uri = u"/c42api/v3/BackupSets/{}/{}".format(device_guid, destination_guid)
```
unparalleled-js, line 14: As long as we are forced to support python 2, we need unicode prefixes on all strings

unparalleled-js:
Looks like we are missing tests for `ArchiveModule`?

unparalleled-js (changes requested):
See the above comments!  Thank you :)

src/py42/_internal/clients/archive.py
DiscoRiver: It looks like they were all removed when I ran `tox -e style`. I'll watch out for that in the future.

unparalleled-js: Interesting, I hadn't seen that happen before

DiscoRiver: Looks like it's only happening for `src/py42/_internal/clients/archive.py`. The other files are untouched.

PR #174, opened 2020-08-18
Refactor, .netrc support, and socket leak fix

alanag13:
### Description of Change ###

- Adds support for .netrc
- Fixes socket leaking issue

Substantially reorganizes the codebase, namely:
- the notion of `_internal` modules has been removed. This was confusing; the only "internal" objects should be those that have nothing to do with business logic, i.e. session / authentication management modules. Those modules are now prefixed with a `_`.
- "Module" classes are now called Clients. The term "Module" is overloaded in python, and "client" seemed appropriate since it's the user will be most directly interfacing with to make requests.
- What we used to have called Clients are now called Services. This again seemed more fitting since they can be mapped one-for-one to corresponding web services.
- Py42Session (and its corresponding variables) has been renamed Connection to better distinguish it from the `requests` object type `Session`.
- MicroserviceClientFactory has been replaced by a `HostResolver` system that allows Connections to be created before their root domain is known (the connection will resolve the root domain when it makes its first request).
- auth now uses the `requests.AuthBase` classes to handle authentication instead of manually modifying headers.

Most importantly, the following have not changed their public signatures at all:
- everything under py42.sdk
- py42.utils
- py42.response
- py42.exceptions
- py42.usercontext

### Issues Resolved ###
- closes https://jira.corp.code42.com/browse/INTEG-983
- closes https://jira.corp.code42.com/browse/INTEG-1092

### Testing Procedure ###
To test .netrc:
- create a .netrc file with the proper configuration in your home directory
- call `sdk.from_local_account` passing `None` for username and password. You should still be able to do everything!

This is what I have tested in the refactor (thought that certainly doesn't mean that you shouldn't also test):
- stream by md5/sha256
- web restore
- FED security events
- at least one call to each of the clients / microservices
- cli unit tests all pass when this installed into it

### PR Checklist ###


src/py42/clients/_archive_access.py
```diff
@@ -17,9 +17,9 @@ class FileType(object):
 
 
 class ArchiveAccessorManager(object):
-    def __init__(self, archive_client, storage_client_factory):
-        self._archive_client = archive_client
-        self._storage_client_factory = storage_client_factory
+    def __init__(self, archive_service, storage_service_factory):
+        self._archive_service = archive_service
```
alanag13, line 21: a lot the changes in these files are just around changing module -> client, client -> service, etc.

src/py42/clients/_archive_access.py
```diff
@@ -278,11 +276,11 @@ def _wait_for_job(self, job_id):
             time.sleep(self._job_polling_interval)
 
     def _is_job_complete(self, job_id):
-        response = self._storage_archive_client.get_restore_status(job_id)
+        response = self._storage_archive_service.get_restore_status(job_id)
         is_done = response[u"done"]
         percentage_dict = {
             u"jobId": job_id,
-            u"status": response[u"status"],
+            u"status": response.data.get("status"),
```
alanag13, line 283: when I was testing this I saw that `status` was sometimes not in the response, particularly with smaller files, this prevents that from blowing up.

src/py42/clients/securitydata.py
```diff
@@ -278,13 +213,83 @@ def stream_file_by_md5(self, checksum):
...
+        return self._stream_file(checksum, info)
+
+    def _search_by_hash(self, hash, type):
+        query = FileEventQuery.all(type.eq(hash))
+        query.sort_key = "eventTimestamp"
+        query.sort_direction = "desc"
+        response = self.search_file_events(query)
+        return response
+
+    def _stream_file(self, checksum, version_info):
```
alanag13, line 228: This is the fix to the stream by hash logic that I mentioned. Before, we were writing a generator to loop through a list of versions and try each one. It turns out that the service accepts a list of versions and does that for you, so because of that we were able to get rid of a lot of extra log + retry logic.

src/py42/sdk/__init__.py
```diff
@@ -44,77 +76,74 @@ def from_local_account(cls, host_address, username, password):
...
-        )
-        basic_auth_session = session_factory.create_basic_auth_session(
-            host_address, username, password
-        )
-        sdk_dependencies = SDKDependencies(
-            host_address, session_factory, basic_auth_session
-        )
-        return cls(sdk_dependencies)
+        auth = None
+        if username and password:
```
alanag13, line 80: this is essentially what makes .netrc work, we just don't pass an `auth` object in if no username or password is supplied.

src/py42/sdk/__init__.py
```diff
@@ -134,25 +163,121 @@ def securitydata(self):
...
 
         Returns:
-            :class:`py42.modules.alertrules.AlertRulesModule`
+            :class:`py42.services.alertrules.AlertRulesClient`
         """
-        return self._sdk_dependencies.alerts_module
+        return self._clients.alerts
+
+
+def _init_services(main_connection, main_auth):
```
alanag13, line 190: all of this configuration lives in the same file as the SDK object itself now, I think that makes it much easier to visualize the hierarchy of how everything is composed.

src/py42/sdk/__init__.py
```diff
@@ -134,25 +163,121 @@ def securitydata(self):
...
+    alerts_key = u"AlertService-API_URL"
+    file_events_key = u"FORENSIC_SEARCH-API_URL"
+    preservation_data_key = u"PRESERVATION-DATA-SERVICE_API-URL"
+    employee_case_mgmt_key = u"employeecasemanagement-API_URL"
+    kv_prefix = u"simple-key-value-store"
+
+    kv_connection = Connection.from_microservice_prefix(main_connection, kv_prefix)
+    kv_service = KeyValueStoreClient(kv_connection)
+
+    alert_rules_conn = Connection.from_microservice_key(
```
alanag13, line 201: note that we can create the microservice connections right at startup, the new implementation make it so that we don't look up the url of the microservice until the first time we try to make a request with the connection

src/py42/services/_connection.py
```diff
@@ -0,0 +1,258 @@
...
+        params=None,
+        data=None,
+        json=None,
+        headers=None,
+        cookies=None,
+        files=None,
+        auth=None,
+        hooks=None,
+    ):
+        url = urljoin(self.host_address, url)
```
alanag13, line 200: this line ultimately triggers the host resolver to fire

src/py42/services/_connection.py
```diff
@@ -0,0 +1,258 @@
...
+    u"Connection": u"keep-alive",
+}
+
+
+class HostResolver(object):
+    def get_host_address(self):
+        raise NotImplementedError()
+
+
+class KnownUrlHostResolver(HostResolver):
```
alanag13, line 40: connections now take a `HostResolver` in their contstructor, these are used so that you can build the connection object before you know their url.

src/py42/services/alertrules.py
```diff
@@ -0,0 +1,136 @@
+from py42.services import BaseService
```
alanag13, line 1: im not sure why this whole file is green, much like the others it just has naming changes / accomodations for not needing microserviceclientfactory anymore

src/py42/services/_connection.py
```diff
@@ -0,0 +1,258 @@
...
+from py42.exceptions import Py42FeatureUnavailableError
+from py42.exceptions import raise_py42_error
+from py42.response import Py42Response
+from py42.services._auth import C42RenewableAuth
+from py42.settings import debug
+from py42.util import format_dict
+
+SESSION_ADAPTER = HTTPAdapter(pool_connections=200, pool_maxsize=4, pool_block=True)
+
+ROOT_SESSION = Session()
```
alanag13, line 24: This is the one session to rule them all, all requests got through this and simply have different domains / auth passed through it. This being a singleton combined with the session adapter above is essentially what solves the sockets problem.

src/py42/clients/securitydata.py
```diff
@@ -397,14 +402,32 @@ def _get_plans_in_node(destination, node):
...
         device_id = location[u"deviceUid"]
-        paths.append(u"{}{}".format(location[u"filePath"], file_name))
-        yield device_id, paths
+        device_entry = devices.get(device_id)
+        if device_entry:
+            device_entry["paths"].append(file_path)
+        else:
+            devices[device_id] = {"deviceGuid": device_id, "paths": [file_path]}
+
+    return [devices[key] for key in devices]
```
kiran-chaudhary, line 417: `return devices.values()`  or `devices.append({"deviceGuid".....}) return devices` where `devices = []`

`device_entry` is being defined but not being used, I guess 
 `devices[device_id] = {"deviceGuid": device_id, "paths": [file_path]}` should be changed to  `devices[device_id] = {"deviceGuid": device_id, "paths": device_entry["paths"]}`

src/py42/clients/securitydata.py
```diff
@@ -397,14 +402,32 @@ def _get_plans_in_node(destination, node):
...
+    for event in events:
+        device_guid = event[u"deviceUid"]
+        md5 = event[u"md5Checksum"]
+        sha256 = event[u"sha256Checksum"]
+        fileName = event[u"fileName"]
+        filePath = event[u"filePath"]
+
+        if device_guid and md5 and sha256 and fileName and filePath:
+            path = "{}{}".format(filePath, fileName)
+            return device_guid, md5, sha256, path
```
kiran-chaudhary, line 430: here for all `events`, are we interested only in one version look up?? if there are multiple `events` and if all properties are defined for the first `event` rest would be ignored!!! is that ok?

src/py42/services/alertrules.py
kiran-chaudhary: > im not sure why this whole file is green, much like the others it just has naming changes / accomodations for not needing microserviceclientfactory anymore

it has moved to a new directory `services` which din't exist earlier.

kiran-chaudhary (commented):
Some minor changes expected around file streaming refactor. 

Looking at the changes in `SDKClient` and `_init_*` methods, it seems we make connection to all `clients/services` at once whereas earlier it was on demand, i.e if we call `sdk.legalhold` that's when connection to `legalhold client/service` was initiated, now it seems we make connection to all clients while `SDKClient` is instantiated. If this understanding is correct, is this the intended change?

src/py42/clients/securitydata.py
alanag13: This is exactly what were doing before (we were just taking the fist event that was found). The only difference is that this makes sure that the event we found actually has all of these properties before attempting to use it (cloud events in particular will not have a device guid.

alanag13:
> Some minor changes expected around file streaming refactor.
> 
> Looking at the changes in `SDKClient` and `_init_*` methods, it seems we make connection to all `clients/services` at once whereas earlier it was on demand, i.e if we call `sdk.legalhold` that's when connection to `legalhold client/service` was initiated, now it seems we make connection to all clients while `SDKClient` is instantiated. If this understanding is correct, is this the intended change?

I explained above that the new `HostResolver` class takes care of this now. You no longer need to know a microservice's url in order to create the `Connection` object, it is instead determined the first time you try to make a request with the connection. No requests are during the SDKClient initialization with the exception of `user.get_current()` which is used to check that the credentials are valid.

src/py42/services/alertrules.py
alanag13: many other files were moved to that directory (everything that used to be in `clients`, see my renaming message in the PR description), and they still had the diff displayed appropriately.

src/py42/clients/securitydata.py
alanag13: This implementation is correct, but I should change line 413 to:
`devices.get(device_id)["paths"]` to be more clear. That line is modifying the `devices` dictionary, even though it kind of doesn't look like it is.

src/py42/clients/securitydata.py
alanag13: Using only the first event should be fine because pds will give us other device locations to use for the stream if the one on the event does not work.

src/py42/_compat.py
```diff
@@ -10,7 +10,7 @@
 is_py2 = _ver[0] == 2
 
 if is_py2:
-    from urllib import quote
+    from urllib import quote, urlencode
```
unparalleled-js, line 13: This is not our formatters' preferred way to import multiple items from a module

src/py42/clients/_archive_access.py
unparalleled-js: Is there basis for this convention? All three of the words `module`, `client`, and `service` can easily mean the same thing.

Glad to get rid of the use of `module` though, because of python modules

src/py42/clients/_archive_access.py
unparalleled-js: Yay we can use `get()` now. Thank you Alan.

unparalleled-js: Might be worth adding a test for this if you haven't already. I am fearful that this problem could potentially resurface.

src/py42/clients/securitydata.py
```diff
@@ -1,27 +1,30 @@
...
+        security_service,
+        file_event_service,
+        preservation_data_service,
+        saved_search_service,
+        storage_service_factory,
     ):
-        self._security_client = security_client
-        self._storage_client_factory = storage_client_factory
-        self._microservices_client_factory = microservices_client_factory
+        self._security_client = security_service
```
unparalleled-js, line 23: I believe this should be `self._security_service = security_service`

src/py42/clients/securitydata.py
```diff
@@ -278,13 +213,83 @@ def stream_file_by_md5(self, checksum):
...
+        if not len(events) or not info:
             raise Py42ChecksumNotFoundError(response, u"MD5", checksum)
-        sha256_hash = events[0][u"sha256Checksum"]
-        return self._stream_file(
-            self._find_file_versions(checksum, sha256_hash), checksum
+        return self._stream_file(checksum, info)
+
+    def _search_by_hash(self, hash, type):
+        query = FileEventQuery.all(type.eq(hash))
+        query.sort_key = "eventTimestamp"
```
unparalleled-js, line 223: Missing `u`'s here

src/py42/clients/securitydata.py
```diff
@@ -278,13 +213,83 @@ def stream_file_by_md5(self, checksum):
...
+        )
+        if not version:
+            version = self._get_other_file_location_version(md5_hash, sha256_hash)
+        return version
+
+    def _get_device_file_version(self, device_guid, md5_hash, sha256_hash, path):
+        response = self._preservation_data_service.get_file_version_list(
+            device_guid, md5_hash, sha256_hash, path
+        )
+        versions = response["versions"]
```
unparalleled-js, line 262: `u`

src/py42/clients/securitydata.py
unparalleled-js: I second changing line 413

src/py42/sdk/__init__.py
```diff
@@ -44,77 +76,74 @@ def from_local_account(cls, host_address, username, password):
...
-        )
-        sdk_dependencies = SDKDependencies(
-            host_address, session_factory, basic_auth_session
-        )
-        return cls(sdk_dependencies)
+        auth = None
+        if username and password:
+            auth = HTTPBasicAuth(username, password)
+        auth_connection = Connection.from_host_address(host_address, auth=auth)
+        auth = V3Auth(auth_connection)
```
unparalleled-js, line 83: Aren't you overriding line 81 here?

unparalleled-js: I see it gets uses, just sharing a variable is confusing

src/py42/sdk/__init__.py
```diff
@@ -21,12 +47,18 @@ def from_local_account(host_address, username, password):
     Returns:
         :class:`py42.sdk.SDKClient`
     """
-    return SDKClient.from_local_account(host_address, username, password)
+    client = SDKClient.from_local_account(host_address, username, password)
+
+    # test credentials
+    client.users.get_current()
```
unparalleled-js, line 53: Part of doesn't like that we are testing the connection here... I know we were before, but there have been a few circumstances where I wanted to test the connection outside of creating the SDK so I can handle the error differently.  Think of an integration's perspective where we are using py42 in a different application.

src/py42/services/_connection.py
```diff
@@ -0,0 +1,258 @@
...
+    def _get_host_address(self):
+        if not self._host_address:
+            with self._resolve_lock:
+                if not self._host_address:
+                    host = self._host_resolver.get_host_address()
+                    self._init_host_info(host)
+        return self._host_address
+
+    def _init_host_info(self, host):
+        if not host.startswith(u"http://") and not host.startswith(u"https://"):
```
unparalleled-js, line 235: ha, is non-encrypted http really allowed?

src/py42/services/_keyvaluestore.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class KeyValueStoreClient(BaseService):
+    def __init__(self, connection):
+        super(KeyValueStoreClient, self).__init__(connection)
+
+    def get_stored_value(self, key):
+        uri = u"/v1/{}".format(key)
+        # this request doesn't expect json, unlike most.
+        headers = {"Accept": "*/*"}
```
unparalleled-js, line 11: u

src/py42/services/alertrules.py
unparalleled-js: That happens sometimes, I think it might have to do with how you moved it.

src/py42/services/alerts.py
```diff
@@ -1,34 +1,34 @@
...
-        return self._session.post(uri, data=query)
+        return self._connection.post(uri, data=query)
 
     def get_details(self, alert_ids):
         if not isinstance(alert_ids, (list, tuple)):
             alert_ids = [alert_ids]
         tenant_id = self._user_context.get_current_tenant_id()
         uri = self._uri_prefix.format(u"query-details")
         data = {u"tenantId": tenant_id, u"alertIds": alert_ids}
-        results = self._session.post(uri, data=json.dumps(data))
```
unparalleled-js, line 31: What is going on here?  I don't see `json` as an explicit parameter for `Connection.post`

src/py42/_compat.py
alanag13: This file is ignored from formatting rules because it has a bunch of import that it will think are unused, but i can still change the way this import is done.

src/py42/clients/securitydata.py
alanag13: good catch for clarity, yes

src/py42/sdk/__init__.py
alanag13: def can change this to be less confusing

alanag13: worth discussing for sure, but as part of this refactor i think we should just aim to preserve existing behavior

src/py42/sdk/__init__.py
alanag13: worth discussing for sure, but as part of this refactor i think we should just aim to preserve existing behavior

src/py42/services/_connection.py
alanag13: always has been to support on prem, where whether or not you do that is totally up to them. it of course will not work at all when targeting cloud.

src/py42/services/alerts.py
alanag13: I can make it explicit. I changed everything that was using `data=json.dumps()` to `json=dict` for a couple reasons:
1. There was already inconsistency here.
2. Because earlier versions of python are not guaranteed to have consistent outputs of `json.dumps()` for the same dict, it makes it difficult to test that a `post` was issued with the correct data. checking for dict equivalency is as simple as `assert dict1 == dict2`.

src/py42/clients/_archive_access.py
alanag13: i kinda explained my reasoning in the description of the PR:
client = thing user interfaces with to make requests
service = web service, microservice, etc

unparalleled-js:
When I try to run the tests locally, I get this error: 

```
ImportError while loading conftest '/Users/juliya.smith/PycharmProjects/py42/tests/conftest.py'.
tests/conftest.py:11: in <module>
    from py42.sdk.queries.query_filter import QueryFilter
src/py42/sdk/__init__.py:5: in <module>
    from py42.clients.alertrules import AlertRulesClient
E   ImportError: bad magic number in 'py42.clients.alertrules': b'\x03\xf3\r\n'
```

src/py42/clients/__init__.py
```diff
@@ -1,6 +1,6 @@
-class BaseClient(object):
+from collections import namedtuple
 
-    __slots__ = ["_session"]
-
-    def __init__(self, session):
-        self._session = session
+Clients = namedtuple(
```
unparalleled-js, line 3: These are in the wrong order from how it gets constructed causing nothing to work for me

src/py42/sdk/__init__.py
```diff
@@ -134,25 +163,121 @@ def securitydata(self):
...
+        services.preservationdata,
+        services.savedsearch,
+        storage_service_factory,
+    )
+    alerts = AlertsClient(services.alerts, alertrules)
+    archive_accessor_mgr = ArchiveAccessorManager(
+        services.archive, storage_service_factory
+    )
+    archive = ArchiveClient(archive_accessor_mgr, services.archive)
+    clients = Clients(
```
unparalleled-js, line 276: This is a different order than defined in `clients/__init__.py`, causing nothin to work for me

src/py42/clients/__init__.py
unparalleled-js: Perhaps this the fault of one of the formatters?

src/py42/clients/__init__.py
```diff
@@ -1,6 +1,6 @@
...
+    u"Clients",
+    [u"alerts", u"archive", u"authority", u"detectionlists", u"securitydata"],
```
unparalleled-js, line 5: Pretty sure `AlertRules` is missing here

unparalleled-js (changes requested):
This is not working for me, curious as to why it is for you.

src/py42/clients/__init__.py
alanag13: the order of these shouldn't matter, are you observing otherwise?

alanag13:
> When I try to run the tests locally, I get this error:
> 
> ```
> ImportError while loading conftest '/Users/juliya.smith/PycharmProjects/py42/tests/conftest.py'.
> tests/conftest.py:11: in <module>
>     from py42.sdk.queries.query_filter import QueryFilter
> src/py42/sdk/__init__.py:5: in <module>
>     from py42.clients.alertrules import AlertRulesClient
> E   ImportError: bad magic number in 'py42.clients.alertrules': b'\x03\xf3\r\n'
> ```

Everything works on CI and on my machine. It looks like that error has something to do with `.pyc` files, which are byte code that gets generated when python files run. Those files won't go away when you switch branches, so I think that's whats going on here. If you haven't already, do a `git clean -xfd` and `pip install -e .[dev]` over again.

unparalleled-js:
Should update basics.html to include part about response.data

tests/sdk/test_sdk.py
```diff
@@ -22,95 +21,51 @@
...
 
-    def test_has_administation_client_set(
-        self, mock_session_factory, success_requests_session
-    ):
-        deps = SDKDependencies(
-            HOST_ADDRESS, mock_session_factory, success_requests_session
-        )
-        sdk = SDKClient(deps)
-        assert type(sdk.serveradmin) == administration.AdministrationClient
+    def test_has_administation_client_set(self, py42_connection, mock_auth):
```
unparalleled-js, line 33: administration spelling

tests/services/test_alert_rules.py
```diff
@@ -13,31 +11,31 @@
 
 
 @pytest.fixture
-def mock_response(self, mock_session, py42_response):
-    mock_session.post.return_value = py42_response
-    return mock_session
+def mock_response(self, mock_connection, py42_response):
```
unparalleled-js, line 14: Does this fixture need `self`?

src/py42/clients/_archive_access.py
```diff
@@ -184,8 +180,8 @@ def _get_default_file_size():
 class _RestorePoller(object):
     JOB_POLLING_INTERVAL_SECONDS = 1
 
-    def __init__(self, storage_archive_client, device_guid, job_polling_interval=None):
-        self._storage_archive_client = storage_archive_client
+    def __init__(self, storage_archive_service, device_guid, job_polling_interval=None):
+        self._storage_archive_service = storage_archive_service
         self._device_guid = device_guid
         self._job_polling_interval = (
             job_polling_interval or self.JOB_POLLING_INTERVAL_SECONDS
```
unparalleled-js, line 187: Line 192 is an example where we can use `response.data` now

unparalleled-js, line 187: Line 192 is an example where we can use response.data now

src/py42/clients/securitydata.py
```diff
@@ -322,9 +327,9 @@ def _try_get_security_detection_event_client(self, plan_storage_info):
...
         # otherwise, create it
         if client is None:
-            client = self._storage_client_factory.from_plan_info(
+            client = self._storage_service_factory.create_security_data_service(
                 plan_storage_info.plan_uid, plan_storage_info.destination_guid
-            ).securitydata
+            )
 
             # store this client via its guid so that we don't have to call StorageAuthToken
             # just to determine what storage client to use
```
unparalleled-js, line 335: Line 373 needs updating .... can just use `data.get()` and the comment above can be remove or altered

src/py42/clients/securitydata.py
```diff
@@ -397,14 +402,32 @@ def _get_plans_in_node(destination, node):
...
+    devices = {}
+    for location in locations:
         file_name = location[u"fileName"]
+        file_path = u"{}{}".format(location[u"filePath"], file_name)
         device_id = location[u"deviceUid"]
-        paths.append(u"{}{}".format(location[u"filePath"], file_name))
-        yield device_id, paths
+        device_entry = devices.get(device_id)
+        if device_entry:
+            device_entry["paths"].append(file_path)
```
unparalleled-js, line 413: `u`s

src/py42/services/fileevent.py
```diff
@@ -23,7 +23,7 @@ def search(self, query):
         """
         query = str(query)
         uri = u"/forensic-search/queryservice/api/v1/fileevent"
-        return self._session.post(uri, data=query)
+        return self._connection.post(uri, data=query)
 
     def get_file_location_detail_by_sha256(self, hash):
```
unparalleled-js, line 28: Overriding a built-in here

src/py42/services/_connection.py
```diff
@@ -0,0 +1,258 @@
...
+    def delete(self, url, **kwargs):
+        return self.request(u"DELETE", url, **kwargs)
+
+    def request(
+        self,
+        method,
+        url,
+        params=None,
+        data=None,
+        json=None,
```
unparalleled-js, line 138: Python dictionaries are technically not JSON.  JSON is a string.

alanag13: This mirrors the `requests` implementation of `json`.

src/py42/services/fileevent.py
alanag13: assuming you mean `hash`. This wasn't changed as part of this PR and is technically a breaking change (in the case that a keyword argument is used), but I'm still willing to change it to `checksum` or something

src/py42/services/_connection.py
unparalleled-js: Or is this just a requests thing?

src/py42/services/storage/_auth.py
```diff
@@ -0,0 +1,64 @@
...
+        return self._server_url
+
+    def get_tmp_auth(self):
+        raise NotImplementedError()
+
+    def _get_credentials(self):
+        login_info = self.get_tmp_auth()
+        login_token = login_info[u"loginToken"]
+        self._server_url = login_info[u"serverUrl"]
+        return u"{} {}".format(u"login_token", login_token)
```
unparalleled-js, line 20: No need to include hardcoded strings in a format statement, you can just put it in the original string

src/py42/sdk/__init__.py
```diff
@@ -134,25 +163,121 @@ def securitydata(self):
...
+    pds_conn = Connection.from_microservice_key(
+        kv_service, preservation_data_key, auth=main_auth
+    )
+    ecm_conn = Connection.from_microservice_key(
+        kv_service, employee_case_mgmt_key, auth=main_auth
+    )
+
+    user_svc = UserService(main_connection)
+    administration_svc = AdministrationService(main_connection)
+    file_events_service = FileEventService(file_events_conn)
```
unparalleled-js, line 219: Some minor discrepancy between`file_events..` and `FileEvent..`

unparalleled-js (approved):

unparalleled-js:
Screen Shot 2020-08-24 at 17 52 34
I wish I could figure out why this is happening. It has no effect and goes away on `git clean -xfd` but always comes back

alanag13:
> Screen Shot 2020-08-24 at 17 52 34
> 
> I wish I could figure out why this is happening. It has no effect and goes away on `git clean -xfd` but always comes back, even when not changing branches

This might have something to do with using named tuples instead of a class (I think maybe the intellisense can't "see" the properties with named tuples).

CHANGELOG.md
```diff
@@ -17,9 +17,15 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.alerts.resolve()`
     - `sdk.alerts.reopen()`
 
+## Fixed
+
+- Issue that in rare circumstance could cause py42 to exhaust network sockets. This could sometimes occur if you were running a multi-threaded program that communicated with many Code42 storage nodes.
+
 ### Added
 
-- Debug logs for restore progress during the method call `py42.archive.stream_from_back()`.
```
unparalleled-js, line 22: I feel like this might be a good change to make users aware of, if they want to track progress. Else, how are they supposed to know?

alanag13: it isn't removed, the method name was misspelled (should be `backup()`), so I corrected that.

docs/methoddocs/archive.md
```diff
@@ -1,7 +1,7 @@
 # Archive
 
 ```eval_rst
-.. autoclass:: py42.modules.archive.ArchiveModule
```
unparalleled-js, line 4: What about storage archive servife?

unparalleled-js:
Am I missing it, or is `preservationservice` missing from the docs?

src/py42/clients/_archive_access.py
unparalleled-js: @alanag13 Don't want to do this?

CHANGELOG.md
unparalleled-js: Oh I see it is below

src/py42/clients/securitydata.py
```diff
@@ -278,13 +213,83 @@ def stream_file_by_md5(self, checksum):
...
+        info = _get_version_lookup_info(events)
+        if not len(events) or not info:
             raise Py42ChecksumNotFoundError(response, u"MD5", checksum)
-        sha256_hash = events[0][u"sha256Checksum"]
-        return self._stream_file(
-            self._find_file_versions(checksum, sha256_hash), checksum
+        return self._stream_file(checksum, info)
+
+    def _search_by_hash(self, hash, type):
+        query = FileEventQuery.all(type.eq(hash))
```
unparalleled-js, line 222: `type` is a commonly used python function, threw me off a little seeing it here

src/py42/clients/securitydata.py
```diff
@@ -278,13 +213,83 @@ def stream_file_by_md5(self, checksum):
...
+        query.sort_key = u"eventTimestamp"
+        query.sort_direction = u"desc"
```
unparalleled-js, line 224: We have an enum for this called `SortDirection`

src/py42/clients/securitydata.py
alanag13: oh yeah, i meant to change `hash` too, thanks.

docs/methoddocs/archive.md
alanag13: it is not currently directly exposed. The plan is to eventually make it easier to create the individual services "standalone" so that people could call those without creating the whole sdk, but right now nothing has changed in terms of what users are expected to call directly.

alanag13:
> Am I missing it, or is `preservationservice` missing from the docs?

it is not currently directly exposed. The plan is to eventually make it easier to create the individual services "standalone" so that people could call those without creating the whole sdk, but right now nothing has changed in terms of what users are expected to call directly.

src/py42/services/alertrules.py
```diff
@@ -0,0 +1,136 @@
...
+            tenant_id = self._user_context.get_current_tenant_id()
+            self._file_type_mismatch = FileTypeMismatchService(
+                self._connection, tenant_id
+            )
+        return self._file_type_mismatch
+
+    def add_user(self, rule_id, user_id):
+        tenant_id = self._user_context.get_current_tenant_id()
+        user_details = self._user_profile_service.get_by_id(user_id)
+        user_aliases = user_details["cloudUsernames"] or []
```
unparalleled-js, line 45: Hmm... Should this be `user_details.dataget(cloudUsernames")`?

alanag13: it hasn't been a problem so far and I would say most users do not have cloud usernames -- but I do agree that it feels slightly safer.

src/py42/services/alertrules.py
```diff
@@ -0,0 +1,136 @@
...
+    def get(self, rule_id):
+        """Fetch File type mismatch alert rules by rule id.
+
+        Args:
+            rule_id (str): Observer rule Id of a rule to be fetched.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        data = {"tenantId": self._tenant_id, "ruleIds": [rule_id]}
```
unparalleled-js, line 135: Is this an example of where the strings may be concat'ed and is dangerous not to have `u`?

src/py42/services/orgs.py
```diff
@@ -95,7 +93,7 @@ def get_all(self, **kwargs):
     def block(self, org_id):
         """Blocks the organization with the given org ID as well as its child organizations. A
         blocked organization will not allow any of its users or devices to log in. New
-        registrations will be rejected and all currently logged in clients will be logged out.
+        registrations will be rejected and all currently logged in services will be logged out.
```
unparalleled-js, line 96: lol I think this one should _actually_ be `clients`.

src/py42/services/alertrules.py
alanag13: for dict keys it would probably not be a concern unless we expected a use case where those keys would be concatenated with a unicode string. A better example would be some thing like an exception message that contained a file name -- there's a decent change the file name could have unicode characters, which would cause building the error message to fail when the file name is appended to the (non-unicode) rest of the message.

src/py42/services/orgs.py
alanag13: lol I tried to be careful about this, good catch

src/py42/services/storage/_service_factory.py
```diff
@@ -0,0 +1,76 @@
...
+        try:
+            url = tmp_auth.get_storage_url()
+            connection = self.get_saved_connection_for_url(url)
+            if connection is None:
+                with self._list_update_lock:
+                    connection = self.get_saved_connection_for_url(url)
+                    if connection is None:
+                        connection = Connection.from_host_address(url, auth=tmp_auth)
+                        self._session_cache[url.lower()] = connection
+        except Py42HTTPError as ex:
```
unparalleled-js, line 71: Maybe this should be `Py42Error`.  Who is to say that we don't add a non-http related error to one of these calls down the line that causes this to fail?  IDK, just a thought.

tests/services/storage/test_archive.py
```diff
@@ -49,26 +49,26 @@
...
 
 
 @pytest.fixture
-def storage_archive_client(mocker):
-    return mocker.MagicMock(spec=StorageArchiveClient)
+def storage_archive_service(mocker):
+    return mocker.MagicMock(spec=StorageArchiveService)
 
 
 class TestStorageArchiveClient(object):
```
unparalleled-js, line 65: This should be `TestStorageArciveService`

tests/services/detectionlists/test_user_profile.py
```diff
@@ -58,118 +56,124 @@ def test_create_posts_expected_data(
...
             and posted_data["username"] == "942897397520289999"
         )
 
     def test_get_by_id_posts_expected_data(
-        self, mock_session, user_context, mock_user_client
+        self, mock_connection, user_context, mock_user_client
     ):
-        detection_list_user_client = DetectionListUserClient(
-            mock_session, user_context, mock_user_client
+        detection_list_user_client = DetectionListUserService(
```
unparalleled-js, line 77: service

tests/services/detectionlists/test_user_profile.py
```diff
@@ -1,54 +1,52 @@
...
         response.status_code = 400
         exception = mocker.MagicMock(spec=HTTPError)
         exception.response = response
-        mock_session.post.side_effect = Py42BadRequestError(exception)
-        return mock_session
+        mock_connection.post.side_effect = Py42BadRequestError(exception)
+        return mock_connection
 
     @pytest.fixture
     def mock_user_client_raises_exception(
```
unparalleled-js, line 28: mock_user_service_

tests/services/detectionlists/test_user_profile.py
```diff
@@ -1,54 +1,52 @@
...
 from requests.exceptions import HTTPError
 
-from py42._internal.clients.detection_list_user import DetectionListUserClient
-from py42.clients.users import UserClient
 from py42.exceptions import Py42BadRequestError
+from py42.services.detectionlists.user_profile import DetectionListUserService
+from py42.services.users import UserService
 
 
 class TestDetectionListUserClient(object):
```
unparalleled-js, line 10: TestDetectionListUserService

src/py42/services/storage/_service_factory.py
alanag13: Nothing in this block raises`Py42Error`, but this might be a rare case where `catch Exception` is okay since we are rethrowing with the cause.

tests/services/detectionlists/test_high_risk_employee.py
```diff
@@ -29,40 +27,43 @@ def test_choices_are_correct(self):
 
 class TestHighRiskEmployeeClient(object):
```
unparalleled-js, line 28: Check for Client is the wrong spots in these tests... This one should be `TestHighRiskEmployeeService`

alanag13: thanks

tests/test_usercontext.py
```diff
@@ -24,7 +24,7 @@
 class TestUserContext(object):
     @pytest.fixture
     def successful_administration_client(self, mocker, py42_response):
```
unparalleled-js, line 26: successful_administration_service

tests/services/test_orgs.py
```diff
@@ -33,51 +33,51 @@ def mock_get_all_empty_response(self, mocker):
         return Py42Response(response)
 
     def test_get_org_by_id_calls_get_with_uri_and_params(
```
unparalleled-js, line 35: This test seems like it is not the right directory as well... This is testing a Service but it is in `tests/clients/`

unparalleled-js:
Sorry I missed a lot the first time I reviewed.

unparalleled-js (changes requested):
At minimum, the test updates should be addressed

src/py42/clients/_archive_access.py
alanag13: This is actually assigning a jobId, not getting one, so we don't need to use `data.get` to check to see if anything exists

unparalleled-js: I don't mean data.get, I mean `size_dict = size_response.data`

unparalleled-js: I don't see a reason why we would ever need to call `json.loads()` on a py42response's text anymore

unparalleled-js:
https://github.com/code42/py42/pull/174/files#r477412053

src/py42/clients/securitydata.py
unparalleled-js: @alanag13 This is also another example where we don't need `json.loads()` still

src/py42/clients/securitydata.py
```diff
@@ -370,7 +370,7 @@ def _get_security_detection_events(
                 if response.text:
                     # we use json.loads here because the cursor prop doesn't appear
```
unparalleled-js, line 370: CAn you remove this comment?

unparalleled-js: Or change it?

tests/clients/test_alertrules.py
```diff
@@ -0,0 +1,138 @@
...
+def mock_alert_rules_service(mocker):
+    return mocker.MagicMock(spec=AlertRulesService)
+
+
+@pytest.fixture
+def mock_alerts_service(mocker):
+    return mocker.MagicMock(spec=AlertService)
+
+
+class TestAlertRulesService(object):
```
unparalleled-js, line 45: This should be `TestAlertRulesClient`

tests/sdk/test_sdk.py
```diff
@@ -22,95 +21,51 @@
...
+        assert type(service.legalhold) == legalhold.LegalHoldService
 
-    def test_has_org_client_set(self, mock_session_factory, success_requests_session):
-        deps = SDKDependencies(
-            HOST_ADDRESS, mock_session_factory, success_requests_session
-        )
-        sdk = SDKClient(deps)
-        assert type(sdk.orgs) == orgs.OrgClient
+    def test_has_org_service_set(self, py42_connection, mock_auth):
+        service = SDKClient(py42_connection, mock_auth)
```
unparalleled-js, line 58: local variable should be probably be client?

tests/services/storage/test_preservationdata.py
```diff
@@ -1,7 +1,7 @@
 import pytest
 import requests
 
-from py42._internal.clients.storage.storagenode import StoragePreservationDataClient
+from py42.services.storage.preservationdata import StoragePreservationDataService
 
 
 class TestStoragePreservationDataClient(object):
```
unparalleled-js, line 7: Should be Service

tests/services/storage/test_preservationdata.py
```diff
@@ -12,14 +12,14 @@ def mock_request(self, mocker):
         return request
 
     def test_get_download_token_calls_get_with_valid_params(
-        self, mock_successful_session
+        self, mock_successful_connection
     ):
-        client = StoragePreservationDataClient(
-            mock_successful_session, mock_successful_session
+        client = StoragePreservationDataService(
```
unparalleled-js, line 17: local var should be service

tests/services/storage/test_securitydata.py
```diff
@@ -77,12 +77,12 @@ def security_detection_event_summary_params(security_detection_events_params):
 class TestStorageSecurityClient(object):
```
unparalleled-js, line 77: Shold be service

tests/services/test_alert_rules.py
```diff
@@ -13,31 +13,32 @@
...
-def mock_response(self, mock_session, py42_response):
-    mock_session.post.return_value = py42_response
-    return mock_session
+def mock_response(mock_connection, py42_response):
+    mock_connection.post.return_value = py42_response
+    return mock_connection
 
 
 @pytest.fixture
 def mock_detection_list_user_client(mocker, py42_response):
```
unparalleled-js, line 22: All this fixture should be referencing Service

tests/services/test_alert_rules.py
```diff
@@ -13,31 +13,32 @@
...
     py42_response.text = MOCK_DETECTION_LIST_GET_RESPONSE
-    detection_list_user_client = mocker.MagicMock(spec=DetectionListUserClient)
+    py42_response.data = json.loads(MOCK_DETECTION_LIST_GET_RESPONSE)
+    detection_list_user_client = mocker.MagicMock(spec=DetectionListUserService)
     detection_list_user_client.get_by_id.return_value = py42_response
     return detection_list_user_client
 
 
 class TestAlertRulesClient(object):
```
unparalleled-js, line 30: Service

tests/services/test_alerts.py
```diff
@@ -0,0 +1,369 @@
...
+
+@pytest.fixture
+def mock_get_all_session(mocker, py42_response):
+    py42_response.text = TEST_RESPONSE
+    connection = mocker.MagicMock(spec=Connection)
+    connection.post.return_value = py42_response
+    return connection
+
+
+class TestAlertClient(object):
```
unparalleled-js, line 68: Service

unparalleled-js:
A few more missed renames.

tests/services/test_alert_rules.py
```diff
@@ -0,0 +1,83 @@
...
+    detection_list_user_service = mocker.MagicMock(spec=DetectionListUserService)
+    detection_list_user_service.get_by_id.return_value = py42_response
+    return detection_list_user_service
+
+
+class TestAlertRulesService(object):
+    def test_add_user_posts_expected_data(
+        self, mock_connection, user_context, mock_detection_list_user_service
+    ):
+        alert_rule_client = AlertRulesService(
```
unparalleled-js, line 34: .... technically should be `alert_rule_service` here, but I will still approve

unparalleled-js (approved):

tests/services/test_alert_rules.py
unparalleled-js: HEre and other tests in this mod

PR #180, opened 2020-08-26
Chore/exception compatibility

alanag13:
### Description of Change ###

- removes `Py42UserNotFoundError`. This would be a breaking change for situations where a username did not exist (normally an empty list is returned).
- makes `Py42UserAlreadyAddedError` an instance of `Py42BadRequestError` in order to make exception handling backwards compatible.
- makes `Py42LegalHoldNotFoundOrPermissionDeniedError` an instance of `Py42ForbiddenError` in order to make exception handling backwards compatible.

### DIFF
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -87,7 +87,27 @@ class Py42SessionInitializationError(Py42Error):
         super(Py42SessionInitializationError, self).__init__(exception, error_message)
 
 
-class Py42UserAlreadyAddedError(Py42HTTPError):
+class Py42BadRequestError(Py42HTTPError):
+    """A wrapper to represent an HTTP 400 error."""
+
+
+class Py42UnauthorizedError(Py42HTTPError):
+    """A wrapper to represent an HTTP 401 error."""
+
+
+class Py42ForbiddenError(Py42HTTPError):
+    """A wrapper to represent an HTTP 403 error."""
+
+
+class Py42NotFoundError(Py42HTTPError):
+    """A wrapper to represent an HTTP 404 error."""
+
+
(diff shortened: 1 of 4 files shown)


src/py42/exceptions.py
```diff
@@ -96,7 +87,27 @@ def __init__(self, exception):
         super(Py42SessionInitializationError, self).__init__(exception, error_message)
 
 
-class Py42UserAlreadyAddedError(Py42HTTPError):
+class Py42BadRequestError(Py42HTTPError):
```
alanag13, line 90: These had to be moved up because super classes have to come before their inheritors if they are in the same file

kiran-chaudhary (changes requested):
The changes seem fine. Need change log for breaking change.

unparalleled-js (approved):

unparalleled-js:
> The changes seem fine. Need change log for breaking change.

Oh yeah.  This isn't released, but the CHANGELOG needs to be amended to get rid of the User exception one.

unparalleled-js:
We could potentially leave User Does Not Exists Exception and just import it other places... There is a strong chance we will need this exception in more places than just the CLI.

unparalleled-js (commented):
Fix Changelog.
Consider allowing `Py42UserDoesNotExists` be importable from py42 yet, even if it not raised anywhere.  If that is the case, make sure it's doc str reflects that.

alanag13:
It doesn't make much sense to me to have Py42 define exceptions that it does not raise - I have never imported an exception from a library to raise it, just to catch it.

Thanks for the reminder to edit the change log, I'll do that this morning

PR #186, opened 2020-09-07
Chore alert state

kiran-chaudhary:
### Description of Change ###

Changed `resolve-alert` and `reopen-alert` endpoints to `update-state` endpoint.
Added methods to support `in progress` and `pending` state.

### Issues Resolved ###

fixes #INTEG-1152

- closes # [INTEG-1152](https://jira.corp.code42.com/browse/INTEG-1152)

### Testing Procedure ###
```
sdk.alerts.reopen('b9ac929f-1bc4-4f6f-8103-4674e65edade', "test note ")
sdk.alerts.resolve('b9ac929f-1bc4-4f6f-8103-4674e65edade', "test note 2")
sdk.alerts.in_progress('b9ac929f-1bc4-4f6f-8103-4674e65edade', "test note 4")
sdk.alerts.pending(['b9ac929f-1bc4-4f6f-8103-4674e65edade'], "test note 3")
sdk.alerts.reopen('b9ac929f-1bc4-4f6f-8103-4674e65edade')
```

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Removed faulty `within_the_last()` method from `sdk.queries.alerts.filters.alert_filter.DateObserved`.
 
+### Changed
+- Updated Alert `resolve-alert` and `reopen-alert` endpoints to `update-state` endpoint.
+
+### Added
+
+- `sdk.alerts` methods to support new alert states `pending` and `in progress`:
+    `sdk.alerts.pending()`
+    `sdk.alerts.in_progress()`
+
 ## 1.8.1 - 2020-08-28
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
-- `sdk.alerts` methods to support new alert states `pending` and `in progress`:
-    `sdk.alerts.pending()`
-    `sdk.alerts.in_progress()`
+- `sdk.alerts.update_state()` method to update state, `sdk.alerts.resolve` and `sdk.alerts.reopen` are deprecated and will be removed from version 2.0
 
 ## 1.8.1 - 2020-08-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,6 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Removed faulty `within_the_last()` method from `sdk.queries.alerts.filters.alert_filter.DateObserved`.
 
-### Changed
-- Updated Alert `resolve-alert` and `reopen-alert` endpoints to `update-state` endpoint.
-
 ### Added
 
 - `sdk.alerts.update_state()` method to update state, `sdk.alerts.resolve` and `sdk.alerts.reopen` are deprecated and will be removed from version 2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
-- `sdk.alerts.update_state()` method to update state, `sdk.alerts.resolve` and `sdk.alerts.reopen` are deprecated and will be removed from version 2.0
+- `sdk.alerts.update_state()` method to update state.
 
 ## 1.8.1 - 2020-08-28
 
--- a/docs/methoddocs/alerts.md
+++ b/docs/methoddocs/alerts.md
@@ -1,17 +1,3 @@
-# Alerts
-
-```eval_rst
-.. autoclass:: py42.clients.alerts.AlertsClient
-    :members:
-    :show-inheritance:
-```
-
-```eval_rst
-.. autoclass:: py42.sdk.queries.alerts.alert_query.AlertQuery
-    :members:
-    :show-inheritance:
-```
-
 ## Filter Classes
 
 The following classes construct filters for file event queries. Each filter class corresponds to an alert detail.
@@ -25,5 +11,20 @@ See [Executing Searches](../userguides/searches.md) for more on building search 
 .. automodule:: py42.sdk.queries.alerts.filters.alert_filter
     :members:
     :inherited-members:
+    :show-inheritance:
+```
+
+```eval_rst
+.. autoclass:: py42.sdk.queries.alerts.alert_query.AlertQuery
+    :members:
+    :show-inheritance:
+```
+
+
+# Alerts
+
+```eval_rst
+.. autoclass:: py42.clients.alerts.AlertsClient
+    :members:
     :show-inheritance:
 ```
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -1,3 +1,11 @@
+class AlertState(object):
+
+    IN_PROGRESS = "IN_PROGRESS"
+    OPEN = "OPEN"
+    PENDING = "PENDING"
+    RESOLVED = "RESOLVED"
+
+
 class AlertsClient(object):
     def __init__(self, alert_service, alert_rules_client):
         self._alert_service = alert_service
@@ -49,7 +57,9 @@ class AlertsClient(object):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        return self._alert_service.resolve(alert_ids, reason=reason)
+        return self._alert_service.update_state(
+            AlertState.RESOLVED, alert_ids, reason=reason
+        )
 
     def reopen(self, alert_ids, reason=None):
         """Reopens the resolved alerts with the given IDs.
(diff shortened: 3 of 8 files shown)


src/py42/clients/alerts.py
```diff
@@ -49,7 +57,9 @@ def resolve(self, alert_ids, reason=None):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        return self._alert_service.resolve(alert_ids, reason=reason)
+        return self._alert_service.update_state(
```
unparalleled-js, line 60: Wow thanks for doing this

src/py42/clients/alerts.py
```diff
@@ -1,3 +1,11 @@
+class AlertState(object):
```
unparalleled-js, line 1: We already have an `AlertState` enum in `py42.sdk.queries.alerts.filters.alert_filter`.... I believe this are the same thing too, so we should leverage this. Correct me if I am  wrong

src/py42/clients/alerts.py
```diff
@@ -61,4 +71,34 @@ def reopen(self, alert_ids, reason=None):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        return self._alert_service.reopen(alert_ids, reason=reason)
+        return self._alert_service.update_state(
+            AlertState.OPEN, alert_ids, reason=reason
+        )
+
+    def pending(self, alert_ids, reason=None):
```
unparalleled-js, line 78: I think instead of adding a new method for each of the other states, we should just have a catch-all like `up=date_state()`.  `pending()` is not a function-y sounding name and If other states come along, this way we already can support them.  I feel like we need other devs opinions on this design decision though:  @alanag13 , @timabrmsn , @kiran-chaudhary

src/py42/clients/alerts.py
kiran-chaudhary: Yea they are one and the same, good that you brought it up, Filters only supports `resolved` and `open` choices, need to add the new ones there.
I don't see any harm in using them, only that they were specifically suited for filters.

src/py42/clients/alerts.py
kiran-chaudhary: Yea, we can redesign it.  I did think about it but then thought that would be considered a breaking change, also thought  of `to_pending` but later on went ahead with `pending`.

unparalleled-js: It is not a breaking change as long as we leave the original `resolve` and `reopen` methods

src/py42/clients/alerts.py
unparalleled-js: Yeah I think that is ok in this case.  We are not using it as a filter here but it is still the same constants.

kiran-chaudhary: FYI: Adding import statement for filters raises warning while generating docs. Investigating the error it seems need to add some path statements to `conf.py` in `docs`. Working on it.

```
WARNING: autodoc: failed to import class 'alerts.AlertsClient' from module 'py42.clients'; the following exception was raised:
cannot import name 'AlertsClient'
```

src/py42/clients/alerts.py
kiran-chaudhary: Changed.

src/py42/clients/alerts.py
kiran-chaudhary: Changed.

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Removed faulty `within_the_last()` method from `sdk.queries.alerts.filters.alert_filter.DateObserved`.
 
+### Changed
+- Updated Alert `resolve-alert` and `reopen-alert` endpoints to `update-state` endpoint.
```
unparalleled-js, line 18: I don't think we need to add this to the Changelog since it happens behind the scenes

docs/methoddocs/alerts.md
```diff
@@ -1,8 +1,16 @@
-# Alerts
```
unparalleled-js, line 1: Other reviewers: Note that this file was re-ordered to handle a docs build issue that arose from importing the `AlertState` constants

CHANGELOG.md
```diff
@@ -14,6 +14,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+
+### Added
+
+- `sdk.alerts.update_state()` method to update state, `sdk.alerts.resolve` and `sdk.alerts.reopen` are deprecated and will be removed from version 2.0
```
unparalleled-js, line 22: I'd like to know others thoughts on removing `resolve()` and `reopen()`... Maybe it is too soon to mark these as deprecated

timabrmsn: I think I'm ok with removing them in 2.0 where breaking changes are expected. Is your concern that it's not enough time for deprecating? Or that you prefer having the explicit methods vs the more generic one?

unparalleled-js: My only hesitation is that `resolve()` and `reopen()` seem (keyword: _seem_) like they would be expected methods to have on an Alert... What can you do with alerts?  You can resolve them.  You can reopen them.... The other states do not so easily come to mind.

unparalleled-js: That being said, I am still ok with deprecating them just because `change_state(AlertState.RESOLVED)` seems perfectly readable to me...

src/py42/clients/alerts.py
```diff
@@ -42,23 +45,44 @@ def get_details(self, alert_ids):
...
         """
-        return self._alert_service.resolve(alert_ids, reason=reason)
+        return self._alert_service.update_state(
+            AlertState.DISMISSED, alert_ids, reason=reason
+        )
 
     def reopen(self, alert_ids, reason=None):
         """Reopens the resolved alerts with the given IDs.
 
+        Deprecated. Support will be removed from version 2.0.0. Use `update_state` instead.
```
unparalleled-js, line 64: Can you use a doc str link for `update_state` ?

`:meth:\`~py42.clients.alerts.AlertClient.update_state()\``

unparalleled-js: Replace &#96 with backtick

src/py42/services/alerts.py
```diff
@@ -31,21 +31,18 @@ def get_details(self, alert_ids):
         results = self._connection.post(uri, json=data)
         return _convert_observation_json_strings_to_objects(results)
 
-    def resolve(self, alert_ids, reason=None):
+    def update_state(self, state, alert_ids, reason=None):
```
unparalleled-js, line 34: Since this is a new method, can we have the parameter name be `note` instead of `reason`? Now that there are more states, `note` makes more sense.

src/py42/clients/alerts.py
```diff
@@ -42,23 +45,44 @@ def get_details(self, alert_ids):
...
+            reason (str, optional): The reason the alerts are reopened. Defaults to None.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        return self._alert_service.update_state(
+            AlertState.OPEN, alert_ids, reason=reason
+        )
+
+    def update_state(self, status, alert_ids, reason=None):
```
unparalleled-js, line 77: See https://github.com/code42/py42/pull/186/files#r485637405 ....

This does not mean you should change the parameter name in `resolve()` or `reopen()`... Leaving those as `reason` is perfectly fine, especially if it's deprecated.

src/py42/services/alerts.py
unparalleled-js: It is ok for `resolve` and `reopen` to use `reason` but the `update_state` methods should use `note`.

src/py42/clients/alerts.py
kiran-chaudhary: Its already part of AlertClient doc, do we need a separate one?

image

src/py42/services/alerts.py
unparalleled-js: And pass `reason` as note in `resolve` and `reopen`

src/py42/clients/alerts.py
kiran-chaudhary: ohh, ignore, I see what you mean.

tests/services/test_alerts.py
```diff
@@ -182,88 +182,59 @@ def test_get_details_when_observation_data_not_parseable_remains_unchanged(
...
         assert observation_data == expected_observation_data
 
-    def test_resolve_when_not_given_tenant_id_posts_expected_data(
+    def test_update_state_when_not_given_tenant_id_posts_expected_data(
         self, mock_connection, user_context, successful_post
     ):
         alert_service = AlertService(mock_connection, user_context)
         alert_ids = ["ALERT_ID_1", "ALERT_ID_2"]
-        alert_service.resolve(alert_ids)
+        alert_service.update_state("RESOLVED", alert_ids, "")
```
unparalleled-js, line 190: It would be a slightly better test if you use the non-empty string for the `note` parameter.

tests/services/test_alerts.py
```diff
@@ -182,88 +182,59 @@ def test_get_details_when_observation_data_not_parseable_remains_unchanged(
         expected_observation_data = '{"invalid_json": ][ }'
         assert observation_data == expected_observation_data
 
-    def test_resolve_when_not_given_tenant_id_posts_expected_data(
```
unparalleled-js, line 185: There should be a test that makes sure when you pass `None` for the`note` parameter, that it actually uses the empty string

src/py42/clients/alerts.py
kiran-chaudhary: Updated.

src/py42/services/alerts.py
kiran-chaudhary: renamed.

tests/services/test_alerts.py
kiran-chaudhary: Done.

tests/services/test_alerts.py
kiran-chaudhary: Added new test.

CHANGELOG.md
alanag13: Definitely don't remove these now. If we arent sure about whether or not these should be deprecated, they probably shouldn't be. I know that there are multiple statuses for alerts now, but opening/closing will still probably be the most common actions against them, and a convenience method for this seems like a good idea, at least to me. We don't need to remove methods just because the API does.

src/py42/clients/alerts.py
```diff
@@ -42,23 +45,42 @@ def get_details(self, alert_ids):
     def resolve(self, alert_ids, reason=None):
         """Resolves the alerts with the given IDs.
 
+        Deprecated. Support will be removed from version 2.0.0. Use :meth:`~py42.clients.alerts.AlertClient.update_state()` instead.
```
alanag13, line 48: I think whether or not we deprecate any of these methods is a decision for another PR that we should complete before we start work on 2.0

CHANGELOG.md
unparalleled-js: @alanag13 We are not talking about removing them, we are talking about planning to remove them in 2.0, but I agree that this decision can more easily made post this PR so @kiran-chaudhary You might as well remove the deprecation notes.

unparalleled-js (approved):

PR #187, opened 2020-09-07
Fix SavedSearchClient.execute() paging

### LINKED ISSUE #179: [Bug] pg_size and pg_num have no effect in the saved search execute method
### Description

The `execute()` method for saved searches is not doing anything with `pg_num` or `pg_size`.

### Steps to Reproduce

1. Try using `pg_num` or `pg_size` on a saved search `execute()` call.
2. In debug mode, notice that the default params are always in the request
3. Also, look at the code and see that they are not used.

### Expected Behavior
To be able to use the `pg_size` and `pg_num` parameters when executing a saved search

### Actual Behavior
The default values for the query are always used.

DiscoRiver:
### Description of Change ###

This code fixes #179, and changes the following;

- `py42.sdk.queries.BaseQuery.from_dict` now accepts keyword args, which are included in the returning class object. In this case they are used in `FileEventQuery`
- `SavedSearchClient.get_query` now accepts `pg_num` and `pg_size` keyword arguments.
- `SavedSearchClient.execute` now works with user-defined `pg_num` and `pg_size` arguments

I _believe_ the test `test_execute_calls_post_with_expected_query` covers this change. 

### Issues Resolved ###

- fixes #179 

### Testing Procedure ###

To test pages are returned, and error 400 is handled appropriately (since `SavedSearchClient.execute` doesn't return a generator), the following code can be used to test;

```
import py42.sdk 
import py42.settings as settings
from py42.exceptions import Py42BadRequestError
import urllib3

urllib3.disable_warnings()

settings.verify_ssl_certs = False

sdk = py42.sdk.from_local_account("https://console.us.code42.com", "username", "password")

pg_num = 1
while True:
    try:
        resp = sdk.securitydata.savedsearches.execute("saved-search-id", pg_num=pg_num, pg_size=1000)
    except Py42BadRequestError:
        print("Got all {} pages!".format(pg_num))
        break
    pg_num+=1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
+- Added additional user-adjustable setting for security events page size:
+    - `py42.settings.security_events_per_page`
+
+- Page pg_num for saved search queries (pg_num):
+    - `py42.securitydata.savedsearches.get_query()`
+    - `py42.securitydata.savedsearches.execute()`
+
 - Methods for calling the agent-state APIs:
     - `sdk.devices.get_agent_state()`
     - `sdk.devices.get_agent_full_disk_access_state()`
@@ -74,7 +81,6 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- `py42.securitydata.savedsearches.get_query()` now accepts `pg_num` and `pg_size` arguments.
 - `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when the user is already on the list.
 - `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when the user already on the list.
 - `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when the user is already on the matter.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
+- `py42.securitydata.savedsearches.get_query()` now accepts `pg_num` and `pg_size` arguments.
 - `py42.detectionlists.departing_employee.add()` now raises `Py42UserAlreadyAddedError` when the user is already on the list.
 - `py42.detectionlists.high_risk_employee.add()` now raises `Py42UserAlreadyAddedError` when the user already on the list.
 - `py42.legalhold.add_to_matter()` now raises `Py42UserAlreadyAddedError` when the user is already on the matter.
--- a/src/py42/clients/savedsearch.py
+++ b/src/py42/clients/savedsearch.py
@@ -32,17 +32,19 @@ class SavedSearchClient(BaseClient):
         uri = u"{}/{}".format(self._resource, search_id)
         return self._session.get(uri)
 
-    def get_query(self, search_id):
+    def get_query(self, search_id, pg_num=1, pg_size=10000):
         """Get the saved search in form of a query(`py42.sdk.queries.fileevents.file_event_query`).
 
         Args:
             search_id (str): Unique search Id of the saved search.
+            pg_num (int, optional): The consecutive group of results of size pg_size in the result set to return. Defaults to 1.
+            pg_size (int, optional): The maximum number of results to be returned. Defaults to 10,000.
         Returns:
             :class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`
         """
         response = self.get_by_id(search_id)
         search = response[u"searches"][0]
-        return FileEventQuery.from_dict(search)
+        return FileEventQuery.from_dict(search, page_number=pg_num, page_size=pg_size)
 
     def execute(self, search_id, pg_num=1, pg_size=10000):
         """
@@ -55,5 +57,6 @@ class SavedSearchClient(BaseClient):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        query = self.get_query(search_id)
+        query = self.get_query(search_id, pg_num=pg_num, pg_size=pg_size)
+        print(query)
         return self._file_event_client.search(query)
--- a/src/py42/clients/savedsearch.py
+++ b/src/py42/clients/savedsearch.py
@@ -32,30 +32,28 @@ class SavedSearchClient(BaseClient):
         uri = u"{}/{}".format(self._resource, search_id)
         return self._session.get(uri)
 
-    def get_query(self, search_id, pg_num=1, pg_size=10000):
+    def get_query(self, search_id, pg_num=None):
         """Get the saved search in form of a query(`py42.sdk.queries.fileevents.file_event_query`).
 
         Args:
             search_id (str): Unique search Id of the saved search.
             pg_num (int, optional): The consecutive group of results of size pg_size in the result set to return. Defaults to 1.
-            pg_size (int, optional): The maximum number of results to be returned. Defaults to 10,000.
         Returns:
             :class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`
         """
         response = self.get_by_id(search_id)
         search = response[u"searches"][0]
-        return FileEventQuery.from_dict(search, page_number=pg_num, page_size=pg_size)
+        return FileEventQuery.from_dict(search, page_number=pg_num)
 
-    def execute(self, search_id, pg_num=1, pg_size=10000):
+    def execute(self, search_id, pg_num=None):
         """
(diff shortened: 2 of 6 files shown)


DiscoRiver:
I'm not sure if this is the best approach, or if we should update `SavedSearchClient.execute()` to return a generator.

Looking forward some feedback on this.

DiscoRiver:
I've converted this to a draft, as it seemed more appropriate for discussion.

src/py42/clients/savedsearch.py
```diff
@@ -32,17 +32,19 @@ def get_by_id(self, search_id):
         uri = u"{}/{}".format(self._resource, search_id)
         return self._session.get(uri)
 
-    def get_query(self, search_id):
+    def get_query(self, search_id, pg_num=1, pg_size=10000):
```
unparalleled-js, line 35: Since this was working before without the need for passing values here, we should default to `None` and let the other py42 mechanisms figure out the `pg_num` and `pg_size` values.  This way, if we ever need to change 10,000 to something else, we don't have to fix the bug in multiple places.

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
+- `py42.securitydata.savedsearches.get_query()` now accepts `pg_num` and `pg_size` arguments.
```
unparalleled-js, line 77: This might make more sense under `## Added`.

unparalleled-js:
@DiscoRiver Have you tested this without 10,000 as the page size?

unparalleled-js:
I see that `test_execute_calls_post_with_expected_query` does test the default values, but we should add a test that actually changes the `pg_num` and `pg_size` and makes sure the default values are not used in the request

DiscoRiver:
> @DiscoRiver Have you tested this without 10,000 as the page size?

@unparalleled-js Yep! I updated the `pytest` command in tox.ini to include `--capture=no`, and added some print arguments to make the the request was correct. The test code I was using is set to 1000 too.

Should I add a unit test to test the custom value?

DiscoRiver:
> I see that `test_execute_calls_post_with_expected_query` does test the default values, but we should add a test that actually changes the `pg_num` and `pg_size` and makes sure the default values are not used in the request

No problem, I'll build one for the custom value.

src/py42/clients/savedsearch.py
DiscoRiver: Ah, yes. I did consider that, but didn't change it. I'll update. Thanks!

src/py42/sdk/queries/__init__.py
```diff
@@ -5,17 +5,17 @@ class BaseQuery(object):
     def __init__(self, *args, **kwargs):
         self._filter_group_list = list(args)
         self._group_clause = kwargs.get(u"group_clause", u"AND")
-        self.page_size = 10000
+        self.page_number = kwargs.get(u"page_number", 1)
```
unparalleled-js, line 8: This kind of feeds is relates the problem I listed here: https://github.com/code42/py42/pull/187/files#r484455200

We should either _only_ set these properties when they are not None (so not providing a default value in the `get()` and only setting if `is not None`) --- which I think is a good idea since we know the default values are already getting set somewhere.... OR we should define constants for default `PG_NUM_START` and default `SECURITYDATA_PG_SIZE`.

Do you get what I am getting at?

DiscoRiver: I think so. Given how some other similar values are set, could we do something like this?

```
PG_NUM_START = 1
SECURITYDATA_PG_SIZE = 10000

self.page_number = page_number or PG_NUM_START
self.page_size = page_size or SECURITYDATA_PG_SIZE
```

DiscoRiver: Perhaps we could add the variable defining page size for security events to `sdk.settings`? 

Then `FileEventQuery` and `BaseQuery` could pull it from that setting? That would allow it to exist in a single place, and also for the user to define it like they would do with `settings.items_per_page`, but specifically for security queries.

unparalleled-js: Yeah that makes sense

DiscoRiver:
@unparalleled-js - I just updated with the new code, and addressed the other points. Here is the updated example code for how a user would interact. The only real difference is the `settings.security_events_per_page` setting.

```
import py42.sdk 
import py42.settings as settings
from py42.exceptions import Py42BadRequestError
import urllib3

urllib3.disable_warnings()

settings.verify_ssl_certs = False

sdk = py42.sdk.from_local_account("https://console.us.code42.com", "username", "password")

settings.security_events_per_page = 5000
pg_num = 1
while True:
    try:
        resp = sdk.securitydata.savedsearches.execute("saved-search-id", pg_num=pg_num)
    except Py42BadRequestError:
        print("Got all {} pages!".format(pg_num))
        break
    pg_num+=1
```

src/py42/settings/__init__.py
```diff
@@ -9,6 +9,7 @@
 verify_ssl_certs = True
 
 items_per_page = 500
+security_events_per_page = 10000
```
DiscoRiver, line 12: I considered naming this `security_items_per_page`, but I think `security_events_per_page` better indicates it's intended use.

src/py42/sdk/queries/__init__.py
```diff
@@ -1,21 +1,22 @@
 from py42.sdk.queries.query_filter import FilterGroup
+from py42 import settings
 
 
 class BaseQuery(object):
     def __init__(self, *args, **kwargs):
         self._filter_group_list = list(args)
         self._group_clause = kwargs.get(u"group_clause", u"AND")
-        self.page_size = 10000
+        self.page_number = kwargs.get(u"page_number") or 1
```
DiscoRiver, line 9: Updated this to properly set the value to 1 is `page_number` is `None`. This one shouldn't need separating the same as `page_size`.

unparalleled-js: Yeah, mostly agree, except that alerts use `0` as their starting page num

src/py42/clients/savedsearch.py
```diff
@@ -32,28 +32,28 @@ def get_by_id(self, search_id):
...
+            pg_num (int, optional): The consecutive group of results of size pg_size in the result set to return. Defaults to 1.
         Returns:
             :class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`
         """
         response = self.get_by_id(search_id)
         search = response[u"searches"][0]
-        return FileEventQuery.from_dict(search)
+        return FileEventQuery.from_dict(search, page_number=pg_num)
 
-    def execute(self, search_id, pg_num=1, pg_size=10000):
```
unparalleled-js, line 47: I think I was not clear in my last advice! I still think that `pg_size` still should be a parameter, but it should just default to None here since somewhere else it automatically turns into the default value on its own.

src/py42/sdk/queries/__init__.py
DiscoRiver: Sure, I looked at that and unless I'm mistaken, the changes to `BaseQuery` shouldn't affect `AlertQuery` because `self.page_number` is overridden in `AlertQuery`. 

Do you think we need to change this, or is it okay how it is now?

src/py42/clients/savedsearch.py
DiscoRiver: Oh sorry, I removed it in favour of `settings.security_events_per_page` because I thought being able to set it in two places would be potentially confusing, but now I realise it makes sense to add it back because then a user can set `settings.security_events_per_page` but still provide the parameter if needed for individual calls.

DiscoRiver:
Closing as I am now branching from the refactored master.

src/py42/clients/savedsearch.py
unparalleled-js: Exactly... The settings is the default. The user does not even have to know about the setting to use the parameter. I personally kind of expect `pg_num` and `pg_size` as parameters in methods that page, that is why we have `get_all()` that ignores them and `get_page()` for allowing them. Now with these changes, we can have them in saved search as well!

src/py42/sdk/queries/__init__.py
unparalleled-js: Ah, that is correct. I think it is good the way it is. I think it is also good that you removed setting page num in the FileEventQuery because that could potentially override what it is set here.. And it will always get set in the BaseQuery.  However, there is a chance that some day the Base Query and File Event Query have different page nums.. That seems highly unlikely, but I supposed a simple safer guard can be to ensure `page_num > 0` in `FileEventQuery`.

unparalleled-js: But actually, don't do that.... I think if a user passes in a negative or incorrect page num, the query should just explode.

PR #190, opened 2020-09-09
178 move filter attributes

### LINKED ISSUE #178: [Bug] filter_attributes function is located in an unexpected location
### Description

The `filter_attributes()` function, located here: https://github.com/code42/py42/blob/master/src/py42/sdk/queries/query_filter.py#L188 does not make a lot of sense.  It is used on constants all around py42.  It should be moved to `util.py` or somewhere better.

### Steps to Reproduce

1. Follow link and see that the function is in a weird spot

### Expected Behavior
This be importable from util.py

### Actual Behavior
Have to import from query_filter.py

DiscoRiver:
### Description of Change ###

- Refactored `filter_attributes()` from query_filter.py to util.py, and updated imports/usage.

### Issues Resolved ###

- Fixes #178 

### Testing Procedure ###

- Import using `from py42.util import filter_attributes` instead of `from py42.sdk.queries.query_filter import filter_attributes`


src/py42/util.py
```diff
@@ -72,3 +72,12 @@ def format_dict(dict_, label=None):
     if label:
         return u"{} {}".format(label, indented_dict)
     return indented_dict
+
+
+def filter_attributes(cls):
```
unparalleled-js, line 77: I wonder why this was originally called `filter_attributes()`. Just now realizing that something like `get_attributes_from_class()` would be a more appropriate name.

unparalleled-js: Do you mind making this change for this helpful internal util method?

DiscoRiver: No problem. I'm on it.

unparalleled-js: Or actually how about `get_attribute_keys_from_class()` ?

tests/sdk/test_util.py
```diff
@@ -10,3 +11,17 @@ def test_convert_timestamp_to_str_returns_expected_str():
 def test_convert_datetime_to_timestamp_str_returns_expected_str():
     d = datetime(2020, 4, 19, 13, 3, 2, 3)
     assert util.convert_datetime_to_timestamp_str(d) == "2020-04-19T13:03:02.000Z"
+
+
+class FilterClassTest(object):
```
unparalleled-js, line 16: See this comment: https://github.com/code42/py42/pull/190/files#r485867556

This test is what prompted that. I don't think this class needs to be a `Filter` class. This util method does not care about that.  Any class would work... So, a less confusing test class here would be `TestClass(object)` and the test below would be `test_get_attribute_keys_from_class_returns_public_class_attribute_keys()`

src/py42/util.py
```diff
@@ -72,3 +72,12 @@ def format_dict(dict_, label=None):
     if label:
         return u"{} {}".format(label, indented_dict)
     return indented_dict
+
+
```
unparalleled-js, line 76: So this is technically a public facing method that does not have documentation.  Do you mind also adding documentation for this method since it is now public facing?  It could be simple like """Returns attribute names on the given class""" And then have the normal `Args:` and `Returns:` parts.

unparalleled-js: The docs will show up the `methoddocs/util` but you should not have to do anything except add the doc string for it to work.

unparalleled-js:
Also, don't forgot to resolve the merge conflicts!

DiscoRiver:
I've done something catastrophic so I'm going to start again. I'll update the issues as commented.

unparalleled-js:
> I've done something catastrophic so I'm going to start again. I'll update the issues as commented.

Haha no worries

PR #193, opened 2020-09-10
Implement CLA workflow

timabrmsn:
### Description of Change ###

Implements the [cla-assistant GitHub Action](https://github.com/cla-assistant/github-action) in this repo. 

All existing contributors are in the allowlist, so won't be prompted on new PRs to master, but anyone not in the allowlist will
be required to reply to their PR with the following comment before the PR can be merged:

"I have read the CLA Document and I hereby sign the CLA"

Once they've commented, the action will add their "signature" to the `.github/signatures/cla.json` file and future PRs will not require signing any longer.


.github/workflows/cla.yml
```diff
@@ -0,0 +1,34 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # the below token should have repo scope and must be manually added by you in the repository's secret
+          PERSONAL_ACCESS_TOKEN : ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: '.github/signatures/cla.json'
+          path-to-cla-document: 'https://github.com/code42/py42/blob/master/CONTRIBUTOR_LICENSE_AGREEMENT.md'
+          # branch should not be protected
+          branch: 'master'
+          allowlist: alang13,unparalleled-js,kiran-chaudhary,timabrmsn,ceciliastevens,DiscoRiver,annie-payseur,amoravec
```
unparalleled-js, line 25: Could add Sagar just to be safe

timabrmsn: Good catch, I was just pulling from the list of users who have already contributed.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,33 @@
...
+        # Alpha Release
+        uses: cla-assistant/github-action@v2.0.1-alpha
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # the below token should have repo scope and must be manually added by you in the repository's secret
+          PERSONAL_ACCESS_TOKEN : ${{ secrets.PERSONAL_ACCESS_TOKEN }}
+        with:
+          path-to-signatures: '.github/signatures/cla.json'
+          path-to-cla-document: 'https://github.com/code42/py42/blob/master/CONTRIBUTOR_LICENSE_AGREEMENT.md'
+          # branch should not be protected
```
alanag13, line 23: our master branch _is_ protected. Will this work?

This looks a little different than what I was originally thinking we would be doing. It looks like this works by storing the "signatures" of everyone in the repo itself, which is why the branch can't be protected.

This looks a little different than the approach I looked at initially -- I think all that would be needed would be for an admin of the github org to log in to https://cla-assistant.io and configure our repo there (no changes to the repo itself), and to create a gist where the CLA would be hosted from. That site would then allow anyone with the proper permission to also view a list of contributor info from there.

timabrmsn: I'm not sure, as I can't see what protections master has. I thought this might be better since it's self-contained and not dependent on a third party host (the action is developed by the same people as https://cla-assistant.io).

We could have it write the signatures to a separate repository.. which is actually probably better, as we could point any repos that need CLA agreement to that same repo and then users would only need to sign once. 

Although we might need to tweak the language saying that agreeing to contribute to one Code42 repo grants agreement for _all_ Code42 repos?

alanag13: "protected" just means "cannot be pushed to directly (i.e. can only be changed via a pull request)".

The "alphaness" of the github actions solution gives me a little pause -- it seems like the hosted version is probably more popular and likely to have better support. 

However, I do like the idea of having the signatures in a separate repo -- the added benefit of only needing to accept once of the whole organization instead of once per project is nice. I don't think the hosted version offers such a solution. If this can be a private repo with no protection on the master branch, that seems ideal. If we can make this work that way that sounds best imo, otherwise there's still the hosted version to fall back to.

timabrmsn: I think it will need to be public, since we'll want to host the CLA agreement text within it that the action will link to.

unparalleled-js: Are there any legal ramifications for having a mono-CLA for all repos? Is this a decision we are able to make on our own or should we seek advice regarding this?

alanag13: @unparalleled-js good point, I _feel_ like there shouldn't be any problem there but I'd rather know for sure. Let's check with Tony on this.
@timabrmsn let's also check with Tony re: which of these approaches is a more favorable solution.

timabrmsn: I chatted with @amoravec and he didn't think there was any concern about a `sign once` option, especially since the CLA language itself is pretty broad about contributions to Code42 projects, not a specific repo (the agreement covers mailing list threads and issues, not just code).

unparalleled-js (approved):

PR #197, opened 2020-09-14
Feature audit logs

kiran-chaudhary:
### Description of Change ###

Added audit-logs feature in py42.

- closes #
Fixes #1180

### Testing Procedure ###

To test without any filters: 
`sdk.auditlogs.get_all()`
`sdk.auditlogs.get_page()`
Edit: `sdk.auditlogs.export()` with `format="csv"` or `format="cef"` - Removed

To test filtered by dates:

```
from datetime import datetime
start_time ="1591401600"
end_time ="1591401601"
sdk.auditlogs.get_all(begin_time=start_time, end_time=end_time)
```

Similarly other filters are 'user_ids, user_names, affected_user_ids, affected_user_names, user_ip_addresses' all of which accept a string or a list of string.

### PR Checklist ###


kiran-chaudhary:
Need to discuss whether we can use `filters` for `dateRange` and `eventTypes`? Whether they should be passed by user to `get_all`, current implementation accepts simple data types.

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
...
+        user_ids=None,
+        user_names=None,
+        user_ip_addresses=None,
+        affected_user_ids=None,
+        affected_user_names=None,
+    ):
+        """Retrieve audit logs, filtered based on given arguments.
+
+        Args:
+            begin_time (datetime, optional): A datetime.datetime instance for a given timestamp. Defaults to None.
```
unparalleled-js, line 19: No other py42 API uses `datetime`. For consistency, these be unix timestamps.

src/py42/services/audit_logs.py
```diff
@@ -0,0 +1,63 @@
...
+        user_ids = user_ids.split(comma) if user_ids else []
+        user_names = user_names.split(comma) if user_names else []
+        user_ip_addresses = user_ip_addresses.split(comma) if user_ip_addresses else []
+        affected_user_ids = affected_user_ids.split(comma) if affected_user_ids else []
+        affected_user_names = (
+            affected_user_names.split(comma) if affected_user_names else []
+        )
+
+        date_range = {}
+        if begin_time:
```
unparalleled-js, line 48: `begin_time` and `end_time` should already be timestamps.

src/py42/services/audit_logs.py
```diff
@@ -0,0 +1,63 @@
...
+            pageSize=page_size,
+            dateRange=kwargs["date_range"],
+            eventTypes=kwargs["event_types"],
+            actorIds=kwargs["user_ids"],
+            actorNames=kwargs["user_names"],
+            actorIpAddresses=kwargs["user_ip_addresses"],
+            affectedUserIds=kwargs["affected_user_ids"],
+            affectedUserNames=kwargs["affected_user_names"],
+        )
+        params["type$"] = "audit_log::audit_log_queries.search_audit_log/1"
```
unparalleled-js, line 23: Can you explain this line?  What is the /1 a the end mean?

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
+class AuditLogsClient(object):
+    def __init__(self, audit_log_service):
+        self._audit_log_service = audit_log_service
+
+    def get_all(
```
unparalleled-js, line 5: Can we add a method for getting the page here?

tests/services/test_auditlogs.py
```diff
@@ -0,0 +1,99 @@
+from py42.services.audit_logs import AuditLogsService
+
+
+class TestAuditLogService(object):
```
unparalleled-js, line 4: For record keeping purposes, it'd be nice to have a test for that is only for `get_page()`... I know that's probably covered in the other tests, but it would still be better to have.

src/py42/services/audit_logs.py
```diff
@@ -0,0 +1,63 @@
...
+        begin_time=None,
+        end_time=None,
+        event_types=None,
+        user_ids=None,
+        user_names=None,
+        user_ip_addresses=None,
+        affected_user_ids=None,
+        affected_user_names=None,
+    ):
+        comma = u","
```
alanag13, line 37: I think it makes more sense for all of these to be arrays instead of strings that got split on a comma. We can also accept plain strings and turn them into arrays of just one item.

src/py42/services/audit_logs.py
```diff
@@ -0,0 +1,63 @@
+from py42 import settings
```
unparalleled-js, line 1: This file should be named `auditlogs.py`, like the other multi-worded files here, for example `legalhold.py`, `securitydata.py`, and `savedsearch.py`.

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
+class AuditLogsClient(object):
```
unparalleled-js, line 1: This file should be named `auditlogs.py` like the other multi-worded files here, such as `securitydata.py` or `detectionlists.py`.

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, audit_log_service):
+        self._audit_log_service = audit_log_service
+
+    def get_all(
+        self,
+        begin_time=None,
+        end_time=None,
+        event_types=None,
+        user_ids=None,
+        user_names=None,
```
unparalleled-js, line 11: `usernames` if we want to be consistent with other parameters such as `cloud_usernames`.

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
...
+    ):
+        """Retrieve audit logs, filtered based on given arguments.
+
+        Args:
+            begin_time (datetime, optional): A datetime.datetime instance for a given timestamp. Defaults to None.
+            end_time (datetime, optional): A datetime.datetime instance for a given timestamp. Defaults to None.
+            event_types (str, optional): Comma separated str of valid event types. Defaults to None.
+            user_ids (str, optional): Comma separated str of user ids. Defaults to None.
+            user_names (str, optional): Comma separated str of user names. Defaults to None.
+            user_ip_addresses (str, optional): Comma separated str of user ip addresses. Defaults to None.
```
unparalleled-js, line 24: `Comma-separated` (with hyphen) to be more grammatically correct.

src/py42/services/audit_logs.py
alanag13: I think this is an oddity caused by some of the in-house code generation tools we use. Have we tried see if this request works without it? My guess is that it won't but I think this is a good example of API weirdness that we should avoid as an organization.

src/py42/clients/audit_logs.py
```diff
@@ -0,0 +1,41 @@
...
+        user_ip_addresses=None,
+        affected_user_ids=None,
+        affected_user_names=None,
+    ):
+        """Retrieve audit logs, filtered based on given arguments.
+
+        Args:
+            begin_time (datetime, optional): A datetime.datetime instance for a given timestamp. Defaults to None.
+            end_time (datetime, optional): A datetime.datetime instance for a given timestamp. Defaults to None.
+            event_types (str, optional): Comma separated str of valid event types. Defaults to None.
```
alanag13, line 21: see me not in the `Service` class, I think all these parameters should be lists instead comma separated strings

src/py42/services/audit_logs.py
```diff
@@ -0,0 +1,63 @@
...
+            page=page_num - 1,
+            pageSize=page_size,
+            dateRange=kwargs["date_range"],
+            eventTypes=kwargs["event_types"],
+            actorIds=kwargs["user_ids"],
+            actorNames=kwargs["user_names"],
+            actorIpAddresses=kwargs["user_ip_addresses"],
+            affectedUserIds=kwargs["affected_user_ids"],
+            affectedUserNames=kwargs["affected_user_names"],
+        )
```
timabrmsn, line 22: I anticipate that as we (as a company) add functionality to the audit log system, this set of available params will expand. We should maybe design this so that if additional unknown kwargs are passed, they're just added to the params, allowing it to be possible to use new params before py42 is updated. 

Something like: 

```
date_range = kwargs.pop("date_range")
event_types = kwargs.pop("event_types")
... etc

params = dict(
            page=page_num - 1,
            pageSize=page_size,
            dateRange=date_range,
            eventTypes=event_types,
            ....etc
            **kwargs)
```

kiran-chaudhary: It depends on how API behaves, as this particular API does so right now, if API fails if even one parameter/field name is incorrect then the suggested approach will fail when incorrect param names are passed, e.g we need to convert `event_types` to `eventTypes` such names cannot be anticipated advance and we would end up passing 'event_types'.

Having said that this is a good point, since we are building API gateway @alanag13 can make a note. The contract of every API should be that adding new optional parameters in future should not break/fail when those new fields/params are not passed. That way at-least we will save the existing implementation from breaking with new releases.

src/py42/services/audit_logs.py
kiran-chaudhary: Done.

src/py42/clients/audit_logs.py
kiran-chaudhary: Renamed.

src/py42/clients/audit_logs.py
kiran-chaudhary: changed.

src/py42/clients/audit_logs.py
kiran-chaudhary: renamed.

src/py42/services/audit_logs.py
kiran-chaudhary: Renamed.

src/py42/services/audit_logs.py
kiran-chaudhary: This works without the `type$` field. Developed it based on `console.us.code42.com`. I will refer the swagger docs to check its significance.

src/py42/services/audit_logs.py
unparalleled-js: If the API fails because the user put something weird in `**kwargs`, I think that is on them.  Having `kwargs` allows more people to have support for the time being and to be honest, I think every API method could have them.

kiran-chaudhary: So does it mean `get_all` method should accept `**kwargs`?

src/py42/services/audit_logs.py
kiran-chaudhary: Removed.

src/py42/services/audit_logs.py
kiran-chaudhary: Added.

tests/services/test_auditlogs.py
kiran-chaudhary: added.

src/py42/clients/audit_logs.py
kiran-chaudhary: Done.

src/py42/services/audit_logs.py
kiran-chaudhary: changed.

src/py42/clients/audit_logs.py
kiran-chaudhary: done.

src/py42/services/audit_logs.py
kiran-chaudhary: `sdk.auditlogs.get_all(deviceUserName='xyz')` will return all results, hope that is expected.

timabrmsn: The main goal here is making our own lives easier if the api gets updated. This will allow us to "patch" any integrations using an older py42 with the new values rather than forcing a full update of py42 (which could take awhile to get released, could have other changes that make upgrading more work, etc.). 

End-users don't really need to know why kwargs are accepted, but we _should_ have all currently known args spelled out as optional args, to make it easy for them to use (we don't want them to have to always look to the docs to get possible arg values). 

That said, we _do_ need to be a bit careful on how we process these. For example:

An integration consuming audit logs is using the bare function call to get all results:

```
results = sdk.auditlogs.get_all()
```

The auditlogs service is updated to include new event types, and adds a new param option: `affectedOrgs`. Unfortunately they made it a required param, so our integration is currently broken. 

We can patch the integration by just adding the missing param to the extra kwargs:

```
results = sdk.auditlogs.get_all(affectedOrgs=None)
```

However, on the next py42 update, if we add an official `affected_orgs` param but our downstream integration doesn't change the param name when updating py42, it could break, because then the params dict:

```
params = dict(
            page=page_num - 1,
            pageSize=page_size,
            dateRange=date_range,
            eventTypes=event_types,
            actorIds=user_ids,
            actorNames=usernames,
            actorIpAddresses=user_ip_addresses,
            affectedUserIds=affected_user_ids,
            affectedUserNames=affected_usernames,
            affectedOrgs=affected_orgs,
            **kwargs
        )
``` 
will now be given two instances of `affectedOrgs` args. 

So to be safe we should instead of passing the final `**kwargs` into the dict() call, we should construct the dict with all of our known params, then update it with the `**kwargs`:

```
params = dict(
            page=page_num - 1,
            pageSize=page_size,
            dateRange=date_range,
            eventTypes=event_types,
            actorIds=user_ids,
            actorNames=usernames,
            actorIpAddresses=user_ip_addresses,
            affectedUserIds=affected_user_ids,
            affectedUserNames=affected_usernames,
            affectedOrgs=affected_orgs
)
params.update(kwargs)
```

That way we _can_ patch if needed, but also not break the patched code once py42 is upgraded later on.

src/py42/services/auditlogs.py
```diff
@@ -0,0 +1,67 @@
...
+from py42.util import to_list
+
+
+class AuditLogsService(BaseService):
+    def get_page(self, page_num, page_size=None, **kwargs):
+
+        event_types = to_list(kwargs.pop(u"event_types"))
+        user_ids = to_list(kwargs.pop(u"user_ids"))
+        usernames = to_list(kwargs.pop(u"usernames"))
+        user_ip_addresses = to_list(kwargs.pop(u"user_ip_addresses"))
```
unparalleled-js, line 14: Should these list types default to empty list when not provided in `kwargs`?  Not a big deal since this is internal, but would be safer.

unparalleled-js (approved):

src/py42/services/audit_logs.py
alanag13: I'm fine with adding `**kwargs`, but we should keep the known arguments as named. This is mainly to provide a workaround in the event that an api feature gets introduced before we add an accompanying named parameter, and if the user uses parameter names that don't exist, that will be on them.

src/py42/services/auditlogs.py
alanag13: agreed

src/py42/services/auditlogs.py
```diff
@@ -0,0 +1,67 @@
+from py42 import settings
+from py42.services import BaseService
+from py42.services.util import get_all_pages
+from py42.util import convert_timestamp_to_str
+from py42.util import to_list
+
+
+class AuditLogsService(BaseService):
```
timabrmsn, line 8: Can we add a docstring with the swagger docs for this microservice?

src/py42/services/auditlogs.py
kiran-chaudhary: ummm changed.

src/py42/services/audit_logs.py
kiran-chaudhary: First we would do the patch work and then the proper work kind of seems weird, changed nevertheless.

src/py42/services/auditlogs.py
```diff
@@ -29,14 +33,21 @@ def get_page(self, page_num, page_size=None, **kwargs):
...
             pageSize=page_size,
             dateRange=date_range,
-            eventTypes=event_types,
-            actorIds=user_ids,
-            actorNames=usernames,
-            actorIpAddresses=user_ip_addresses,
-            affectedUserIds=affected_user_ids,
-            affectedUserNames=affected_usernames,
-            **kwargs
+            eventTypes=to_list(kwargs.get(u"event_types")),
```
unparalleled-js, line 36: Why not just have the expected params be listed out in the definition? `kwargs` is just for unexpected params, like ones we don't support yet.

unparalleled-js (changes requested):
`get_page()` to list out expected params so no need for popping them off kwargs

src/py42/services/auditlogs.py
kiran-chaudhary: Given that we found the docs are not right or helpful, is it ok if this is not done?

src/py42/services/auditlogs.py
kiran-chaudhary: changed.

src/py42/services/auditlogs.py
kiran-chaudhary: Added.

src/py42/clients/auditlogs.py
```diff
@@ -0,0 +1,139 @@
...
+            event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
+            user_ids (str or list, optional): A str or list of str of user ids. Defaults to None.
+            usernames (str or list, optional): A str or list of str of user names. Defaults to None.
+            user_ip_addresses (str or list, optional): A str or list of str of user ip addresses. Defaults to None.
+            affected_user_ids (str or list, optional): A str or list of str of affected user ids. Defaults to None.
+            affected_usernames  (str or list, optional): A str or list of str of affected user names. Defaults to None.
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        # As per document, https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Search_Audit_Log_events_with_the_Code42_API,
```
kiran-chaudhary, line 123: These comments are meant for developers.

src/py42/clients/auditlogs.py
```diff
@@ -0,0 +1,93 @@
...
+        the given `page_num` in the implementation as the Code42 Audit-Logs API expects
+        the start page to be zero.
+
+        Args:
+            page_num (int, optional): The page number to get. Defaults to 1.
+            page_size (int, optional): The number of items per page. Defaults to `py42.settings.items_per_page`.
+            begin_time (int, optional): Timestamp in milliseconds. Defaults to None.
+            end_time (int, optional): Timestamp in milliseconds. Defaults to None.
+            event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
+            user_ids (str or list, optional): A str or list of str of user ids. Defaults to None.
```
unparalleled-js, line 32: Are these the legacy IDs or the user UIDs ?

unparalleled-js (approved):

src/py42/clients/auditlogs.py
kiran-chaudhary: Thanks, changed to user uid.

PR #198, opened 2020-09-14
add support for 2FA

### LINKED ISSUE #195: [Enhancement] Support 2FA
## Summary
Code42 supports [two-factor authentication](https://support.code42.com/Administrator/Cloud/Configuring/Two-factor_authentication_for_local_users) for locally-authorized accounts. Py42 currently errors out if you attempt to set up an SDK connection using an account with two-factor.

We could investigate implementing support for accounts which require this auth type.

## Proposed API

By default, if a user requires 2FA, prompt them:

```
>>> c42 = sdk.from_local_account(hostAddress, username, password)
Please input your authentication code: ******
>>> c42.user.get_current()
```

After time has passed, prompt again if necessary when renewing the v3_user_token.

Additionally, accept an optional function that returns the 2FA code (for automation):

```
>>> c42 = sdk.from_local_account(hostAddress, username, password, totp_function)
>>> c42.user.get_current()
```

If a function is provided, it will be called again when the token needs to be renewed, allowing developers to provide interfaces to alternative TOTP solutions/implementations.

## Intended Use Case
This could be used by Code42 users who wish to automate API interactions, but also want an additional layer of security protecting API accounts.

ceciliastevens:
### Description of Change ###

This change adds support for Code42's two-factor authentication implementation using time-based one-time passwords

### Issues Resolved ###

- closes #195

### Testing Procedure ###
1. Create a Code42 account in an organization that has two-factor authentication turned on
2. Attempt to create a SDK object using this account
3. Verify that Py42MFARequiredError is raised
4. Attempt to create a SDK object and include the TOTP token (`sdk.from_local_account(hostaddress,username,password,totp="123456"`)
5. Verify that the SDK object is created successfully

### PR Checklist ###

I have not added tests yet as I am not sure how best to do so. I would appreciate recommendations/input.


src/py42/sdk/__init__.py
```diff
@@ -72,14 +74,17 @@ def from_local_account(cls, host_address, username, password):
                 console.us.code42.com
             username (str): The username of the authenticating account.
             password (str): The password of the authenticating account.
-
+            totp (str): The time-based one-time password of the authenticating account. Include only
```
unparalleled-js, line 77: The type of `totp` should be `(str, optional)` and the doc should end with the sentence ` Defaults to None.`

src/py42/sdk/__init__.py
```diff
@@ -72,14 +74,17 @@ def from_local_account(cls, host_address, username, password):
...
+                if the account uses Code42's two-factor authentication.
         Returns:
             :class:`py42.sdk.SDKClient`
         """
         basic_auth = None
         if username and password:
             basic_auth = HTTPBasicAuth(username, password)
         auth_connection = Connection.from_host_address(host_address, auth=basic_auth)
+        if totp:
+            auth_connection._headers.update({"totp-auth": str(totp)})
```
unparalleled-js, line 87: Might be worth adding a `add_header()` method to `py42.services._connection.Connection` class so we don't have to refer to internal attribute `_headers` from outside the class.

src/py42/sdk/__init__.py
ceciliastevens: Updated!

src/py42/sdk/__init__.py
ceciliastevens: Added the method and updated the call in __init__.py.

src/py42/services/_connection.py
```diff
@@ -247,7 +248,13 @@ def _handle_error(method, url, response):
     try:
         response.raise_for_status()
     except HTTPError as ex:
-        raise_py42_error(ex)
+        if response.json()["error"][0]["primaryErrorKey"] in [
```
alanag13, line 254: we currently need to support versions of `requests` where the `.json()` method doesn't exist :-(.
The easiest way to port this is:
```python
response_dict = json.loads(response.text)
```

alanag13, line 254: the whole if block could be added to `raise_py42_error` so that it continues to be the main function for determining what error should be raised

src/py42/sdk/__init__.py
alanag13: instead of doing this here, we can add an optional constructor param to `V3Auth` and use the `headers` keyword arg in the `get` call that requests the token.

src/py42/services/_connection.py
ceciliastevens: Moved everything over there in 3513cf0

ceciliastevens: Updated in 3513cf0

src/py42/sdk/__init__.py
timabrmsn: What if we had a new `C42RenewableAuth` TOTP subclass for this, that could take either a single token (and if the auth expired would not be able to auto-renew), but if given a callable that returned a new totp _would_ be able to auto-renew?

alanag13:
> I have not added tests yet as I am not sure how best to do so. I would appreciate recommendations/input.

As suggested above, moving the implementation to `V3Auth` should make it more testable.

- Test that `raise_py42_error` raises the new exception (using `pytest.raises` when passed a mock `HTTPError` with the qualifying criteria on it, i.e. a version of `error_response` containing the failure text).
- Change `test_call_calls_auth_api_with_expected_url` on `TestV3Auth` to test that `get` gets called with appropriate `headers` param when the new optional `totp` param is passed to the constructor.
- Add a test that nothing gets passed to `headers` when there is no `totp` param (essentially, keep the old test)

src/py42/services/_connection.py
```diff
@@ -104,6 +104,9 @@ def from_microservice_prefix(cls, connection, prefix, auth=None, session=None):
     def host_address(self):
         return self._get_host_address()
 
+    def add_header(self, header):
```
alanag13, line 107: Sorry to make you rework things, but I don't think this should be needed, see my discussion on how to approach needing the extra header by implementing this in `V3Auth`.

src/py42/sdk/__init__.py
ceciliastevens: Would it be a subclass of `C42RenewableAuth` or of `V3Auth`? Since we do want to use V3Auth regardless?

alanag13: @timabrmsn we don't need a new subclass, but I like the idea of changing the optional constructor param to be the callable you're talking about instead of the totp itself. This would allow the user to do something as simple as write a function that prompts for input _or_ something more complex like actually determining their totp from the secret key, and if you know that you will be needing 2fa this approach makes it so that you don't _have_ to catch `Py42MFAError`.

@ceciliastevens let me know if you have any questions about making that change.

ceciliastevens: Makes sense to me. I've made the change that I think accomplishes this, if it looks good to you I'll go ahead and work on adding some tests next.

ceciliastevens: oh wait just saw your latest comment. Give me a moment to revise out the add_header method

src/py42/services/_connection.py
ceciliastevens: Makes sense! I've updated V3Auth to add the header directly to the request, and removed the add_header method from Connection

ceciliastevens:
> > I have not added tests yet as I am not sure how best to do so. I would appreciate recommendations/input.
> 
> As suggested above, moving the implementation to `V3Auth` should make it more testable.
> 
> * Test that `raise_py42_error` raises the new exception (using `pytest.raises` when passed a mock `HTTPError` with the qualifying criteria on it, i.e. a version of `error_response` containing the failure text).
> * Change `test_call_calls_auth_api_with_expected_url` on `TestV3Auth` to test that `get` gets called with appropriate `headers` param when the new optional `totp` param is passed to the constructor.
> * Add a test that nothing gets passed to `headers` when there is no `totp` param (essentially, keep the old test)

I've added these tests

timabrmsn:
@ceciliastevens can you update the `basics` and `gettingstarted` user guides in the docs to include 2FA instructions?

src/py42/sdk/__init__.py
```diff
@@ -43,11 +43,13 @@ def from_local_account(host_address, username, password):
             console.us.code42.com
         username (str): The username of the authenticating account.
         password (str): The password of the authenticating account.
+        totp (str, optional): The time-based one-time password of the authenticating account. Include only
```
alanag13, line 46: this can now be a callable or str

src/py42/sdk/__init__.py
```diff
@@ -72,15 +74,16 @@ def from_local_account(cls, host_address, username, password):
                 console.us.code42.com
             username (str): The username of the authenticating account.
             password (str): The password of the authenticating account.
-
+            totp (str, optional): The time-based one-time password of the authenticating account. Include only
```
alanag13, line 77: callable or str

ceciliastevens:
> @ceciliastevens can you update the `basics` and `gettingstarted` user guides in the docs to include 2FA instructions?

Updated!

src/py42/sdk/__init__.py
ceciliastevens: Updated.

src/py42/sdk/__init__.py
ceciliastevens: Updated.

docs/userguides/basics.md
```diff
@@ -22,6 +22,12 @@ import py42.sdk
 sdk = py42.sdk.from_local_account("https://console.us.code42.com", "my_username", "my_password")
 ```
 
+If your account uses [two-factor authentication](https://support.code42.com/Administrator/Cloud/Configuring/Two-factor_authentication_for_local_users), include the time-based one-time password:
+
+```
+sdk = py42.sdk.from_local_account("https://console.us.code42.com", "my_username", "my_password",totp="123456")
```
unparalleled-js, line 28: Missing space after comma before `totp=`

src/py42/services/_auth.py
```diff
@@ -29,12 +29,15 @@ def _get_credentials(self):
...
         super(V3Auth, self).__init__()
         self._auth_connection = auth_connection
+        self._totp = totp if callable(totp) else lambda: totp
 
     def _get_credentials(self):
         uri = u"/c42api/v3/auth/jwt"
         params = {u"useBody": True}
-        response = self._auth_connection.get(uri, params=params)
+        currentToken = self._totp()
+        headers = {"totp-auth": str(currentToken)} if currentToken else None
```
unparalleled-js, line 41: Should we do a retry here in case the current token expired after getting the current token and making the request to get the v3 token?

src/py42/exceptions.py
```diff
@@ -140,13 +142,28 @@ def __init__(self, exception, rule_id, source):
...
 def raise_py42_error(raised_error):
     """Raises the appropriate :class:`py42.exceptions.Py42HttpError` based on the given
     HTTPError's response status code.
     """
     if raised_error.response.status_code == 400:
         raise Py42BadRequestError(raised_error)
     elif raised_error.response.status_code == 401:
+        if raised_error.response.text:
+            response_dict = json.loads(raised_error.response.text)
+            if response_dict["error"][0]["primaryErrorKey"] in [
```
alanag13, line 162: I just realized that this will blow up if we ever get a 401 from anywhere that does not have this json structure. We need to make sure that this logic is nicely skipped if that happens. The easiest way is probably not just check for those error strings being _somewhere_ in the response (no need to use json.loads and check specific properties).

ceciliastevens: Updated

docs/userguides/basics.md
ceciliastevens: Updated

src/py42/services/_auth.py
ceciliastevens: Tokens are still accepted by the server for a period of time (not sure exactly how long but pretty considerable) even after they have "expired" in the authenticator app, so in order for us to hit this we'd need a very significant amount of time between the user inputting the code and py42 actually using it. That said, I'm not opposed to having a retry, just not sure of the best way to implement it.

docs/userguides/basics.md
```diff
@@ -22,6 +22,12 @@ import py42.sdk
 sdk = py42.sdk.from_local_account("https://console.us.code42.com", "my_username", "my_password")
 ```
 
+If your account uses [two-factor authentication](https://support.code42.com/Administrator/Cloud/Configuring/Two-factor_authentication_for_local_users), include the time-based one-time password:
+
+```
+sdk = py42.sdk.from_local_account("https://console.us.code42.com", "my_username", "my_password", totp="123456")
```
alanag13, line 28: Could you provide an example that uses a callable? I feel like that is going to be the far more common implementation.

src/py42/services/_auth.py
alanag13: I'm not sure what the exact window of time is in our, but it's typical for TOTPs to be valid for some amount of time _after_ a new one is generated, (i.e., for some period of time, there will actually be _two_ codes that work) to handle this scenario. I'm fine with leaving this as-is and adding something to handle it if we actually experience problems.

docs/userguides/basics.md
alanag13: also if you put `python` after the three tick marks, you will get python syntax highlighting, e.g.

```
print("Hello World!")
```

vs

```python
print("Hello World!")
```

ceciliastevens: updated!

alanag13:
This is looking really good! I'm going to carve out some time to try it out myself soon, I'll let you know if anything else is needed.

alanag13:
- - -
   ***I have read the CLA Document and I hereby sign the CLA***
   - - -

alanag13:
recheckcla

alanag13:
recheckla

alanag13:
I have read the CLA Document and I hereby sign the CLA

alanag13 (approved):

PR #199, opened 2020-09-14
Upgrade web restore API usage to v9

unparalleled-js:
### Description of Change ###

Upgrades the endpoint for starting a web restore to use v9. This mirrors the console UI. It is in our developers' best interest to mirror the console UI whenever possible such that debugging requests can be much easier.

### Issues Resolved ###
N/a

### Testing Procedure ###
Verify you can still restore files
Verify you can still restore multiple files
Verify you can cancel a restore job after starting it.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -30,7 +30,7 @@ class ArchiveAccessorManager(object):
         encryption_key=None,
     ):
         service = self._storage_service_factory.create_archive_service(
-            device_guid=device_guid, destination_guid=destination_guid
+            device_guid, destination_guid=destination_guid
         )
         decryption_keys = self._get_decryption_keys(
             device_guid=device_guid,
@@ -38,16 +38,14 @@ class ArchiveAccessorManager(object):
             encryption_key=encryption_key,
         )
         session_id = self._create_restore_session(
-            storage_archive_service=service, device_guid=device_guid, **decryption_keys
+            service, device_guid, **decryption_keys
         )
         restore_job_manager = create_restore_job_manager(
             storage_archive_service=service,
             device_guid=device_guid,
             archive_session_id=session_id,
         )
-        file_size_poller = create_file_size_poller(
-            storage_archive_service=service, device_guid=device_guid
-        )
+        file_size_poller = create_file_size_poller(service, device_guid)
         return ArchiveAccessor(
             device_guid=device_guid,
             archive_session_id=session_id,
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -81,10 +81,7 @@ def _create_file_selections(file_paths, metadata_list, file_sizes=None):
             u"selected": True,
         }
         selection = FileSelection(
-            path_set,
-            size_info[u"numFiles"],
-            size_info[u"numDirs"],
-            size_info[u"size"],
+            path_set, size_info[u"numFiles"], size_info[u"numDirs"], size_info[u"size"],
         )
         file_selections.append(selection)
 
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -303,7 +303,7 @@ class RestoreJobManager(_RestorePoller):
         is_done = response[u"done"]
         percentage_dict = {
             u"jobId": job_id,
-            u"status": response.data.get("status"),
+            u"status": response.data.get(u"status"),
             u"percentComplete": response[u"percentComplete"] if not is_done else 100,
         }
         debug.logger.debug(format_dict(percentage_dict))
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -13,8 +13,8 @@ FileSelection = namedtuple(
 
 
 class FileType(object):
-    DIRECTORY = u"directory"
-    FILE = u"file"
+    DIRECTORY = u"DIRECTORY"
+    FILE = u"FILE"
 
 
 class ArchiveAccessorManager(object):
@@ -316,12 +316,14 @@ class RestoreJobManager(_RestorePoller):
         num_dirs = sum([fs.num_dirs for fs in file_selections])
         num_bytes = sum([fs.num_bytes for fs in file_selections])
         return self._storage_archive_service.start_restore(
-            guid=self._device_guid,
+            device_guid=self._device_guid,
             web_restore_session_id=self._archive_session_id,
-            restore_groups={
-                u"backupSetId": backup_set_id,
-                u"files": [f.path_set for f in file_selections],
-            },
+            restore_groups=[
+                {
+                    u"backupSetId": backup_set_id,
+                    u"files": [f.path_set for f in file_selections],
+                }
+            ],
             num_files=num_files,
             num_dirs=num_dirs,
             num_bytes=num_bytes,
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -283,10 +283,6 @@ class RestoreJobManager(_RestorePoller):
         )
         self._archive_session_id = archive_session_id
 
-    @property
-    def destination_guid(self):
-        return self._storage_archive_service.destination_guid
-
     def get_stream(self, file_selections):
         response = self._start_restore(file_selections)
         job_id = response[u"jobId"]
--- a/src/py42/clients/_archive_access.py
+++ b/src/py42/clients/_archive_access.py
@@ -29,10 +29,7 @@ class ArchiveAccessorManager(object):
         private_password=None,
         encryption_key=None,
     ):
-        (
-            service,
-            destination_guid,
-        ) = self._storage_service_factory.create_archive_service(
+        service = self._storage_service_factory.create_archive_service(
             device_guid, destination_guid=destination_guid
         )
         decryption_keys = self._get_decryption_keys(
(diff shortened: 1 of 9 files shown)


src/py42/services/storage/archive.py
```diff
@@ -109,38 +113,31 @@ def create_restore_session(
...
-        timestamp=None,
-        backup_set_id=None,
+        restore_to_server=None,
     ):
-        """Submits a web restore job.
-        See https://console.us.code42.com/apidocviewer/#WebRestoreJob-post
-        """
-        uri = u"/api/WebRestoreJob"
+        """Submits a web restore job."""
+        uri = u"/api/v9/restore/web"
```
kiran-chaudhary, line 128: can we add test for the change.

src/py42/clients/archive.py
```diff
@@ -92,8 +92,11 @@ def stream_from_backup(
             private_password=archive_password,
             encryption_key=encryption_key,
         )
+        backup_set_id = self.get_backup_sets(
+            device_guid, archive_accessor.destination_guid
+        )[u"backupSets"][0][u"backupSetId"]
```
kiran-chaudhary, line 97: Need to understand, Why only the first `backupSetId`? is it the latest `backupSetId`?

src/py42/services/storage/archive.py
unparalleled-js: There is a test already that makes sure the URL is correct.

src/py42/clients/archive.py
unparalleled-js: It seems -1 works as backup set ID when you don't need to know it, so I switched to just use -1

tests/services/storage/test_archive.py
```diff
@@ -199,236 +203,48 @@ def test_create_restore_session_with_encryption_key_calls_post_with_encryption_k
...
-
-        storage_archive_service.start_restore(
-            DEVICE_GUID,
-            WEB_RESTORE_SESSION_ID,
-            PATH_SET,
-            NUM_FILES,
-            NUM_DIRS,
-            SIZE,
-            zip_result=True,
+        assert (
```
unparalleled-js, line 214: I deleted a lot of tests here because they were testing each parameter individually and we don't typically do that in our tests, we usually just do `tests_calls_url_with_expected_data` and it just tests all the params and the url in the same test... which is much easier to maintain.

@kiran-chaudhary here is where the URL gets tested for the updated call

unparalleled-js: View raw file

tests/services/storage/test_archive.py
```diff
@@ -10,6 +10,7 @@
 KWARGS_INDEX = 1
 
 WEB_RESTORE_SESSION_URL = "/api/WebRestoreSession"
+START_WEB_RESTORE_JOB_URL = "/api/v9/restore/web"
```
unparalleled-js, line 13: @kiran-chaudhary

src/py42/clients/archive.py
alanag13: do we know if this works as expected when we use v3 archives? Will v3 archives require that you specify backup set(s)? @timabrmsn ?

timabrmsn: I'm not sure, as I've never poked at v3 archive restore. Our internal Code42 clients should have a v3 destination, so we could test using one of our own accounts.

unparalleled-js: Because of Okta, I don't think we can use py42 for our own accounts.

kiran-chaudhary (approved):
Changes LGTM. 

- Not part of this task, as a suggestion, I would like to see some methods that list the files/paths that are available in archive, to test this feature I have to log in to `console.us.code42.com` find the paths and do trial and error. Or may be I am missing something in my understanding.  

Concern I am trying to address is Py42 cannot be used independently for this feature. 

- #201  's work has made some structural changes and hence it would be good if this gets merged and then work on #201 continues otherwise we will have to go through merging challenges.

unparalleled-js:
> Changes LGTM.
> 
> * Not part of this task, as a suggestion, I would like to see some methods that list the files/paths that are available in archive, to test this feature I have to log in to `console.us.code42.com` find the paths and do trial and error. Or may be I am missing something in my understanding.
> 
> Concern I am trying to address is Py42 cannot be used independently for this feature.
> 
> * #201  's work has made some structural changes and hence it would be good if this gets merged and then work on #201 continues otherwise we will have to go through merging challenges.

These are some fair points. We should consider exposing the methods for getting archive file metadata OR add this API to py42 https://east-client-2.us.code42.com/api/backupTreeNode ... It will literally return all of your files...

src/py42/clients/archive.py
alanag13: > Because of Okta, I don't think we can use py42 for our own accounts.

@unparalleled-js true, but we should be able to login, do a restore, and see in the console if the request behaves any differently for v3

As an aside Okta just ends up creating a v3_user_token, so with some breakpoint trickery or just testing the api without py42 you could still test this with an Okta-based account

alanag13:
> We should consider exposing the methods for getting archive file metadata OR add this API to py42 https://east-client-2.us.code42.com/api/backupTreeNode ... It will literally return all of your files...

These are very computationally expensive apis (you need to decrypt every file in the archive in order get all of its paths), and some members of PS have expressed that they would not want to encourage them to be used in automated fashion.

src/py42/clients/_archive_access.py
```diff
@@ -31,17 +33,25 @@ def get_archive_accessor(
             device_guid, destination_guid=destination_guid
         )
         decryption_keys = self._get_decryption_keys(
-            device_guid, private_password, encryption_key
+            device_guid=device_guid,
```
alanag13, line 36: is there a reason all these keyword args got added? These aren't optional params.

src/py42/clients/_archive_access.py
```diff
@@ -288,16 +302,16 @@ def _is_job_complete(self, job_id):
...
-        size = sum([fs.size for fs in file_selections])
-        zip_result = _check_for_multiple_files(file_selections) or None
+        num_bytes = sum([fs.num_bytes for fs in file_selections])
         return self._storage_archive_service.start_restore(
-            guid=self._device_guid,
+            device_guid=self._device_guid,
             web_restore_session_id=self._archive_session_id,
-            path_set=[fs.path_set for fs in file_selections],
+            restore_groups=[
+                {u"backupSetId": -1, u"files": [f.path_set for f in file_selections]}
```
alanag13, line 310: I checked my own account but it looks like all my archives are v1. Making sure this behaves as expected with a v3 archive is one thing i'd like to verify if possible @timabrmsn do you have a v3 archive or now how we could test one?

alanag13 (changes requested):
I'm a little confused about if there's a reason for all the added keyword args, and would like to verify the v3 behavior when we use `-1` for the backup set.

src/py42/clients/_archive_access.py
unparalleled-js: > I checked my own account but it looks like all my archives are v1. Making sure this behaves as expected with a v3 archive is one thing i'd like to verify if possible @timabrmsn do you have a v3 archive or now how we could test one?

I talked to Tim Friday, I don't this will work
for v3. We need to figure out if we need to support
restoring from
multiple backup sets or v3 at all. Also, question:  either way, right now we don't
support v3 archives, right? If so, these changes don't contribute to that lack necessarily. But now would
be a good time
to make that work or at least plan for it

unparalleled-js:
> > We should consider exposing the methods for getting archive file metadata OR add this API to py42 https://east-client-2.us.code42.com/api/backupTreeNode ... It will literally return all of your files...
> 
> 
> 
> These are very computationally expensive apis (you need to decrypt every file in the archive in order get all of its paths), and some members of PS have expressed that they would not want to encourage them to be used in automated fashion.
> 
> 
> 
> 
> 

Yeah, file events are a better use case anyway, at least for push restore where there is no PDS. Get the path from
the event and push the restore.

unparalleled-js:
> I'm a little confused about if there's a reason for all the added keyword args, and would like to verify the v3 behavior when we use `-1` for the backup set.

What is the behavior of v3 archives when using the v1 API where backupset ID is optional and 1-1 on restores?

src/py42/clients/_archive_access.py
unparalleled-js: Because there are so many of them and trying to maintain the order is challenging. It became painful to work with, so I made it like this to ease the pain...  Is there a rule that key value args can't be required params? This is all internal stuff.

unparalleled-js: I changed it back, but I still think positional args are harder to work with in these situations...

src/py42/clients/_archive_access.py
```diff
@@ -288,16 +291,14 @@ def _is_job_complete(self, job_id):
...
+        num_bytes = sum([fs.num_bytes for fs in file_selections])
         return self._storage_archive_service.start_restore(
-            guid=self._device_guid,
-            web_restore_session_id=self._archive_session_id,
-            path_set=[fs.path_set for fs in file_selections],
-            num_files=num_files,
-            num_dirs=num_dirs,
-            size=size,
-            zip_result=zip_result,
+            self._device_guid,
```
unparalleled-js, line 296: @alanag13 Removed some more arg specifications for you

unparalleled-js:
I am closing this since these changes are in the push restore branch and I don't want to keep fixing feedback in both.

src/py42/clients/archive.py
unparalleled-js: Ah that makes sense. Thankfully we found a testing environment with V3 archives in it so we don't have to do that

PR #200, opened 2020-09-17
Bugfix/handle locked settings

timabrmsn:
### Description of Change ###

Updates `DeviceSettings` class to handle cases where backup sets are locked.

Adds a `.locked` (read only) property on backup sets to indicate if they're locked and also prevents modifications to
non-destination settings on that set if `True`. 

### Issues Resolved ###
Fixes bug found in testing.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/settings/device_settings.py
+++ b/src/py42/clients/settings/device_settings.py
@@ -289,7 +289,7 @@ class BackupSet(UserDict, object):
         """Indicates whether the backup set as a whole is locked. If True, individual
         settings for this backup set (except for Destination settings), cannot be modified.
         """
-        return u"@locked" in self.data and self.data[u"@locked"] == u"true"
+        return u"@locked" in self.data and str_to_bool(self.data[u"@locked"])
 
     @property
     def included_files(self):
--- a/src/py42/clients/settings/device_settings.py
+++ b/src/py42/clients/settings/device_settings.py
@@ -29,12 +29,20 @@ class DeviceSettingsDefaults(UserDict, object):
         self.data[u"settings"] = {
             u"serviceBackupConfig": self.data[u"serviceBackupConfig"]
         }
-        self.backup_sets = [
-            BackupSet(self, self.changes, backup_set)
-            for backup_set in self.data[u"serviceBackupConfig"][u"backupConfig"][
-                u"backupSets"
-            ]
-        ]
+        bs = self.data[u"serviceBackupConfig"][u"backupConfig"][u"backupSets"]
+        self.backup_sets = self._extract_backup_sets(bs)
+
+    def _extract_backup_sets(self, backup_sets):
+        if isinstance(backup_sets, dict):  # number of sets are locked
+            backup_sets = backup_sets["backupSet"]
+            if isinstance(backup_sets, dict):  # there's only one set configured
+                return [BackupSet(self, backup_sets)]
+            if isinstance(backup_sets, list):
+                return [BackupSet(self, bs) for bs in backup_sets]
+            else:
+                raise Py42Error("Unable to extract backup sets: {}".format(backup_sets))
+        else:
+            return [BackupSet(self, bs) for bs in backup_sets]
 
     @property
     def available_destinations(self):
@@ -110,12 +118,10 @@ class DeviceSettings(DeviceSettingsDefaults):
         self.changes = {}
         self.data = device_dict
         self._destinations = device_dict[u"availableDestinations"]
-        self.backup_sets = [
-            BackupSet(self, self.changes, backup_set)
-            for backup_set in self.data[u"settings"][u"serviceBackupConfig"][
-                u"backupConfig"
-            ][u"backupSets"]
+        bs = self.data[u"settings"][u"serviceBackupConfig"][u"backupConfig"][
+            u"backupSets"
         ]
+        self.backup_sets = self._extract_backup_sets(bs)
         """List of :class:`BackupSet` objects used to manage this device's backup set configurations."""
 
     @property
@@ -168,9 +174,9 @@ class DeviceSettings(DeviceSettingsDefaults):
 class BackupSet(UserDict, object):
     """Helper class for managing device backup sets and Org device default backup sets."""
 
-    def __init__(self, settings_manager, changes_dict, backup_set_dict):
+    def __init__(self, settings_manager, backup_set_dict):
         self._manager = settings_manager
-        self._changes = changes_dict
+        self._changes = settings_manager.changes
         self.data = backup_set_dict
         includes, excludes = self._extract_file_selection_lists()
         regex_excludes = self._extract_regex_exclusions()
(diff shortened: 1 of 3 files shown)


src/py42/clients/settings/device_settings.py
```diff
@@ -168,9 +174,9 @@ def __repr__(self):
 class BackupSet(UserDict, object):
     """Helper class for managing device backup sets and Org device default backup sets."""
 
-    def __init__(self, settings_manager, changes_dict, backup_set_dict):
+    def __init__(self, settings_manager, backup_set_dict):
         self._manager = settings_manager
-        self._changes = changes_dict
+        self._changes = settings_manager.changes
```
timabrmsn, line 179: This change wasn't related to the bug, just made constructing the BackupSet object a bit simpler.

timabrmsn:
I have read the CLA Document and I hereby sign the CLA

src/py42/clients/settings/device_settings.py
```diff
@@ -29,12 +29,20 @@ def __init__(self, device_dict, org_settings):
...
-        ]
+        bs = self.data[u"serviceBackupConfig"][u"backupConfig"][u"backupSets"]
+        self.backup_sets = self._extract_backup_sets(bs)
+
+    def _extract_backup_sets(self, backup_sets):
+        if isinstance(backup_sets, dict):  # number of sets are locked
+            backup_sets = backup_sets["backupSet"]
+            if isinstance(backup_sets, dict):  # there's only one set configured
+                return [BackupSet(self, backup_sets)]
+            if isinstance(backup_sets, list):
```
unparalleled-js, line 40: This `if` should be an `elif`.

Because in case this implementation changes to not just be `returns` statements and because you are using `else`.

src/py42/clients/settings/device_settings.py
```diff
@@ -264,6 +283,13 @@ def _build_regex_exclusions(self):
         }
         self.data[u"backupPaths"][u"excludeUser"] = user_exclude_dict
 
+    @property
+    def locked(self):
+        """Indicates whether the backup set as a whole is locked. If True, individual
+        settings for this backup set (except for Destination settings), cannot be modified.
+        """
+        return u"@locked" in self.data
```
unparalleled-js, line 291: Is there never a '@locked': 'false'?

Would it be better to check for to make sure the value is actually `true` anyway?

src/py42/clients/settings/device_settings.py
```diff
@@ -414,42 +442,52 @@ def __str__(self):
...
-        self.manager._build_regex_exclusions()
+        self.backup_set._build_file_selection()
+        self.backup_set._build_regex_exclusions()
         if set(self.orig) != set(self.data):
             self._changes[self.name] = show_change(self.orig, self.data)
         elif self.name in self._changes:
             del self._changes[self.name]
 
     def append(self, item):
+        self.check_lock()
```
unparalleled-js, line 465: `lock_check` might make a good decorator

tests/clients/test_device_settings.py
```diff
@@ -827,3 +836,45 @@ def test_backup_set_filename_exclusions_append_produces_expected_values(
...
+        with pytest.raises(AttributeError):
+            device_settings.backup_sets[0].excluded_files = ["test"]
+
+        with pytest.raises(AttributeError):
+            device_settings.backup_sets[0].filename_exclusions.append("test")
+
+        with pytest.raises(AttributeError):
+            device_settings.backup_sets[0].filename_exclusions = ["test"]
+
+    def test_backup_set_when_set_locked_allows_destination_modifications(
```
unparalleled-js, line 868: Curious if this is supposed to be `test_backup_set_when_not_locked_allows_destination_modifications()`
(second `set` -> `not`).

tests/clients/test_device_settings.py
```diff
@@ -403,6 +403,15 @@ def device_settings_with_multiple_values():
     return device_settings_dict
 
 
+@pytest.fixture
+def device_settings_with_locked_backup_set():
+    device_settings_dict = deepcopy(DEVICE_DICT_W_SETTINGS)
+    device_settings_dict["settings"]["serviceBackupConfig"]["backupConfig"][
+        "backupSets"
```
unparalleled-js, line 410: Shouldn't this be changing the whole structure to be a `dict` as well rather than a list?

src/py42/clients/settings/device_settings.py
```diff
@@ -189,38 +195,51 @@ def _extract_file_selection_lists(self):
...
+            One+ exclusions:        `[{"windows": [], "pattern": [{"@regex": ".*1"}, {"@regex": ".*2"}],"linux": [],"macintosh": []}]
+            One+ exclusion locked:  `{'@locked': 'true', 'patternList': {'pattern': [{'@regex': '.*1'}, {'@regex': '.*2'}], 'windows': [], 'macintosh': [], 'linux': []}}`
         """
-        exclude_user = self.data[u"backupPaths"][u"excludeUser"][0]
-        pattern_list = exclude_user.get(u"pattern")
+        exclude_user = self.data[u"backupPaths"][u"excludeUser"]
+        if isinstance(exclude_user, dict):  # exclusions are locked
+            pattern_list = exclude_user[u"patternList"].get(u"pattern")
+        else:
+            pattern_list = exclude_user[0].get(u"pattern")
```
kiran-chaudhary, line 242: `No exclusions` sample above doesn't contain "pattern" key, so `pattern_list` would be empty for `No exclusions`, is it expected?

kiran-chaudhary (commented):
Not sure whether the below applies in this context, just for the sake of discussion and understanding the feature better.

How would this feature behave when multiple users try to update the configuration? I see we raise an exception and fail when a `lock` is found. How often settings would be updated? 

Do we really want to fail early? or can we push each task to a queue and kind of guarantee that each update would be successful, where every task waits for the lock to be released by earlier task for certain number of attempts and only then fail if it doesn't acquire the lock, and so on for all tasks in the queue.

Also I understand we need to keep `py42` simple but just a thought, it would have been better that the API would implement such logic and we didn't have to bother for such stuff.

src/py42/clients/settings/device_settings.py
unparalleled-js: I was assuming the locks are in place because they are set at the org level, or is that not it?

alanag13: > it would have been better that the API would implement such logic and we didn't have to bother for such stuff.

yes.

> Do we really want to fail early? or can we push each task to a queue and kind of guarantee that each update would be successful, where every task waits for the lock to be released by earlier task for certain number of attempts and only then fail if it doesn't acquire the lock, and so on for all tasks in the queue.

@timabrmsn  should probably confirm but my understanding is that this isn't a _concurrency_ lock (that is, its not a lock that only exists while someone else is making modifications but then later goes away). I think this is a _restrictive_ lock, i.e. an admin disabling the ability of other users to modify this setting.

timabrmsn: > pattern_list would be empty for No exclusions, is it expected?
yep, if there aren't any exclusion patterns we just want to return an empty list. 

And for locking, @alanag13 and @unparalleled-js are correct, it's an administrative restriction on the device settings, not a temporary lock because changes are in flight. 

And what makes it even more complex is that each individual item can be locked/unlocked, and the _whole backup set_ can be locked, which has the effect of overriding any individual lock settings. It's only in the case of the whole backup set being locked that we fail fast and just not let the attribute be changed, since it's possible to update those values in the json and post it in in the API, but if the whole set is locked the server will just ignore any changes you make, so we're calling attention to the fact that they'd _have_ to unlock the backup set first before making any changes.

alanag13 (approved):
@unparalleled-js has some small points / questions that I'm also curious about, but otherwise this LGTM.

tests/clients/test_device_settings.py
timabrmsn: No, this is correct, because destination usages _are_ still modifiable when the whole backup set is locked.

tests/clients/test_device_settings.py
timabrmsn: No, because here we're only locking a single backup set. The change from list > dict happens when you lock the _number_ of backup sets, preventing a user from creating any new sets.

src/py42/clients/settings/device_settings.py
timabrmsn: I like it. done

timabrmsn:
All feedback addressed.

src/py42/clients/settings/device_settings.py
```diff
@@ -264,6 +284,13 @@ def _build_regex_exclusions(self):
         }
         self.data[u"backupPaths"][u"excludeUser"] = user_exclude_dict
 
+    @property
+    def locked(self):
+        """Indicates whether the backup set as a whole is locked. If True, individual
+        settings for this backup set (except for Destination settings), cannot be modified.
+        """
+        return u"@locked" in self.data and self.data[u"@locked"] == u"true"
```
unparalleled-js, line 292: To be safe, it might be better to call your `str_to_bool()` method here and make sure the result of that call is `True`.

unparalleled-js: `and str_to_bool(self.data[u"@locked"])`

unparalleled-js (approved):
Approving, but consider using `str_to_bool()` to check if the value is "true" for the lock.

PR #201, opened 2020-09-17
Feature/pushrestore

unparalleled-js:
### Description of Change ###

Support for push restore: `sdk.archive.stream_to_device()`.
Switch to the v9 API for web restore to mirror the console UI.

### Issues Resolved ###
N/a

### Testing Procedure ###
Make sure you can perform both a push restore and that web restore still works.

`sdk.archive.stream_to_device(["file/to/restore"], "device_guid", "accepting_device_guid", restore_path="C:/")`

### PR Checklist ###

Tests are coming.


unparalleled-js:
Ugh a lot is going on to make this work.  Basically can either create two restore session and use one for creating the file selections and the other for doing the restore, or you can ignore file sizes and rely on the user inputting the correct path and then you only need to create the push restore session. This made it make most sense to break up `ArchiveAccessor` into several use-cases. Also, push restore requires the master URL, so that is the reason for the connection changes.

src/py42/clients/_archiveaccess/restoremanager.py
```diff
@@ -0,0 +1,171 @@
...
+        return self._storage_archive_service.start_restore(
+            device_guid=self._device_guid,
+            web_restore_session_id=self._archive_session_id,
+            restore_groups=[
+                {u"backupSetId": -1, u"files": [f.file for f in file_selections]}
+            ],
+            num_files=num_files,
+            num_dirs=num_dirs,
+            num_bytes=num_bytes,
+            show_deleted=True,
```
timabrmsn, line 156: "show deleted" should default to False, and should also be a configurable parameter. Most people _don't_ want all their deleted files when restoring.

unparalleled-js: `show_deleted` is and has been `True` for web restore in py42, which is why I made it True here.  @alanag13  is there a reason it's True there that you know of?

I like to keep in consistent with the default values and not have that be a breaking change, but we definitely should make it configurable.

alanag13: it's probably just true because that's what it does in the ui (or did at the time). We can make it default to false for 2.0 but changing the default would be breaking. I agree that it should be configurable though, and we could do that now if we wanted to.

alanag13: we can definitely have false be what passed in when your doing a push restore rather than a web, though.

src/py42/services/_connection.py
```diff
@@ -76,6 +76,19 @@ def get_host_address(self):
...
 
+class ConnectedServerHostResolver(HostResolver):
+    def __init__(self, connection, device_guid):
+        self._connection = connection
+        self._device_guid = device_guid
+        super().__init__()
+
+    def get_host_address(self):
+        response = self._connection.get(
+            u"api/connectedServerUrl", params={u"guid": self._device_guid}
```
alanag13, line 87: I don't recall needing anything like this the first time I attempted doing push restore via the API. Has something changed? I do remember that:
- there were some endpoints that I thought should be called on the storage node that were actually called on the authority instead
- it was important that some of the api calls contain `/console/` in the url in order to force the load balancer to have us stick with a given server.

unparalleled-js: The only way I was able
to get the API's to work
was this way or to use some cookie in the browser with c42_auth, I think something did change.

src/py42/clients/_archiveaccess/restoremanager.py
unparalleled-js: If it's what the browser does, then why is it unexpected for it to be default?

The only time I have ever restored is because I accidentally deleted something, not sure if that is the same
thing though

timabrmsn: The UI default has always been to _not_ show deleted files. And while there's definitely a use for showing deleted files, you more often than not don't want everything you've ever deleted to come back when restoring a folder or a whole system. It can get quite messy.

src/py42/services/_connection.py
timabrmsn: This is a push-restore-only thing. Since the client you are pushing to has to be online to receive the restore request, we have to find out if it's actively online, which is what `connectedServerUrl` does. The UI does this behind the scenes when you put in the target guid and doesn't let you start the restore if the client is offline.

src/py42/clients/_archiveaccess/restoremanager.py
unparalleled-js: You are correct, the default is supposed to be False. Also, it's false in the code: https://stash.corp.code42.com/projects/SVR/repos/core/browse/pro_core/mod_backup/src/api/com/code42/backup/webrestore/getcontents/CreateGetContentsJobCmd.java#31,37,58,77,85-86

unparalleled-js: I also appreciate your product knowledge @timabrmsn , thanks

unparalleled-js: I will for sure it in push restore, but if it is really kind of messy like you say, could we pass this off as a bugfix like, 

`Fix bug in web restore where we accidentally defaulted to restoring deleted files in a directory.`

src/py42/services/_connection.py
alanag13: > The only way I was able
> to get the API's to work
> was this way or to use some cookie in the browser with c42_auth, I think something did change.

This read like a poem, beautiful stuff

alanag13: > This is a push-restore-only thing. Since the client you are pushing to has to be online to receive the restore request, we have to find out if it's actively online, which is what `connectedServerUrl` does. The UI does this behind the scenes when you put in the target guid and doesn't let you start the restore if the client is offline.

Cool, makes sense to me

src/py42/clients/archive.py
```diff
@@ -86,16 +93,102 @@ def stream_from_backup(
...
+            show_deleted,
+        )
+
+    def _select_backup_set_id(self, device_guid, destination_guid):
+        backup_sets = self.get_backup_sets(device_guid, destination_guid)[u"backupSets"]
+        if not backup_sets:
+            return None
+        for backup_set in backup_sets:
+            backup_set_id = backup_set[u"backupSetId"]
+            # 1 is the default backup set ID and is preferred.
```
unparalleled-js, line 186: @timabrmsn @alanag13 Looks like the UI prefers the backupset of ID 1 when it exists.

unparalleled-js:
I've got this working for V3 archives now and have tested it.

unparalleled-js:
Keep in mind that you can speed up restores but setting `file_size_calc_timeout=0`, which will skip calculating file size

src/py42/clients/_archiveaccess/__init__.py
```diff
@@ -0,0 +1,165 @@
+import posixpath
```
unparalleled-js, line 1: I know a lot of these files are all green from the refactor, but they could use an overall review anyway, so I don't mind.

Reviewing this PR may take some time...

phirst:
Is it correct to assume `file/to/restore` is a single file or a list of files but not a path?

unparalleled-js:
> Is it correct to assume `file/to/restore` is a single file or a list of files but not a path?

It is a list of string file paths or a single file path. It can also be a path to directories.

src/py42/services/_connection.py
```diff
@@ -76,6 +77,24 @@ def get_host_address(self):
...
+    def __init__(self, connection, device_guid):
+        self._connection = connection
+        self._device_guid = device_guid
+        super(ConnectedServerHostResolver, self).__init__()
+
+    def get_host_address(self):
+        response = self._connection.get(
+            u"api/connectedServerUrl", params={u"guid": self._device_guid}
+        )
+        if response[u"serverUrl"] is None:
```
unparalleled-js, line 93: Response when offine is like `{"serverUrl": null}` which causes the logic for checking if the host address begins with the `http` or `https` to fail with `None has no method startsWith()` which is confusing, so this custom exception is quite helpful

src/py42/clients/_archiveaccess/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+
+FileSelection = namedtuple(u"FileSelection", u"file, num_files, num_dirs, num_bytes")
+
+
+class FileType(object):
+    DIRECTORY = u"DIRECTORY"
+    FILE = u"FILE"
+
+
+class ArchiveAccessor(object):
```
timabrmsn, line 15: Can we get docstrings on all these classes to make it easier to see how they all fit together?

src/py42/services/storage/restore.py
```diff
@@ -0,0 +1,63 @@
...
+        See https://console.us.code42.com/apidocviewer/#WebRestoreSession
+        """
+        uri = u"/api/WebRestoreSession"
+        json_dict = {
+            u"computerGuid": device_guid,
+            u"dataKeyToken": data_key_token,
+            u"privatePassword": private_password,
+            u"encryptionKey": encryption_key,
+        }
+        return self._connection.post(uri, json=json_dict)
```
timabrmsn, line 22: I noticed while testing if an archive has a private_password and you pass the wrong one here, py42 raises a generic 500, but the error text is clear that the password is incorrect. Perhaps we could create a custom exception for these?

timabrmsn: Example exception w/ response text so you don't have to do the work to recreate it: 

```
>>> r
Py42InternalServerError('Failure in HTTP call 500 Server Error: Server Error for url: https://web-epf-sea.crashplan.com/api/WebRestoreSession')
>>> r.response.text
'[{"name":"PRIVATE_PASSWORD_INVALID","description":"An error has occurred. See server logs for more information.","objects":[]}]'
```

timabrmsn: And here's for if a custom encryption key doesn't match: 

```
>>> r.response.text
'[{"name":"CUSTOM_KEY_INVALID","description":"An error has occurred. See server logs for more information.","objects":[]}]'
```

unparalleled-js: Done

timabrmsn (approved):
Looks good to me!

unparalleled-js:
GitHub actions are currently down: https://www.githubstatus.com/

PR #205, opened 2020-09-23
Refactor/timestamp formats

kiran-chaudhary:
### Description of Change ###

Change `timestamp` fields to accept timestamp in string, integer or datetime instance.

- closes #
 #INTEG-304

### Testing Procedure ###
`sdk.auditlogs.get_page(begin_time=100000000)`

`sdk.auditlogs.get_page(begin_time="2020-02-02 10:10:10")`

```
from datetime import datetime
dt = datetime(2020, 2, 2, 10, 10, 10)
sdk.auditlogs.get_page(end_time=dt)

```
Similarly test for 
```
sdk.auditlogs.get_all
sdk.securitydata.get_all_plan_security_events
sdk.securitydata.get_all_user_security_events
```
### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.auditlogs.get_all()` arguments `begin_time` and `end_time`
     - `sdk.securitydata.get_all_plan_security_events()` arguments `min_timestamp` and `max_timestamp`
     - `sdk.securitydata.get_all_user_security_events()` arguments `min_timestamp` and `max_timestamp`
+    - `sdk.detectionlists.departing_employee.add()` argument `departure_date`
+    - `sdk.detectionlists.departing_employee.update_departure_date()` argument `departure_date`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- Timestamp argument accepts input in string type( format `YYYY-MM-DD HH:MM:SS` ), int type(epoch time) as well as a datetime instance for following:
-    - `sdk.auditlogs.get_page`
-    - `sdk.auditlogs.get_all`
-    - `sdk.securitydata.get_all_plan_security_events`
-    - `sdk.securitydata.get_all_user_security_events`
+- Timestamp argument accepts input in string type( format `yyyy-MM-dd HH:MM:SS` ), int or float type( epoch time ) as well as a datetime instance for following:
+    - `sdk.auditlogs.get_page()` arguments `begin_time` and `end_time`
+    - `sdk.auditlogs.get_all()` arguments `begin_time` and `end_time`
+    - `sdk.securitydata.get_all_plan_security_events()` arguments `min_timestamp` and `max_timestamp`
+    - `sdk.securitydata.get_all_user_security_events()` arguments `min_timestamp` and `max_timestamp`
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Changed
+
+- Timestamp argument accepts input in string type( format `YYYY-MM-DD HH:MM:SS` ), int type(epoch time) as well as a datetime instance for following:
+    - `sdk.auditlogs.get_page`
+    - `sdk.auditlogs.get_all`
+    - `sdk.securitydata.get_all_plan_security_events`
+    - `sdk.securitydata.get_all_user_security_events`
+
 ### Removed
 
 - Removed faulty `within_the_last()` method from `sdk.queries.alerts.filters.alert_filter.DateObserved`.
--- a/src/py42/clients/auditlogs.py
+++ b/src/py42/clients/auditlogs.py
@@ -26,10 +26,10 @@ class AuditLogsClient(object):
         Args:
             page_num (int, optional): The page number to get. Defaults to 1.
             page_size (int, optional): The number of items per page. Defaults to `py42.settings.items_per_page`.
-            begin_time (datetime, optional): A datetime instance representing the beginning of the
-                date range of audit logs. Defaults to None.
-            end_time (datetime, optional): A datetime instance representing the beginning of the
-                date range of audit logs. Defaults to None.
+            begin_time (int or str or datetime, optional): Timestamp in milliseconds or
+                str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
+            end_time (int or str or datetime, optional): Timestamp in milliseconds or
+                str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
             event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
             user_ids (str or list, optional): A str or list of str of Code42 userUids. Defaults to None.
             usernames (str or list, optional): A str or list of str of Code42 usernames. Defaults to None.
@@ -70,9 +70,9 @@ class AuditLogsClient(object):
 
         Args:
             begin_time (int or str or datetime, optional): Timestamp in milliseconds or
-            str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
-            end_time (int or str or datetime, optional): Timestamp in millisecondsor
-            str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
+                str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
+            end_time (int or str or datetime, optional): Timestamp in milliseconds or
+                str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
             event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
             user_ids (str or list, optional): A str or list of str of Code42 userUids. Defaults to None.
             usernames (str or list, optional): A str or list of str of Code42 usernames. Defaults to None.
--- a/src/py42/clients/auditlogs.py
+++ b/src/py42/clients/auditlogs.py
@@ -67,8 +67,10 @@ class AuditLogsClient(object):
         """Retrieve audit logs, filtered based on given arguments.
 
         Args:
-            begin_time (int, optional): Timestamp in milliseconds. Defaults to None.
-            end_time (int, optional): Timestamp in milliseconds. Defaults to None.
+            begin_time (int or str or datetime, optional): Timestamp in milliseconds or
+            str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
+            end_time (int or str or datetime, optional): Timestamp in millisecondsor
+            str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
             event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
             user_ids (str or list, optional): A str or list of str of Code42 userUids. Defaults to None.
             usernames (str or list, optional): A str or list of str of Code42 usernames. Defaults to None.
--- a/src/py42/clients/auditlogs.py
+++ b/src/py42/clients/auditlogs.py
@@ -26,9 +26,9 @@ class AuditLogsClient(object):
         Args:
             page_num (int, optional): The page number to get. Defaults to 1.
             page_size (int, optional): The number of items per page. Defaults to `py42.settings.items_per_page`.
-            begin_time (int or str or datetime, optional): Timestamp in milliseconds or
+            begin_time (int or float or str or datetime, optional): Timestamp in milliseconds or
                 str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
-            end_time (int or str or datetime, optional): Timestamp in milliseconds or
+            end_time (int or float or str or datetime, optional): Timestamp in milliseconds or
                 str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
             event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
             user_ids (str or list, optional): A str or list of str of Code42 userUids. Defaults to None.
@@ -69,9 +69,9 @@ class AuditLogsClient(object):
         """Retrieve audit logs, filtered based on given arguments.
 
         Args:
-            begin_time (int or str or datetime, optional): Timestamp in milliseconds or
+            begin_time (int or float or str or datetime, optional): Timestamp in milliseconds or
                 str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
-            end_time (int or str or datetime, optional): Timestamp in milliseconds or
+            end_time (int or float or str or datetime, optional): Timestamp in milliseconds or
                 str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
             event_types (str or list, optional): A str or list of str of valid event types. Defaults to None.
             user_ids (str or list, optional): A str or list of str of Code42 userUids. Defaults to None.
--- a/src/py42/clients/auditlogs.py
+++ b/src/py42/clients/auditlogs.py
@@ -26,8 +26,10 @@ class AuditLogsClient(object):
         Args:
             page_num (int, optional): The page number to get. Defaults to 1.
             page_size (int, optional): The number of items per page. Defaults to `py42.settings.items_per_page`.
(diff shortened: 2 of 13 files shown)


kiran-chaudhary:
Observed 4 methods which were exposed to user, hence changed only those ones.

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+### Changed
+
+- Timestamp argument accepts input in string type( format `YYYY-MM-DD HH:MM:SS` ), int type(epoch time) as well as a datetime instance for following:
```
unparalleled-js, line 15: I believe it is `yyyy-M-dd` for the year part

unparalleled-js, line 15: Spacing is weird be the word `type` and the following parans

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+    - `sdk.auditlogs.get_page`
```
unparalleled-js, line 16: Probably should specify which args this affect for each method

CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+    - `sdk.auditlogs.get_all`
```
unparalleled-js, line 17: It is typical for us to end methods with `()`, so like `sdk.securitydata.get_all_plan_security_events()` etc.

src/py42/clients/auditlogs.py
```diff
@@ -26,8 +26,10 @@ def get_page(
         Args:
             page_num (int, optional): The page number to get. Defaults to 1.
             page_size (int, optional): The number of items per page. Defaults to `py42.settings.items_per_page`.
-            begin_time (int, optional): Timestamp in milliseconds. Defaults to None.
-            end_time (int, optional): Timestamp in milliseconds. Defaults to None.
+            begin_time (int or str or datetime, optional): Timestamp in milliseconds or
+                str format "YYYY-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
```
unparalleled-js, line 30: yyyy-MM-dd

src/py42/clients/securitydata.py
```diff
@@ -102,10 +102,10 @@ def get_all_plan_security_events(
                         - ``PERSONAL_CLOUD_SCAN_RESULT``
 
                     Defaults to None.
-            min_timestamp (float, optional): A POSIX timestamp representing the beginning of the
-                date range of events to get. Defaults to None.
-            max_timestamp (float, optional): A POSIX timestamp representing the end of the date
-                range of events to get. Defaults to None.
+            min_timestamp (int or str or datetime, optional): Timestamp in milliseconds or
```
unparalleled-js, line 105: Looks like you removed `float` here, we should keep float since we supported it before, and I think one of our util methods returns a float.. You can keep `int` as well, .... so it should be `float or int or str or datetime`

src/py42/util.py
```diff
@@ -96,3 +98,29 @@ def to_list(value):
     if not isinstance(value, (list, tuple)):
         return [value]
     return value
+
+
+def parse_timestamp_to_milliseconds_precision(timestamp):
```
unparalleled-js, line 103: These would fail if they get floats

unparalleled-js:
What about the file event and alert filters?  For example `EventTimestamp`.  Will that work with the new formats?  Are there tests for this?

unparalleled-js:
Looks like you forgot `securitydata.get_all_plan_security_events()`

unparalleled-js:
I think we should support the other formats for `departing_employee.add()` `departure_date` param, but keep in mind that the string format only really cares about the month, day, and year (not the time).

kiran-chaudhary:
> Looks like you forgot `securitydata.get_all_plan_security_events()`

both methods internally call `_get_security_detection_events` where the changes are done, hence its only in single place.

kiran-chaudhary:
> What about the file event and alert filters? For example `EventTimestamp`. Will that work with the new formats? Are there tests for this?

Event Filters `EventTimestamp` takes fixed set of values, e.g `FIFTEEN_MINUTES = u"PT15M"` , Do we really want to change query filters?

unparalleled-js:
> > Looks like you forgot `securitydata.get_all_plan_security_events()`
> 
> both methods internally call `_get_security_detection_events` where the changes are done, hence its only in single place.

Ah ok. Well, in any case, it should be reflected in the CHANGELOG.  Only 2/3 of them are in the CHANGELOG.

unparalleled-js:
> > Looks like you forgot `securitydata.get_all_plan_security_events()`
> 
> both methods internally call `_get_security_detection_events` where the changes are done, hence its only in single place.

My bad, I was mistaken, I thought there were 3 methods there for some reason

unparalleled-js:
> > What about the file event and alert filters? For example `EventTimestamp`. Will that work with the new formats? Are there tests for this?
> 
> Event Filters `EventTimestamp` takes fixed set of values, e.g `FIFTEEN_MINUTES = u"PT15M"` , Do we really want to change query filters?

Hmm... I can see `DateTime` being of use for the filters.

kiran-chaudhary:
> I think we should support the other formats for `departing_employee.add()` `departure_date` param, but keep in mind that the string format only really cares about the month, day, and year (not the time).

Need clarification here. 
imo, the timestamp field which represents time in `float/int` format is not a trivial/intuitive/natural type for a user to fill in, if they know the date they have to calculate the float to get the work done, so we are providing a support for natural types where it isn't.   

Do we want to change the user interfaces which are already natural/trivial for a user to fill in? Even if we go ahead with it I guess supporting `datetime` would still be fine but supporting `float/int` might not be needed. Since for a departing employee one would interested only in the date and its use case is not similar to file events etc where time precision is important!!

unparalleled-js:
> > I think we should support the other formats for `departing_employee.add()` `departure_date` param, but keep in mind that the string format only really cares about the month, day, and year (not the time).
> 
> Need clarification here.
> imo, the timestamp field which represents time in `float/int` format is not a trivial/intuitive/natural type for a user to fill in, if they know the date they have to calculate the float to get the work done, so we are providing a support for natural types where it isn't.
> 
> Do we want to change the user interfaces which are already natural/trivial for a user to fill in? Even if we go ahead with it I guess supporting `datetime` would still be fine but supporting `float/int` might not be needed. Since for a departing employee one would interested only in the date and its use case is not similar to file events etc where time precision is important!!

Talked offline, but no we don't want to remove the constants already set in place, we want to add additional support for other date / time formats.

It is not always the case that the user merely fills in these parameters, but rather these parameters can come from other raw-data, in which case the float timestamp fields are more likely.

I think the big one, at least that we all agree on, is the `DateTime` object support.  Everything that accepts a date in our library should support this object, as it is normal to work with in Python.

unparalleled-js:
I am not sure if this matters but when I use a `datetime` object, I get different date-output formats for both parseing methods:

```
2020-09-24 13:17:06.160837
2020-09-24T13:17:06.160837Z
```

Notice the microseconds result has T. Also notice the millisecond result does _not_ have Z and is actually microseconds.

src/py42/util.py
```diff
@@ -96,3 +98,29 @@ def to_list(value):
...
+    if isinstance(timestamp, datetime):
+        return convert_datetime_to_timestamp_str(timestamp)
+
+    if isinstance(timestamp, str):
+        return convert_datetime_to_timestamp_str(
+            datetime.strptime(timestamp, u"%Y-%m-%d %H:%M:%S")
+        )
+
+
+def parse_timestamp_to_microseconds_precision(timestamp):
```
unparalleled-js, line 116: Copy some of the helper method calls from the method above to get this to work correctly.  See https://github.com/code42/py42/pull/205#issuecomment-698339164

CHANGELOG.md
kiran-chaudhary: Thank you, I thought it was same until now.

kiran-chaudhary: Changed...

CHANGELOG.md
kiran-chaudhary: Done.

CHANGELOG.md

src/py42/clients/auditlogs.py
kiran-chaudhary: Changed.

src/py42/util.py
```diff
@@ -96,3 +98,29 @@ def to_list(value):
     if not isinstance(value, (list, tuple)):
         return [value]
     return value
+
+
+def parse_timestamp_to_milliseconds_precision(timestamp):
+    if isinstance(timestamp, int) or isinstance(timestamp, float):
+        return convert_timestamp_to_str(timestamp)
```
unparalleled-js, line 105: In `convert_timestamp_to_str()`, could yo uas `(float or int)` so that it knows it can accept an `int` ?

src/py42/clients/securitydata.py
kiran-chaudhary: Added.

src/py42/util.py
kiran-chaudhary: Refactored.

src/py42/util.py
kiran-chaudhary: addressed.

src/py42/services/detectionlists/departing_employee.py
```diff
@@ -6,6 +8,8 @@
 from py42.services.util import get_all_pages
 from py42.util import get_attribute_keys_from_class
 
+DATE_FORMAT = "%Y-%m-%d"
```
unparalleled-js, line 11: This could be internal `_DATE_FORMAT`

src/py42/services/detectionlists/departing_employee.py
```diff
@@ -40,11 +44,13 @@ def add(self, user_id, departure_date=None):
         Args:
             user_id (str or int): The Code42 userUid of the user you want to add to the departing \
                 employees list.
-            departure_date (str, optional): Date in yyyy-MM-dd format. Date is treated as UTC. Defaults to None.
+            departure_date (str or datetime instance, optional): Date in yyyy-MM-dd format. Date is treated as UTC. Defaults to None.
```
unparalleled-js, line 47: Should mention in the description that it can be a `datetime` instance too:

`Date in yyyy-MM-dd format or instance of `datetime`. Date is treated as UTC. Defaults to None.`

unparalleled-js, line 47: Don't need the word `instance` in the parenthesis .  Should just be `(str or datetime, optional)`

src/py42/services/detectionlists/departing_employee.py
```diff
@@ -178,14 +184,15 @@ def update_departure_date(self, user_id, departure_date):
 
         Args:
             user_id (str): The Code42 userUid of the user.
-            departure_date (date): Date in yyyy-MM-dd format. Date is treated as UTC.
+            departure_date (str or datetime instance): Date in yyyy-MM-dd format. Date is treated as UTC.
```
unparalleled-js, line 187: Don't need instance in the arg type part.  Should be `(str or datetime)`.  In the description, though, mention it can also be an instance of a `datetime` object.

tests/test_util.py
```diff
@@ -31,3 +31,60 @@ def test_to_list():
     assert util.to_list("") == []
     assert util.to_list(["a"]) == ["a"]
     assert util.to_list("a") == ["a"]
+
```
unparalleled-js, line 34: None of these tests have any milliseconds or microseconds.
For floats, just add a decimal, like `1599653541.123` or `1599653541.123456`

kiran-chaudhary: The last two tests here covers float, added more decimal values.

src/py42/services/detectionlists/departing_employee.py
kiran-chaudhary: Changed.

src/py42/services/detectionlists/departing_employee.py
kiran-chaudhary: Changed.

kiran-chaudhary: Changed.

src/py42/services/detectionlists/departing_employee.py
kiran-chaudhary: Changed.

src/py42/util.py
kiran-chaudhary: Changed docstring.

tests/sdk/queries/test_query_filter.py
```diff
@@ -414,3 +419,112 @@ def test_filter_group_when_changed_filter_clause_has_correct_json_representation
         '{"operator":"IS", "term":"term", "value":"value2"},'
         '{"operator":"IS", "term":"term", "value":"value3"}]}'
     )
+
+
+class TestQueryFilterTimestampField:
```
kiran-chaudhary, line 424: There were no tests for class `QueryFilterTimestampField` so had to add new ones.

src/py42/sdk/queries/query_filter.py
```diff
@@ -307,11 +309,15 @@ def on_same_day(cls, value):
...
         Args:
-            value (str or int or float): The value used to filter results.
+            value (str or int or float or datetime): The value used to filter results.
 
         Returns:
             :class:`~py42.sdk.queries.query_filter.FilterGroup`
         """
+        if isinstance(value, str):
+            value = convert_datetime_to_epoch(datetime.strptime(value, DATE_STR_FORMAT))
+        if isinstance(value, datetime):
```
unparalleled-js, line 319: Should be `elif`

src/py42/util.py
```diff
@@ -3,6 +3,11 @@
 import json
 from datetime import datetime
 
+from py42._compat import str
+
+MICROSECOND_FORMAT = u"%Y-%m-%dT%H:%M:%S.%fZ"
```
unparalleled-js, line 8: These could be internal

src/py42/util.py
```diff
@@ -96,3 +101,29 @@ def to_list(value):
     if not isinstance(value, (list, tuple)):
         return [value]
     return value
+
+
+def parse_timestamp_to_milliseconds_precision(timestamp):
+    if isinstance(timestamp, int) or isinstance(timestamp, float):
+        return convert_timestamp_to_str(timestamp)
+
+    if isinstance(timestamp, datetime):
```
unparalleled-js, line 110: These should use `elif`

src/py42/util.py
```diff
@@ -96,3 +101,29 @@ def to_list(value):
...
+        return convert_datetime_to_timestamp_str(
+            datetime.strptime(timestamp, DATE_STR_FORMAT)
+        )
+
+
+def parse_timestamp_to_microseconds_precision(timestamp):
+    if isinstance(timestamp, int) or isinstance(timestamp, float):
+        return datetime.utcfromtimestamp(timestamp).strftime(MICROSECOND_FORMAT)
+
+    if isinstance(timestamp, datetime):
```
unparalleled-js, line 123: These should use `elif`

tests/sdk/queries/test_query_filter.py
```diff
@@ -414,3 +419,112 @@ def test_filter_group_when_changed_filter_clause_has_correct_json_representation
...
+                {
+                    "operator": "ON_OR_BEFORE",
+                    "term": "override_timestamp_field_name",
+                    "value": "2020-09-10T11:12:13.000Z",
+                }
+            ],
+        }
+        assert dict(QueryFilterTimestampField.on_or_before(timestamp)) == filter_query
+
+    @pytest.mark.parametrize(
```
unparalleled-js, line 471: These are good - one other test we could also do is where there are digits after the decimal (not just `.0`) and make sure the timestamp works (doesn't end in `000Z`).

unparalleled-js (approved):
Looks ok. Please consider using `elif` in the spots I called out

src/py42/sdk/queries/query_filter.py
kiran-chaudhary: changed.

src/py42/util.py
kiran-chaudhary: When we have `return` statements does having elif matter?

unparalleled-js: It doesn't matter as much, but if the implementation ever chanages, we may forgot to switch these to use `elif` instead of `if`.. Also, for readability, it improves.

So, 1.) is more maintainable

and 2.). is more readable.

But, it is not that big of a deal, which I why I approved already.

tests/sdk/queries/test_query_filter.py
kiran-chaudhary: Added tests.

src/py42/util.py
kiran-chaudhary: Changed.

src/py42/util.py
kiran-chaudhary: Changed.

PR #206, opened 2020-09-24
Implementing Py42ActiveLegalHoldError

### LINKED ISSUE #169: [Enhancement] Py42UnableToDeactivateError
## Summary
A custom exception for when trying to block a user who is on legalhold

## Proposed API

The `py42.users.deactivate()` and `py42.devices.deactivate()` APIs

## Intended Use Case
Call `users.deactivate()` on a user who is a legal hold custodian should raise a specific error reflected that state.

DiscoRiver:
### Description of Change ###

Added `Py42ActiveLegalHoldError` exception. Raised when attempting to deactivate user/device that is in an active legal hold. 

### Issues Resolved ###

- Fixes #169 

### Testing Procedure ###

1. Use `py42.sdk.devices.deactivate()` or `py42.sdk.users.deactivate()` to deactivate a user in an active legal hold.
2. Observe message `Cannot deactivate from user in an active legal hold.`

### Notes ###

- Create draft PR as I suspect messaging for this exception will need to be updated, and I want to clarify the approach I've taken with this is acceptable before it's reviewed proper.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
+- Added `Py42ActiveLegalHoldError` exception when attempting to deactivate a user or device in an active legal hold.
+
 - Added additional user-adjustable setting for security events page size:
     - `py42.settings.security_events_per_page`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 ### Added
 
 - Added `Py42ActiveLegalHoldError` exception when attempting to deactivate a user or device in an active legal hold.
+    - `py42.sdk.users.deactivate()`
+    - `py42.sdk.devices.deactivate()`
 
 - Added additional user-adjustable setting for security events page size:
     - `py42.settings.security_events_per_page`
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -1,5 +1,6 @@
-from py42._compat import str
 import json
+
+from py42._compat import str
 
 
 class Py42Error(Exception):
@@ -165,7 +166,7 @@ def raise_py42_error(raised_error):
     if raised_error.response.status_code == 400:
         if json.loads(raised_error.response.text)[0]["name"] in [
             "USER_IN_ACTIVE_LEGAL_HOLD",
-            "ACTIVE_LEGAL_HOLD"
+            "ACTIVE_LEGAL_HOLD",
         ]:
             raise Py42ActiveLegalHoldError(raised_error)
         else:
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -113,8 +113,10 @@ class Py42ActiveLegalHoldError(Py42BadRequestError):
     """An exception raised when attempting to deactivate a user or device that is in an
     active legal hold."""
 
-    def __init__(self, exception):
-        msg = u"Cannot deactivate from user in an active legal hold."
+    def __init__(self, exception, resource, id):
+        msg = u"Cannot deactivate the {0} with ID {1} as the {0} is involved in a legal hold matter.".format(
+            resource, id,
+        )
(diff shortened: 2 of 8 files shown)


src/py42/exceptions.py
```diff
@@ -107,6 +109,15 @@ class Py42InternalServerError(Py42HTTPError):
     """A wrapper to represent an HTTP 500 error."""
 
 
+class Py42ActiveLegalHoldError(Py42BadRequestError):
+    """An exception raised when attempting to deactivate a user or device that is in an
+    active legal hold."""
+
+    def __init__(self, exception):
+        msg = u"Cannot deactivate from user in an active legal hold."
```
DiscoRiver, line 117: I suspect we may want to update this message.

src/py42/exceptions.py
```diff
@@ -153,7 +164,13 @@ def raise_py42_error(raised_error):
     HTTPError's response status code.
     """
     if raised_error.response.status_code == 400:
-        raise Py42BadRequestError(raised_error)
+        if json.loads(raised_error.response.text)[0]["name"] in [
```
DiscoRiver, line 167: Since the response text comes back as a string, and the code is simply `400`, I didn't see a better approach than this. We want to capture it before a `Py42BadRequestError` is raised.

src/py42/exceptions.py
unparalleled-js: Could add a `resource` parameter that is either `user` or `device`, along with an `uid` parameter that is either the `device_guid` or the `userUid` and the message could be like `u"Cannot deactivate the {0} with UID {1} as the {0} is involed in a legal hold matter.".format(resource, uid)`.

src/py42/exceptions.py
```diff
@@ -153,7 +164,13 @@ def raise_py42_error(raised_error):
...
+            "USER_IN_ACTIVE_LEGAL_HOLD",
+            "ACTIVE_LEGAL_HOLD",
```
unparalleled-js, line 169: Could just `json.loads(raised_error.response.text)[0][u"name"].contains(u"ACTIVE_LEGAL_HOLD")` to encompass both

src/py42/exceptions.py
unparalleled-js: Could we not just do what we are doing in `py42.services.detectionlists.__init__.py` merthod: `handle_user_already_added_errpr` ?

I may not understand the problem, but see that method if you haven't already.  It checks the bad request error (a py42 error) for the error text.

DiscoRiver: That's a much better way of doing it! I've updated based on this design.

src/py42/exceptions.py
DiscoRiver: I've moved all of this out of `py42.exceptions` based on your other comments. Looking back, it added overhead to any `400` response, which wasn't ideal.

DiscoRiver:
I changed this from a draft because I think it's pretty much complete.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
+- Added `Py42ActiveLegalHoldError` exception when attempting to deactivate a user or device in an active legal hold.
```
unparalleled-js, line 19: Not a big deal here because it is obvious, but for SDK type projects, it is better to mention the exact methods being affected. In this case it would be `sdk.users.deactivate()` and `sdk.devices.deactivate()`.  Just 1 less mental connection we have to put on CHANGELOG consumers.

unparalleled-js (approved):

unparalleled-js (changes requested):
I approved too soon! We need tests :)

Both `users.deactivate()` and `devices.deactivate()` should be tested to make sure that when the connection raises x error containing y text, than it raises this new error ` Py42ActiveLegalHoldError `

DiscoRiver:
> I approved too soon! We need tests :)
> 
> Both `users.deactivate()` and `devices.deactivate()` should be tested to make sure that when the connection raises x error containing y text, than it raises this new error `Py42ActiveLegalHoldError`

My bad. Could've sworn I added those. Just updated :)

unparalleled-js (approved):

PR #207, opened 2020-09-24
(bugfix) nil properties

ceciliastevens:
### Description of Change ###

Some device settings have the format {"@nil": "true"} if they have not previously been set. This updates the getter and setter in the base property to properly handle this format.

Additionally, I added the javaMemoryHeapMax property, which is an example that commonly uses this format.

### Issues Resolved ###
n/a, I did not file an issue around this.

### Testing Procedure ###
1. Create an SDK object
2. Retrieve device settings for a device in Code42 that has not previously altered the javaMemoryHeapMax property
3. Ensure that this property can be properly viewed

### PR Checklist ###

- [n/a] Add an entry to CHANGELOG.md describing this change (it does not look like we document a list of available properties anywhere, which is the only user-facing change made here)


ceciliastevens:
Another approach might be to define a converter that maps `{"@nil": "true"}` to `None` and vice versa? I'm not certain if that would be the best approach, especially since we don't know how many properties might have this structure.

src/py42/clients/settings/device_settings.py
```diff
@@ -166,6 +166,11 @@ def version(self):
     notes = SettingProperty(name=u"notes", location=[u"notes"])
     """Notes field for this device."""
 
+    javaMemoryHeapMax = SettingProperty(
```
unparalleled-js, line 169: Should be snake case

timabrmsn:
Does setting `javaMemoryHeapMax` on the server actually affect the client setting (I've never tested it)? 

Also, I wonder if we should convert the `{@nil: true}` value to `None` to make it more pythonic (like we do with converting string "true/false" to bools elsewhere). 

If `javaMemoryHeapMax` does affect the client setting, what type of value does it take? MB as an int? we should use a converter function to validate and maybe make it dead simple, like someone should be able to do:
 
`settings.java_heap_max_gb = 6`

ceciliastevens:
> Does setting `javaMemoryHeapMax` on the server actually affect the client setting (I've never tested it)?
> 
> Also, I wonder if we should convert the `{@nil: true}` value to `None` to make it more pythonic (like we do with converting string "true/false" to bools elsewhere).
> 
> If `javaMemoryHeapMax` does affect the client setting, what type of value does it take? MB as an int? we should use a converter function to validate and maybe make it dead simple, like someone should be able to do:
> 
> `settings.java_heap_max_gb = 6`

You can update a device's settings by pushing it a new config from the server, although I think the device will have to restart before it'll take.

Converting it to None would be an option, but the problem is where to do that; a lot of settings (not just this one) _could_ have an @nil: true structure.

The most common thing we do to set javaMemoryHeapMax is a number of megabytes with "m" at the end, but you could conceivably do something like "3g" as well.

timabrmsn:
I tried updating the `javaMemoryHeapMax` value for a test device, and while it shows it took on the server after I sent the update, after restarting the client it kept its previous value, and the server value also reverted to what the client has. 

It might be possible to get the client to pull down the new value somehow, but we'd need to figure out exactly what needs to be done to get it to stick before pulling this new property. 

That said, I know the `clientCommand` api works for remotely setting the xmx if the client is online. That might be a nice thing to implement into py42 for a number of reasons.

ceciliastevens:
> I tried updating the `javaMemoryHeapMax` value for a test device, and while it shows it took on the server after I sent the update, after restarting the client it kept its previous value, and the server value also reverted to what the client has.
> 
> It might be possible to get the client to pull down the new value somehow, but we'd need to figure out exactly what needs to be done to get it to stick before pulling this new property.
> 
> That said, I know the `clientCommand` api works for remotely setting the xmx if the client is online. That might be a nice thing to implement into py42 for a number of reasons.

It might be. I'll mark this one as read-only for now.

timabrmsn (approved):
Apologies for this getting lost and sitting so long! This change is good as is.

PR #209, opened 2020-09-25
Add filter fields to list of methods that allow new timestamp formats

alanag13:
### Description of Change ###
#205 added support for new timestamp formats to various methods, but the filters for file events and alerts were excluded from the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,13 +12,17 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- Timestamp argument accepts input in string type( format `yyyy-MM-dd HH:MM:SS` ), int or float type( epoch time ) as well as a datetime instance for following:
+- Timestamp arguments accept input in string type( format `yyyy-MM-dd HH:MM:SS` ), int or float type( epoch time, e.g. using `time.time()` ) as well as a `datetime` instance for the following:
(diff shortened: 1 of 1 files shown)


kiran-chaudhary (approved):
@unparalleled-js has addressed this in #208, hence the conflict, so merge accordingly.

CHANGELOG.md
```diff
@@ -12,13 +12,17 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Changed
 
-- Timestamp argument accepts input in string type( format `yyyy-MM-dd HH:MM:SS` ), int or float type( epoch time ) as well as a datetime instance for following:
+- Timestamp arguments accept input in string type( format `yyyy-MM-dd HH:MM:SS` ), int or float type( epoch time, e.g. using `time.time()` ) as well as a `datetime` instance for the following:
```
unparalleled-js, line 15: There are some other errors with this that I address and that got merged.

For instance:

1. Don't need to mention float again as they are already supported
2. The only new supported type for `departure_date` is `datetime`. `str` was already supported.

unparalleled-js: If you look at what is in master now..

alanag13: thanks, I'll adjust accordingly

PR #210, opened 2020-09-29
Bugfix/org settings handle inherited values

timabrmsn:
### Description of Change ###

Fixes bugs discovered in testing:

- Correctly handles Org setting inheritance by:
    - adding `incInheritedOrgInfo=True` on the `get_settings` method, preventing null values from being present in the json. 
    - updating SettingProperty to indicate if a setting is inheritable, and if so unset the inheritance flag when updating one of those values. 
   - adds new attributes `quota_settings_inherited` and `reporting_settings_inherited` which allows for configuring inheritance on the respective sections of the settings.

- Correctly handles "unlimited" values for certain numeric settings by not doing any transformation on the value if it's `-1`, also updated docstrings to indicate `-1 == unlimited`.


kiran-chaudhary (approved):
Tested, LGTM.

PR #212, opened 2020-09-30
Fix paging when user specified page_size < settings.items_per_page

alanag13:
### Description of Change ###

The way that paging is supposed to work is essentially:
- Get a page of results
- if the number of items returned in that page is less than the the requested page size, we are at the end and can stop getting pages
- otherwise, get the next page

There are two ways the user can set the page size of items to be retrieved:
- `py42.settings.items_per_page` (defaults to 100)
- `page_size` keyword arg on the `get_all()` function. This takes precedence over `items_per_page`.

Our paging logic currently determines whether or not it should continue by _always_ looking at `items_per_page`:
https://github.com/code42/py42/blob/5b693c5cc206c42db5786a5e26696fb372b3f0c0/src/py42/services/util.py#L9

This is a problem because if a `page_size` is specified that is less than `items_per_page` then the first page will always be treated as if it were the last page.

This is particularly problematic for `sdk.detectionlists.departing_employee.get_all()`and `sdk.detectionslists.high_risk_employee.get_all()` because they are exceptions to the default page size due to restrictions set on that service, and they are hard-coded to have a default page size of 100. Since `items_per_page` defaults to 500, you will only get the first page from these methods unless you set `items_per_page` to 100 or lower.

This PR resolves that issue by using `page_size` for that determination if one is supplied and otherwise using `items_per_page`.

### Testing Procedure ###
Adjust `items_per_page` below such that you should retrieve multiple pages of users in order to get all of them in your environment. `page_size` must be less than `items_per_page`.

```python
import py42.sdk
import py42.settings as settings

sdk = py42.sdk.from_local_account("console.us.code42.com", "user", "pass")
settings.items_per_page = 100

# in v1.8.1, this will always only get 1 page
# with this fix it should return `math.ceil(num_users / page_size)` pages.
for page in sdk.users.get_all(page_size=50):
    print("got page") 
```

### PR Checklist ###


src/py42/services/util.py
```diff
@@ -2,11 +2,12 @@
 
 
 def get_all_pages(func, key, *args, **kwargs):
-    item_count = settings.items_per_page
-    page_num = 0
     if kwargs.get("page_size") is None:
         kwargs[u"page_size"] = settings.items_per_page
-    while item_count >= settings.items_per_page:
+
+    item_count = page_size = kwargs[u"page_size"]
```
unparalleled-js, line 8: Can we add a test for this?

alanag13: planning on it, just wanted to make this available for physical testin

alanag13: @unparalleled-js added!

tests/settings/test_util.py
```diff
@@ -0,0 +1,69 @@
...
+    assert func.call_args_list[1][1]["page_size"] == page_size
+    assert func.call_args_list[2][1]["page_size"] == page_size
+    assert func.call_args_list[3][1]["page_size"] == page_size
+    assert func.call_count == 4
+
+
+def test_get_all_pages_with_no_input_page_size_calls_callback_with_expected_values(
+    get_three_three_item_pages,
+):
+    settings.items_per_page = 3
```
unparalleled-js, line 42: This is a good candidate for having a test class that implements `__enter__` and `__exit__` and resets the page .
Doing that ensures that resetting the items per page happens every time,

```
class TestPageSizer:
    def __init__(self, test_count, reset_count=500):
        self.test_count = test_count
        self.reset_count = reset_count

    def __enter__(self, *args, **kwargs):
        settings.items_per_page = self.test_count

    def __exit__(self, *args, **kwargs):
        settings.items_per_page = self.reset_count
```

```
with TestPageSizer(3, 500):
   <test>
```

unparalleled-js (approved):

tests/settings/test_util.py
unparalleled-js: Seeing resets in tests always make think of this

alanag13: totally agree, these tests are definitely a victim of copy/pasting from other places that did not do this

unparalleled-js: Ah that is good to know, I will remember that

PR #222, opened 2020-11-10
Fix/cloud alias limit

kiran-chaudhary:
### Description of Change ###
Added custom exception, when adding cloud alias fails with alias limit exceeded error.

### Issues Resolved ###
- closes #1190

### Testing Procedure ###
```
try:
   sdk.detectionlists.add_cloud_alias()
except Py42CloudAliasLimitExceeded as err:
```

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Bug where an empty destination list in a device's backup set broke creation of DeviceSettings objects for that device.
 
+### Added
+
+- Added new exception `Py42CloudAliasLimitExceeded` to throw if `add_cloud_alias` throws `400` and body contains reason
+`Cloud usernames must be less than or equal to 2`.
+
 ## 1.9.0 - 2020-10-02
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
-- Added new exception `Py42CloudAliasLimitExceeded` to throw if `add_cloud_alias` throws `400` and body contains reason
-`Cloud usernames must be less than or equal to 2`.
+- Added new exception `Py42CloudAliasLimitExceededError` to throw if `add_cloud_alias()` throws `400` and body contains
+reason `Cloud usernames must be less than or equal to`.
 
 ## 1.9.0 - 2020-10-02
 
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -168,7 +168,8 @@ class Py42UserAlreadyExistsError(Py42InternalServerError):
 
 
 class Py42CloudAliasLimitExceeded(Py42BadRequestError):
-    """An exception raised when a user already has 2 cloud aliases set."""
+    """An exception raised when max limit exceeds number of cloud aliases against
+     a user account."""
 
     def __init__(self, exception, message=None):
         message = message or u"Cloud alias limit exceeded."
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -167,9 +167,9 @@ class Py42UserAlreadyExistsError(Py42InternalServerError):
         super(Py42UserAlreadyExistsError, self).__init__(exception, message)
 
 
-class Py42CloudAliasLimitExceeded(Py42BadRequestError):
-    """An exception raised when number of cloud aliases exceeds against max allowed
-    limit."""
+class Py42CloudAliasLimitExceededError(Py42BadRequestError):
+    """An Exception raised when trying to add a cloud alias to a user when that user
+    already has the max amount of supported cloud aliases."""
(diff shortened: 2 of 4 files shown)


src/py42/services/detectionlists/user_profile.py
```diff
@@ -171,7 +172,14 @@ def add_cloud_alias(self, user_id, alias):
             u"cloudUsernames": [alias],
         }
         uri = self._make_uri(u"/addcloudusernames")
-        return self._connection.post(uri, json=data)
+        try:
+            return self._connection.post(uri, json=data)
+        except Py42BadRequestError as err:
+            if str(err.response.text).__contains__(
```
unparalleled-js, line 178: Instead of calling this magic method `__contains__`, you can just do: 

```
if "Cloud usernames must be less than or equal to 2" in err.response.text:
```

tests/services/detectionlists/test_user_profile.py
```diff
@@ -242,3 +266,17 @@ def test_refresh_posts_expected_data(
...
+        mock_connection,
+        user_context,
+        mock_user_client_error_on_adding_cloud_aliases,
+    ):
+        detection_list_user_client = DetectionListUserService(
+            mock_connection,
+            user_context,
+            mock_user_client_error_on_adding_cloud_aliases,
+        )
+        with pytest.raises(Py42CloudAliasLimitExceeded):
```
unparalleled-js, line 281: Additionally, you can verify the error message by doing this:

```
with pytest.raises(Py42CloudAliasLimitExceeded) as err:
   detection_list_user_client.add_cloud_alias("942897397520289999", "Test")

assert ""Cloud alias limit exceeded." in str(err.value)
```

src/py42/services/detectionlists/user_profile.py
kiran-chaudhary: changed.

tests/services/detectionlists/test_user_profile.py
kiran-chaudhary: added.

src/py42/services/detectionlists/user_profile.py
```diff
@@ -171,7 +172,12 @@ def add_cloud_alias(self, user_id, alias):
             u"cloudUsernames": [alias],
         }
         uri = self._make_uri(u"/addcloudusernames")
-        return self._connection.post(uri, json=data)
+        try:
+            return self._connection.post(uri, json=data)
+        except Py42BadRequestError as err:
+            if "Cloud usernames must be less than or equal to 2" in err.response.text:
```
alanag13, line 178: can we remove "2" from the end of this message to protect against the possibility that this limit could be changed? That way this exception will still trigger if we get `Cloud username must be less than or equal to 5`, for example.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Bug where an empty destination list in a device's backup set broke creation of DeviceSettings objects for that device.
 
+### Added
+
+- Added new exception `Py42CloudAliasLimitExceeded` to throw if `add_cloud_alias` throws `400` and body contains reason
+`Cloud usernames must be less than or equal to 2`.
```
alanag13, line 20: Adjust this to reflect the fact that `2` is no longer checked.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Bug where an empty destination list in a device's backup set broke creation of DeviceSettings objects for that device.
 
+### Added
+
+- Added new exception `Py42CloudAliasLimitExceeded` to throw if `add_cloud_alias` throws `400` and body contains reason
```
unparalleled-js, line 19: Error class names should end with `Error`.

`Py42CloudAliasLimitExceededError`

unparalleled-js, line 19: `add_cloud_alias` is a function and looks better if ending with `()`

`add_cloud_alias()`

src/py42/exceptions.py
```diff
@@ -167,6 +167,15 @@ def __init__(self, exception, message=None):
         super(Py42UserAlreadyExistsError, self).__init__(exception, message)
 
 
+class Py42CloudAliasLimitExceeded(Py42BadRequestError):
+    """An exception raised when number of cloud aliases exceeds against max allowed
```
unparalleled-js, line 171: The doc does not describe enough about _when_ it is actually raised.

Perhaps something like `An Exception raised when trying to add a cloud alias to a user when that user already has the max amount of supported cloud aliases`.

Note the missing word `the` between `when` and `number`,

Also note that this more accurately tells users to expect this exception during some type of `add` operation (adding aliases).

CHANGELOG.md

CHANGELOG.md

src/py42/exceptions.py
kiran-chaudhary: Changed.

unparalleled-js (approved):

PR #231, opened 2020-11-25
fix MFAError response check logic

timabrmsn:
The logic for Py42MFARequiredError has a bad "or" clause that always returns true, so any 401 error raises a MFARequiredError: 

```
elif raised_error.response.status_code == 401:
    if raised_error.response.text:
        if (
            "TOTP_AUTH_CONFIGURATION_REQUIRED_FOR_USER"
            or "TIME_BASED_ONE_TIME_PASSWORD_REQUIRED" in raised_error.response.text
        ):
            raise Py42MFARequiredError(raised_error) 
```

This change fixes that.

### DIFF
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -1,3 +1,5 @@
+import re
+
 from py42._compat import str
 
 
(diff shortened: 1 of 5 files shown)


alanag13 (approved):
Good catch.

src/py42/exceptions.py
```diff
@@ -194,12 +196,11 @@ def raise_py42_error(raised_error):
     if raised_error.response.status_code == 400:
         raise Py42BadRequestError(raised_error)
     elif raised_error.response.status_code == 401:
-        if raised_error.response.text:
-            if (
-                "TOTP_AUTH_CONFIGURATION_REQUIRED_FOR_USER"
-                or "TIME_BASED_ONE_TIME_PASSWORD_REQUIRED" in raised_error.response.text
-            ):
-                raise Py42MFARequiredError(raised_error)
+        if raised_error.response.text and re.search(
```
unparalleled-js, line 199: i see the error!  Ha.  I feel like there are missing tests, we should add them so that this never happens again

alanag13: I think there are tests for this, just always (incorrectly) passed -- but if not, then yes, we should add some.

alanag13: Could also add tests to make sure that the error is _not_ raised under the right conditions.

unparalleled-js: Oh I see, well then we would just a test to make sure that this _doesn't_ happen.

unparalleled-js: Looking at it, I only see one test for `TIME_BASED_ONE_TIME_PASSWORD_REQUIRED` but not the other constant, and then there should be 401 test that does not have either, so 2 total missing tests I believe!

timabrmsn: Yeah, there is a test but it just always passed. I can add a negative test too.

unparalleled-js:
What about a CHANGELOG update?  We should add this under `Fixed` something like `Fixed bug where 401s always raised TOCP error incorrectly`

timabrmsn:
> What about a CHANGELOG update? We should add this under `Fixed` something like `Fixed bug where 401s always raised TOCP error incorrectly`

Will add.

PR #232, opened 2020-11-30
Bugfix: storage login tokens > auth tokens

### LINKED ISSUE #230: [Bug] Storage node LoginToken is single-use, causing a lot of spurious 401 requests and retries
### Description

Currently the [LoginToken resource](https://www.crashplan.com/apidocviewer/#LoginToken) is being used as the backing auth for all storage node requests. However, the LoginToken is a single-use token meant to be used to generate a storage-node specific auth-token that can be used for subsequent requests. 

Since py42 is only using the LoginToken, almost every storage node request is duplicated because the storage node returns 401 on the second use of the token, which causes the `StorageTmpAuth` classes to renew with a new LoginToken each time. 

### Steps to Reproduce

1. Enable debug logging
2. Call a function that accesses resources on a storage node, (e.x. `stream_from_backup`)

### Expected Behavior
The storage node requests that get logged would all succeed and only be called once. 

### Actual Behavior
Each request to a storage node after the first success will fail with a 401, then succeed only after a new LoginToken has been requested: 

```
GET     /c42api/v3/auth/jwt
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
  params {
    "useBody": true
}
<Response [200]>
GET     /api/User/my
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
<Response [200]>
GET     /api/Computer/945056748960971397
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
  params {
    "idType": "guid",
    "incBackupUsage": true
}
<Response [200]>
POST    /api/LoginToken
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
<Response [200]>
POST    /api/DataKeyToken
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
<Response [200]>
POST    /api/WebRestoreSession
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
<Response [200]>
GET     /c42api/v3/BackupSets/945056748960971397/43
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
<Response [200]>
GET     /api/WebRestoreTreeNode
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
  params {
    "webRestoreSessionId": "<redacted>",
    "guid": 945056748960971397,
    "backupSetId": "1",
    "fileId": null,
    "timestamp": null,
    "showDeleted": true,
    "batchSize": null,
<Response [401]>
    "lastBatchFileId": null,
    "includeOsMetadata": null
}
POST    /api/LoginToken
<Response [200]>headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
GET     /api/WebRestoreTreeNode
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
  params {
    "webRestoreSessionId": "<redacted>",
<Response [200]>
    "guid": 945056748960971397,
    "backupSetId": "1",
    "fileId": null,
    "timestamp": null,
    "showDeleted": true,
    "batchSize": null,
    "lastBatchFileId": null,
    "includeOsMetadata": null
}
GET     /api/WebRestoreTreeNode
headers: {'Accept': 'application/json', 'Content-Type': 'application/json', 'Accept-Encoding': 'gzip, deflate', 'Connection': 'keep-alive'}
  params {
    "webRestoreSessionId": "<redacted>",
    "guid": 945056748960971397,
    "backupSetId": "1",
    "fileId": "10ac80587a05a2d4d9e37892ca3dc445",
    "timestamp": null,
    "showDeleted": true,
    "batchSize": null,
    "lastBatchFileId": null,
    "includeOsMetadata": null
}
<Response [401]>
POST    /api/LoginToken
<Response [200]>
... etc.
```

### Solution

The `StorageTmpAuth` classes should first get a LoginToken for the node, then use that to call the `AuthToken` resource on the storage node and store _that_ as the credential instead of the login token itself.

timabrmsn:
### Description of Change ###

This change refactors the StorageTmpAuth classes to use AuthTokens rather than the single-use LoginTokens for all storage node authentication. This greatly reduces the number of requests py42 needs to make when interacting with storage nodes, as previously each storage node request after the very first would fail, causing a new LoginToken request to be made to the authority, then back to the storage node to retry the failed request. 

### Issues Resolved ###
- fixes #230 

### Testing Procedure ###
On a pre-fix version, enable debug logging and call a py42 function that makes more than one request to a storage node (e.x. a web restore function that collects multiple files and thus needs to make many file metadata calls). 

Notice in the logging how most calls to the storage node are duplicated (with calls to "api/LoginToken" in between each duplicate). 

Then upgrade to fixed version, and make the exact same function call with debug logging. Notice only a single call to "api/LoginToken" and each storage node call is not duplicated. 

### PR Checklist ###

### DIFF
--- a/src/py42/services/storage/_auth.py
+++ b/src/py42/services/storage/_auth.py
@@ -1,28 +1,40 @@
 from py42.services._auth import C42RenewableAuth
+from py42.services._connection import Connection
 
 
-class StorageTmpAuth(C42RenewableAuth):
+class StorageAuth(C42RenewableAuth):
     def __init__(self):
-        super(StorageTmpAuth, self).__init__()
+        super(StorageAuth, self).__init__()
         self._storage_url = None
+        self._storage_connection = None
 
     def get_storage_url(self):
         self.get_credentials()
-        return self._server_url
+        return self._storage_url
 
     def get_tmp_auth(self):
         raise NotImplementedError()
 
+    def _get_auth_token(self, login_token):
+        uri = u"api/AuthToken"
+        response = self._storage_connection.post(
+            uri, headers={"Authorization": "login_token {}".format(login_token)}
+        )
+        token1, token2 = response.data
+        return u"token {}-{}".format(token1, token2)
+
     def _get_credentials(self):
-        login_info = self.get_tmp_auth()
-        login_token = login_info[u"loginToken"]
-        self._server_url = login_info[u"serverUrl"]
-        return u"login_token {}".format(login_token)
+        storage_url, login_token = self.get_tmp_auth()
+        self._storage_url = storage_url
+        self._storage_connection = (
+            self._storage_connection or _get_new_storage_connection(self._storage_url)
+        )
+        return self._get_auth_token(login_token)
 
 
-class FileArchiveTmpAuth(StorageTmpAuth):
+class FileArchiveAuth(StorageAuth):
     def __init__(self, connection, user_id, device_guid, destination_guid):
-        super(FileArchiveTmpAuth, self).__init__()
+        super(FileArchiveAuth, self).__init__()
         self._connection = connection
         self._user_id = user_id
         self._device_guid = device_guid
@@ -36,12 +48,14 @@ class FileArchiveTmpAuth(StorageTmpAuth):
             u"destinationGuid": self._destination_guid,
         }
         response = self._connection.post(uri, json=data)
-        return response
+        storage_url = response[u"serverUrl"]
+        login_token = response[u"loginToken"]
+        return storage_url, login_token
 
 
-class SecurityArchiveTmpAuth(StorageTmpAuth):
+class SecurityArchiveAuth(StorageAuth):
     def __init__(self, connection, plan_uid, destination_guid):
-        super(SecurityArchiveTmpAuth, self).__init__()
+        super(SecurityArchiveAuth, self).__init__()
         self._connection = connection
         self._plan_uid = plan_uid
         self._destination_guid = destination_guid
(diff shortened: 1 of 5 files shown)


tests/services/storage/test_auth.py
```diff
@@ -28,37 +31,52 @@ def mock_tmp_auth_conn(mock_connection, py42_response):
...
-def mock_v1_auth_conn(mock_connection, py42_response):
-    py42_response.text = '["TEST_V1", "TOKEN_VALUE"]'
-    mock_connection.post.return_value = py42_response
-    return mock_connection
+def mock_storage_auth_token_conn(mocker):
+    response = mocker.MagicMock(spec=Py42Response)
+    response.status_code = 200
+    response.encoding = None
+    response.__getitem__ = lambda _, key: json.loads(response.text)[key]
+    response.text = '["TEST_V1", "TOKEN_VALUE"]'
```
unparalleled-js, line 39: It's much easier to just mock the `requests.Request` and use a legit `Py42Response`, from my experience this saves even future headaches.

Then, all you need to do is set the `.text` on the mock response and let the normal `Py42Response` handle the rest.

Most of the mock response in py42 are currently like this

src/py42/services/storage/_auth.py
```diff
@@ -36,12 +48,14 @@ def get_tmp_auth(self):
             u"destinationGuid": self._destination_guid,
         }
         response = self._connection.post(uri, json=data)
-        return response
+        storage_url = response[u"serverUrl"]
```
alanag13, line 51: moving this and lines 67-69 into a shared private method is just about the only thing I could _maybe_ suggest on this PR. It's debatable whether or not that's even the right thing to do because if one of those endpoint were to change and not the other, we'd want to have the logic separated anyway.

timabrmsn: Yeah, I don't see any real benefit in pulling out such a simple task into its own method.

tests/services/storage/test_auth.py
timabrmsn: Ah, good to know. I was just copying the fixture code from conftest since re-using the same fixtures in the same test seemed to be causing them to overwrite each other.

unparalleled-js (approved):
Well done!  It works way faster, this is great.

PR #233, opened 2020-12-02
Feature/more than 10k results

### LINKED ISSUE #227: [Enhancement] Add support for nextPgToken
## Summary
It seems possible enough to use `query.page_num` to interact with queries of a size < 10,000. Per the documentation the API is supposed to return a nextPgToken if the request has over 10,000 results (Although this doesn't seem to be working in production yet as I continue to get responses with "nextPgToken":null). The documentation indicates that "A pgToken is the only way to page past 10,000 results. 

The documentation states the following: "If pgToken is supplied, pgNum is ignored. Provide empty string to retrieve the 'first page of results and null to use the pgNum value. Default is null." (https://forensicsearch-east.us.code42.com/forensic-search/queryservice/swagger-ui.html#/file-event-controller) .  

## Proposed API

add a self.page_token value in BaseQuery (https://github.com/code42/py42/blob/52abb32bdd397edbcf5824fbf0b2cf155b1128ce/src/py42/sdk/queries/__init__.py#L5) such that it can be overwritten in a similar manner to query.page_num. When pgToken is in use set pgNum to null. Ensure that resulting post request for example (https://github.com/code42/py42/blob/765dee58c90b4e53ea224e95c7d4662193243e62/src/py42/services/fileevent.py#L26) is generated to include pgToken when value is not default (otherwise it should not be included).

## Intended Use Case
Support querying of search_file_events for export to ELK. Even on a minute by minute basis that the cron job will fire at, this may have more than 10,000 events with a moderately sized deploy. It would be best if there was a method whereby these could automatically be pushed to an S3 bucket to prevent taxing the API.

As a note, it seems this approach cannot scale to over 120,000 events per minute, this seems to be a limitation of the product.

kiran-chaudhary:
### Description of Change ###

Added  method `get_all_file_events` that searches for file events for a given query and returns a token to retrieve remaining events along with the response.

### Issues Resolved ###

- closes #227 

### Testing Procedure ###
`sdk.securitydata.get_all_file_events(query)`
`sdk.securitydata.get_all_file_events(query, "Token")`

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,8 @@ reason `Cloud usernames must be less than or equal to`.
 - Added new exception `Py42UserNotOnListError` to throw error if `remove()` throws `404` when the user is not on a detection list.
 
 - Added new exception `Py42TooManyRequestsError` to raise errors with 429 HTTP status code.
-- Added new method `sdk.securitydata.get_all_file_events()` to retrieve all events when events are more than 10,000
+
+- Added new method `sdk.securitydata.search_all_file_events()` to retrieve all events when events are more than 10,000.
 
 ## 1.9.0 - 2020-10-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ reason `Cloud usernames must be less than or equal to`.
 - Added new exception `Py42UserNotOnListError` to throw error if `remove()` throws `404` when the user is not on a detection list.
 
 - Added new exception `Py42TooManyRequestsError` to raise errors with 429 HTTP status code.
+- Added new method `sdk.securitydata.get_all_file_events` to retrieve all events when events are more than 10,000
 
 ## 1.9.0 - 2020-10-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ reason `Cloud usernames must be less than or equal to`.
 - Added new exception `Py42UserNotOnListError` to throw error if `remove()` throws `404` when the user is not on a detection list.
 
 - Added new exception `Py42TooManyRequestsError` to raise errors with 429 HTTP status code.
-- Added new method `sdk.securitydata.get_all_file_events` to retrieve all events when events are more than 10,000
+- Added new method `sdk.securitydata.get_all_file_events()` to retrieve all events when events are more than 10,000
 
 ## 1.9.0 - 2020-10-02
 
--- a/docs/userguides/searches.md
+++ b/docs/userguides/searches.md
@@ -79,6 +79,22 @@ for event in file_events:
     print(event["md5Checksum"])
 ```
 
+If the number of events exceeds 10,000 against a query, use `securitydata.SecurityModule.search_all_file_events()`:
+
+```python
+
+query = FileEventQuery(ExposureType.eq(ExposureType.REMOVABLE_MEDIA))
+token = ""
+while True:
+    response = sdk.securitydata.search_all_file_events(query, page_token=token)
+    file_events = response["fileEvents"]
+    for event in file_events:
+        print(event["md5Checksum"])
+    token = response["nextPgToken"]
+    if token is None:
+        break
+```
+
 ```eval_rst
 .. _anchor_search_alerts:
 ```
--- a/docs/userguides/searches.md
+++ b/docs/userguides/searches.md
@@ -84,15 +84,15 @@ If the number of events exceeds 10,000 against a query, use `securitydata.Securi
 ```python
 
 query = FileEventQuery(ExposureType.eq(ExposureType.REMOVABLE_MEDIA))
-token = ""
-while True:
-    response = sdk.securitydata.search_all_file_events(query, page_token=token)
+response = sdk.securitydata.search_all_file_events(query)
+file_events = response["fileEvents"]
+for event in file_events:
+    print(event["md5Checksum"])
+while response["nextPgToken"] is not None:
+    response = sdk.securitydata.search_all_file_events(query, page_token=response["nextPgToken"])
     file_events = response["fileEvents"]
     for event in file_events:
         print(event["md5Checksum"])
-    token = response["nextPgToken"]
-    if token is None:
-        break
 ```
 
 ```eval_rst
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -193,7 +193,8 @@ class SecurityDataClient(object):
             query (:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`): Also
                 accepts a raw JSON str.
             page_token (str): A token used to indicate the starting point for additional page results. Defaults to empty string.
-                For the first page do not pass page_token, for all consecutive pages pass the token returned from previous response.
+                For the first page do not pass page_token, for all consecutive pages pass the token returned from previous response,
+                from field `nextPgToken`.
 
         Returns:
             tuple (str, :class:`py42.response.Py42Response`): A tuple of token, to fetch remaining events, and response containing events.
@@ -201,7 +202,7 @@ class SecurityDataClient(object):
 
         query.page_token = page_token
         response = self._file_event_service.search(query)
-        return response["nextPgToken"], response
+        return response
 
     def stream_file_by_sha256(self, checksum):
         """Stream file based on SHA256 checksum.
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -191,7 +191,7 @@ class SecurityDataClient(object):
 
         Args:
             query query (`:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery` or str): The file event query
-            to filter search results.
+                to filter search results.
             page_token (str): A token used to indicate the starting point for additional page results. Defaults to empty string.
                 For the first page do not pass page_token, for all consecutive pages pass the token returned from previous response,
                 from field `nextPgToken`.
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -198,9 +198,7 @@ class SecurityDataClient(object):
             events.
         """
 
-        response = self._file_event_service.search(query)
-        yield response
-        query.pgToken = response["nextPgToken"]
+        query.pgToken = ""
         while query.pgToken is not None:
             response = self._file_event_service.search(query)
             yield response
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -172,12 +172,11 @@ class SecurityDataClient(object):
         )
 
     def search_file_events(self, query):
-        """Searches for file events, returns first 10,000 events.
+        """Searches for file events, returns up to the first 10,000 events.
         `REST Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Forensic_File_Search_API>`__
 
         Args:
-            query (:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`): Also
-                accepts a raw JSON str.
+            query query (`:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery` or str): The file event query to filter search results.
 
         Returns:
             :class:`py42.response.Py42Response`: A response containing the first 10,000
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ reason `Cloud usernames must be less than or equal to`.
 - Added new exception `Py42UserNotOnListError` to throw error if `remove()` throws `404` when the user is not on a detection list.
 
 - Added new exception `Py42TooManyRequestsError` to raise errors with 429 HTTP status code.
+- Added new method `sdk.securitydata.get_all_file_events` to retrieve all events when events are more than 10,000
```
unparalleled-js, line 24: Could you put the `()` at the end of the method name?

src/py42/clients/securitydata.py
```diff
@@ -185,6 +185,24 @@ def search_file_events(self, query):
...
+        `REST Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Forensic_File_Search_API>`__
+
+        Args:
+            query (:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`): Also
+                accepts a raw JSON str.
+            pg_token (str): A token used to indicate the starting point for additional page results. Defaults to empty string.
+                For the first page do not pass pg_token, for all consecutive pages pass the token returned from previous response.
+
+        Returns:
+            tuple (str, :class:`py42.response.Py42Response`): A tuple of token, to fetch remaining events, and response containing events.
```
unparalleled-js, line 199: Since the `nextPgToken` is on the response, I don't see why this needs to return a tuple

src/py42/sdk/queries/__init__.py
```diff
@@ -8,6 +8,7 @@ def __init__(self, *args, **kwargs):
         self._group_clause = kwargs.get(u"group_clause", u"AND")
         self.page_number = kwargs.get(u"page_number") or 1
         self.page_size = kwargs.get(u"page_size") or settings.security_events_per_page
+        self.pgToken = kwargs.get(u"pg_token") or None
```
unparalleled-js, line 11: This should be snake_case, not camelCase

src/py42/sdk/queries/fileevents/file_event_query.py
```diff
@@ -34,14 +34,20 @@ def __str__(self):
...
-        json = u'{{"groupClause":"{0}", "groups":[{1}], "pgNum":{2}, "pgSize":{3}, "srtDir":"{4}", "srtKey":"{5}"}}'.format(
-            self._group_clause,
-            groups_string,
-            self.page_number,
-            self.page_size,
-            self.sort_direction,
-            self.sort_key,
-        )
+
+        items = []
```
alanag13, line 38: This can be simplified:
```python
if self.page_token is not None:
    paging_prop = '"pgNum":{0}'.format(self.page_number)
else:
    paging_prop = '"pgToken":{0}'.format(self.page_token)

json = u'{{"groupClause":"{0}", "groups":[{1}], "srtDir":"{2}", "srtKey":"{3}", {4} }}'.format(
    self._group_clause,
    groups_string,
    self.sort_direction,
    self.sort_key,
    self.paging_props
)


```

src/py42/sdk/queries/fileevents/file_event_query.py
```diff
@@ -34,14 +34,20 @@ def __str__(self):
...
+        items.append(u'"groupClause":"{}"'.format(self._group_clause))
+        items.append(u'"groups":[{}]'.format(groups_string))
+        items.append(u'"srtDir":"{}"'.format(self.sort_direction))
+        items.append(u'"srtKey":"{}"'.format(self.sort_key))
+        if self.pgToken is not None:
+            items.append(u'"pgToken":"{}"'.format(self.pgToken))
+        else:
+            items.append(u'"pgNum":{}'.format(self.page_number))
+            items.append(u'"pgSize":{}'.format(self.page_size))
```
alanag13, line 47: `pgSize` can still be used in combination with `pgToken`.

src/py42/sdk/queries/fileevents/file_event_query.py
```diff
@@ -53,6 +59,7 @@ def __iter__(self):
             u"pgSize": self.page_size,
             u"srtDir": self.sort_direction,
             u"srtKey": self.sort_key,
+            u"pgToken": self.pgToken,
```
unparalleled-js, line 62: This is the reason you are getting 500. It is supposed to be `pageToken`, not `pgToken`

src/py42/clients/securitydata.py
alanag13: it doesn't have to, but I thought it might be a nice convenience. We don't have any other methods that do this though, so maybe you're right and should just be consistent here. In that case, we should update the docstring to clearly indicate where the get the next value from.

CHANGELOG.md
kiran-chaudhary: added.

src/py42/sdk/queries/__init__.py
kiran-chaudhary: changed.

src/py42/sdk/queries/fileevents/file_event_query.py
kiran-chaudhary: Changed.

src/py42/sdk/queries/fileevents/file_event_query.py
kiran-chaudhary: Added.

src/py42/clients/securitydata.py
kiran-chaudhary: Changed. I feel since this is a different use-case it seems returning the tuple is more intuitive.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ reason `Cloud usernames must be less than or equal to`.
 - Added new exception `Py42UserNotOnListError` to throw error if `remove()` throws `404` when the user is not on a detection list.
 
 - Added new exception `Py42TooManyRequestsError` to raise errors with 429 HTTP status code.
+- Added new method `sdk.securitydata.get_all_file_events()` to retrieve all events when events are more than 10,000
```
unparalleled-js, line 24: Should have a newline between this entry and the one above

src/py42/clients/securitydata.py
```diff
@@ -172,7 +172,7 @@ def get_all_user_security_events(
         )
 
     def search_file_events(self, query):
-        """Searches for file events.
+        """Searches for file events, returns first 10,000 events.
```
unparalleled-js, line 175: More accurate to sy `returns up to the first 10,000 events`

src/py42/clients/securitydata.py
```diff
@@ -185,6 +185,25 @@ def search_file_events(self, query):
         """
         return self._file_event_service.search(query)
 
+    def get_all_file_events(self, query, page_token=""):
```
unparalleled-js, line 188: We should consider renaming this to `search_for_all_file_events(query)` because all of the search methods in py42 take these query objects, and that maintains that consistency.  @alanag13

src/py42/clients/securitydata.py
```diff
@@ -185,6 +185,25 @@ def search_file_events(self, query):
...
+        """Searches for all file events.
+        `REST Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Forensic_File_Search_API>`__
+
+        Args:
+            query (:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`): Also
```
unparalleled-js, line 193: This is feedback for both this param description and the one in `search_file_events()`:

The type should be 

```
`:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery` or str
```

And the description should just be a description and not mention the type, like

```
The file event query to filter search results.
```

src/py42/clients/securitydata.py
unparalleled-js: We could add a user-guide or amend a user-guide to show an example of getting file events with the tokens..

In fact, we really should do that either way.

src/py42/sdk/queries/fileevents/file_event_query.py
```diff
@@ -34,13 +34,22 @@ def __str__(self):
...
-        json = u'{{"groupClause":"{0}", "groups":[{1}], "pgNum":{2}, "pgSize":{3}, "srtDir":"{4}", "srtKey":"{5}"}}'.format(
-            self._group_clause,
-            groups_string,
-            self.page_number,
-            self.page_size,
-            self.sort_direction,
-            self.sort_key,
+
+        if self.page_token is not None:
+            paging_prop = '"pgToken":"{}"'.format(self.page_token)
```
unparalleled-js, line 39: Missing `u`s here and below

unparalleled-js:
After the other feedback is addressed, I think these changes warrant an update to `docs/userguides/searches.md`

src/py42/clients/securitydata.py
alanag13: agreed, also, since this doesn't use a generator we should not use `get_all`.

src/py42/sdk/queries/fileevents/file_event_query.py
kiran-chaudhary: Done.

src/py42/clients/securitydata.py
kiran-chaudhary: Done.

src/py42/clients/securitydata.py
kiran-chaudhary: Changed.

CHANGELOG.md
kiran-chaudhary: Added.

src/py42/clients/securitydata.py
kiran-chaudhary: Added user-guide in `searches.md`

docs/userguides/searches.md
```diff
@@ -79,6 +79,22 @@ for event in file_events:
     print(event["md5Checksum"])
 ```
 
+If the number of events exceeds 10,000 against a query, use `securitydata.SecurityModule.search_all_file_events()`:
+
+```python
+
+query = FileEventQuery(ExposureType.eq(ExposureType.REMOVABLE_MEDIA))
+token = ""
```
unparalleled-js, line 87: Because of how we made it, this still works when token is initialized to `None`. It might be less confusing if we do that or else they may think that it is required to use empty string at the start.

tests/clients/test_securitydata.py
```diff
@@ -1313,3 +1320,67 @@ def test_stream_file_by_md5_when_find_file_version_returns_204_status_code_raise
             security_client.stream_file_by_md5("mdhash")
 
         assert e.value.args[0] == PDS_EXCEPTION_MESSAGE.format("mdhash")
+
+    def test_get_all_file_events_calls_search_with_expected_params_when_pg_token_is_not_passed(
```
unparalleled-js, line 1324: These test signatures need to be renamed as `test_search_all_file_events...()` since the method name changed

kiran-chaudhary: Changed.

docs/userguides/searches.md
kiran-chaudhary: Changed.

src/py42/clients/securitydata.py
```diff
@@ -172,19 +172,38 @@ def get_all_user_security_events(
...
+            query query (`:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery` or str): The file event query to filter search results.
 
         Returns:
             :class:`py42.response.Py42Response`: A response containing the first 10,000
             events.
         """
         return self._file_event_service.search(query)
 
+    def search_all_file_events(self, query, page_token=""):
+        """Searches for all file events, returning a page of response with a token in the response to retrieve next page.
```
unparalleled-js, line 188: Missing `s` on `responses`

unparalleled-js, line 188: `Page of response` sounds funny, perhaps `page of events`

src/py42/clients/securitydata.py
```diff
@@ -172,19 +172,38 @@ def get_all_user_security_events(
...
+
+        `REST Documentation <https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Forensic_File_Search_API>`__
+
+        Args:
+            query query (`:class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery` or str): The file event query
+                to filter search results.
+            page_token (str): A token used to indicate the starting point for additional page results. Defaults to empty string.
```
unparalleled-js, line 195: This should be `str or None` It should default to None

unparalleled-js: You should not default to empty string in python or weird things can happen, instead default to `None` and set it to emptry string if it's `None` within the method.

unparalleled-js, line 195: Should be `(str, optional)`

src/py42/clients/securitydata.py
```diff
@@ -172,19 +172,38 @@ def get_all_user_security_events(
...
+                For the first page do not pass page_token, for all consecutive pages pass the token returned from previous response,
```
unparalleled-js, line 196: The punctuation is a little off here.

`For the first page, do not pass page_token.  For all consecutive pages, pass the token returns from the previous response from the field `nextPgToken`.

unparalleled-js (approved):
I found a couple more litte documentation improvements, feel free to address those else it looks pretty good. Thanks!

src/py42/clients/securitydata.py
kiran-chaudhary: Changed.

src/py42/clients/securitydata.py
kiran-chaudhary: Added.

src/py42/clients/securitydata.py
kiran-chaudhary: Changed, rephrased `pass the token returns from the previous...` to `pass the token from the previous...`

PR #236, opened 2020-12-03
Bugfix/handle restore strategy

unparalleled-js:
### Description of Change ###

Adding ability to use the `file_location` parameter for restoring to original location during a `archive.stream_to_device()`.
The constants are available in `py42.services.storage.restore.PushRestoreLocation`

Also, you will get a custom exception, `Py42BadRestoreError`, when creating the push restore fails.. 

AND if the GUIDs differ and you are using the `ORIGINAL_LOCATION` location strategy, you will get an additional warning message describing that, as that may be related to the error.

### Issues Resolved ###
Push restore feature, no github issue.

- closes #

### Testing Procedure ###
*Verify you can use the file_location param and that it shows up in DEBUG request logs.

* do a Push restore when you know the devices have differing paths and use the ORIGINAL_LOCATION strategy and get the custom exception.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/_archiveaccess/__init__.py
+++ b/src/py42/clients/_archiveaccess/__init__.py
@@ -2,7 +2,9 @@ import posixpath
 from collections import namedtuple
 
 from py42.exceptions import Py42ArchiveFileNotFoundError
-
+from py42.exceptions import Py42BadRequestError
+from py42.exceptions import Py42BadRestoreError
+from py42.services.storage.restore import PushRestoreLocation
 
 # Data for initiating a web or push restore.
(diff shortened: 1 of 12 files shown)


unparalleled-js:
I am realizing that the`restore_path` parameter, which is currently required, is not really needed or used if using `ORIGINAL_LOCATION` for the `file_location` parameter.  Therefore, I don't think it needs to be required here.

Alternatively, we could combine these params such that it will check if the `restore_path` parameter is the value `ORIGINAL_LOCATION` from which it will set the `file_location` parameter automatically (and set `restore_path` to be the empty string)

PR #237, opened 2020-12-03
Bugfix/handle restore strategy

unparalleled-js:
### Description of Change ###

Ability to restore to original location parameter `archive.stream_to_device()` by passing the string "ORIGINAL_LOCAITON" for the `restore_path` parameter.

Also, you will get a custom exception, `Py42BadRestoreError`, when creating the push restore fails.

AND if the GUIDs differ and you are using the `ORIGINAL_LOCATION` location strategy, you will get an additional warning message describing that, as that may be related to the error.

### Issues Resolved ###
Push restore feature, no github issue.

- closes #

### Testing Procedure ###
*Verify you can use the file_location param and that it shows up in DEBUG request logs.

* do a Push restore when you know the devices have differing paths and use the ORIGINAL_LOCATION strategy and get the custom exception.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/_archiveaccess/restoremanager.py
+++ b/src/py42/clients/_archiveaccess/restoremanager.py
@@ -1,5 +1,6 @@
 import time
 
+from py42.services.storage.restore import PushRestoreLocation
 from py42.settings import debug
 from py42.util import format_dict
 
@@ -114,10 +115,16 @@ class RestoreJobManager(_RestorePoller):
         file_selections,
         backup_set_id,
         show_deleted,
-        file_location,
     ):
         num_files = sum([fs.num_files for fs in file_selections])
         num_bytes = sum([fs.num_bytes for fs in file_selections])
+        file_location = None
+
+        # Use expected request parameters for restoring to the original location.
+        if restore_path == PushRestoreLocation.ORIGINAL:
+            file_location = restore_path
+            restore_path = u""
+
         return self._storage_archive_service.start_push_restore(
             self._device_guid,
             accepting_guid,
--- a/src/py42/clients/_archiveaccess/restoremanager.py
+++ b/src/py42/clients/_archiveaccess/restoremanager.py
@@ -114,6 +114,7 @@ class RestoreJobManager(_RestorePoller):
         file_selections,
         backup_set_id,
         show_deleted,
+        file_location,
     ):
         num_files = sum([fs.num_files for fs in file_selections])
         num_bytes = sum([fs.num_bytes for fs in file_selections])
@@ -132,6 +133,7 @@ class RestoreJobManager(_RestorePoller):
             num_files,
             num_bytes,
             show_deleted=show_deleted,
+            file_location=file_location,
         )
 
     def _wait_for_job(self, response):
--- a/src/py42/clients/archive.py
+++ b/src/py42/clients/archive.py
@@ -130,7 +130,9 @@ class ArchiveClient(object):
                 the archive.
             device_guid (str): The GUID of the device the file belongs to.
             accepting_device_guid (str): The GUID of the device accepting the restore.
-            restore_path (str): The path on the accepting device where the restore will be saved.
+            restore_path (str, optional): The path on the accepting device where the restore will be
+                saved. Alternatively, pass in the value ``ORIGINAL_LOCATION`` to restore the file
+                to the original location, which may be the case if you are replacing a device
             destination_guid (str, optional): The GUID of the destination that stores the backup
                 of the file. If None, it will use the first destination GUID it finds for your
                 device. 'destination_guid' may be useful if the file is missing from one of your
@@ -143,9 +145,6 @@ class ArchiveClient(object):
                 key archive security. Defaults to None.
             show_deleted (bool, optional): Set to True to include deleted files when restoring a directory.
                 Defaults to None.
-            file_location (string or None, optional): Either ``ORIGINAL_LOCATION`` or ``TARGET_DIRECTORY``,
-                as defined in :class:`py42.services.storage.restore.PushRestoreLocation`, that indicates
-                where to restore files to.
             file_size_calc_timeout (int, optional): Set to limit the amount of seconds spent calculating
                 file sizes when crafting the request. Set to 0 or None to ignore file sizes altogether.
                 Defaults to 10.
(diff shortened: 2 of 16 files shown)


src/py42/clients/_archiveaccess/restoremanager.py
```diff
@@ -117,6 +118,13 @@ def send_stream(
     ):
         num_files = sum([fs.num_files for fs in file_selections])
         num_bytes = sum([fs.num_bytes for fs in file_selections])
+        file_location = None
+
+        # Use expected request parameters for restoring to the original location.
+        if restore_path == PushRestoreLocation.ORIGINAL:
```
unparalleled-js, line 124: I didn't see a point in having two parameters, one for the strategy and one for the restore path because if you want `ORIGINAL_LOCAITON`, the restore path parameter does nothing (and yet it is still required 🤔).  Therefore, I thought the best way would be to allow users to say "ORIGINAL_LOCATION" as a special value for `restore_path` and have py42 figure out the rest.

src/py42/clients/archive.py
```diff
@@ -120,6 +120,7 @@ def stream_to_device(
         archive_password=None,
         encryption_key=None,
         show_deleted=None,
+        file_location=None,
```
timabrmsn, line 123: Should this actually be exposed on the public client method?

src/py42/clients/archive.py
```diff
@@ -129,7 +130,9 @@ def stream_to_device(
                 the archive.
             device_guid (str): The GUID of the device the file belongs to.
             accepting_device_guid (str): The GUID of the device accepting the restore.
-            restore_path (str): The path on the accepting device where the restore will be saved.
+            restore_path (str, optional): The path on the accepting device where the restore will be
+                saved. Alternatively, pass in the value ``ORIGINAL_LOCATION`` to restore the file
+                to the original location, which may be the case if you are replacing a device
```
timabrmsn, line 135: needs a period.

src/py42/exceptions.py
```diff
@@ -210,6 +210,17 @@ def __init__(self, exception, message=None):
         super(Py42BadRequestError, self).__init__(exception, message)
 
 
+class Py42BadRestoreError(Py42BadRequestError):
```
timabrmsn, line 213: Would it make sense to name this `Py42BadRestoreRequestError` instead for clarity that it's the request itself that was bad?

src/py42/services/storage/restore.py
```diff
@@ -69,5 +77,23 @@ def start_push_restore(
...
+            if u"CREATE_FAILED" in err.response.text:
+                additional_message = None
+                if (
+                    device_guid != accepting_device_guid
+                    and file_location == PushRestoreLocation.ORIGINAL
+                ):
+                    additional_message = (
+                        u"Warning: Trying to restore to the original "
+                        "location when the accepting GUID '{}' is "
+                        "different from the archive source GUID "
```
timabrmsn, line 94: Can this be the cause of the CREATE_FAILED error, but the server response doesn't make it clear that's why it failed?

unparalleled-js: We are not sure that it is the cause.

However, we are sure that we got the pre-condition for the cause and that the GUIDs are different, thus, we are adding an additional WARNING to the error message.  

However, do you think that is too assumptious?

src/py42/clients/archive.py
unparalleled-js: I don't really see why.  Like, there are only two options for the API parameter:

`ORIGINAL_LOCAITION`: for the restoring to the original location (the one to satisfy the acceptance criteria) and 

`TARGET_LOCATION`: which appears to be the default behavior and requires the `restore_path` parameter.  

When using both `restore_path` and `file_location=ORIGINAL_LOCATION`, it does not honor `restore_path` because you are telling it to restore the files to the original location.

Therefore, it would not make sense that `restore_path` is a required parameter - yet it is.

unparalleled-js: But, we do need to know either the restore path or if they just want to use the `original location` strategy.

`restore_path=ORIGINAL_LOCTION` would suffice for me but at minimum, `restore_path` should not be a required parameter but it would one or the other (either provide the `file_location` param or the `restore_path` param)

src/py42/services/storage/restore.py
unparalleled-js: Or is this obviously the reason of the cause. I don't understand 100% all the use-cases for push restore, but I know tht the `ORIGINAL_LOCATION` strategy can have to do with device migration, in which case the device guids would already be the same one as they would be restoring a device to the new one (putting the files there for the first time).

src/py42/clients/archive.py
unparalleled-js: Haha no

src/py42/services/storage/restore.py
unparalleled-js: The server really does not make it super clear, just says the restore either does not exist or couldn't be created.

timabrmsn: Yeah, I'm not exactly sure how things can break in this specific case, which is why I was asking. 

If you try ORIGINAL_LOCATION with different source/target guids, and it fails. Then switch to TARGET_DIRECTORY and it works with only that change, then it makes sense to have this error text. 

But I wouldn't want to send someone down a rabbit-trail with this message if that isn't actually a problem. 

Does the UI present similar warnings at all? I think there's one about cross-operating system restores (.e.g a Windows archive being pushed to a Mac/Linux device).

unparalleled-js: Yes, that is exactly what I tried - took a working restore, that set the restore_path and everything, and it was using two devices... and then I switched to ORIGINAL_LOCATION and I got this generic error about it not being found or created.

As far as permitting different OS restores, I haven't tried messing around with that parameter at all - do you think that is needed now?

timabrmsn: Ok, if we know that just changing that flag can cause it to fail, then the warning is warranted as is. I don't think we need to do anything specific with cross-OS restores, it's very edge case.

timabrmsn (approved):

src/py42/services/storage/restore.py
unparalleled-js: Hmm... I am wondering if `restoreFullPath` is set True, then I could see this making sense, so maybe it's best not to have this additional warning.

unparalleled-js: Yeah I removed it for now

PR #238, opened 2020-12-07
impl

unparalleled-js:
### Description of Change ###

Allows for overwriting /renaming files when using ORIGINAL_LOCATION as well ake restoring for different OS versions work

### Issues Resolved ###
Push Restore

- closes #

### Testing Procedure ###
Push Restore testing ongoing

### PR Checklist ###

### DIFF
--- a/src/py42/clients/archive.py
+++ b/src/py42/clients/archive.py
@@ -145,8 +145,9 @@ class ArchiveClient(object):
                 key archive security. Defaults to None.
             show_deleted (bool, optional): Set to True to include deleted files when restoring a directory.
                 Defaults to None.
-            overwrite_existing_files (bool, optional): Set to True when using ORIGINAL_LOCATION for the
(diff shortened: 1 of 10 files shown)


src/py42/clients/archive.py
```diff
@@ -144,6 +145,8 @@ def stream_to_device(
                 key archive security. Defaults to None.
             show_deleted (bool, optional): Set to True to include deleted files when restoring a directory.
                 Defaults to None.
+            overwrite_existing_files (bool, optional): Set to True when using ORIGINAL_LOCATION for the
+                restore_path argument to overwrite any existing files part of the restore.
```
timabrmsn, line 149: to overwrite any existing files with the restored data. If False (the default), any existing files that match a path being restored will first get renamed.

timabrmsn (approved):
Apart from my doc note for `overwrite_existing_files` this looks good.

PR #244, opened 2020-12-14
Fix docs

kiran-chaudhary:
### Description of Change ###

Rectified path in docs.

### Issues Resolved ###
None

### Testing Procedure ###
NA
### PR Checklist ###

### DIFF
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -81,7 +81,7 @@ class SecurityDataClient(object):
         `Support Article <https://support.code42.com/Administrator/6/Configuring/Endpoint_monitoring>`__
 
         Args:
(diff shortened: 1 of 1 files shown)


kiran-chaudhary:
observed `get_security_plan_storage_info_list` at time returns `None`, shouldn't the default be an empty list given that the method returns list as a response??

unparalleled-js (approved):

unparalleled-js:
> observed `get_security_plan_storage_info_list` at time returns `None`, shouldn't the default be an empty list given that the method returns list as a response??

Maybe? I don't know. It doesn't matter too much but we also shouldn't change it because it'd be a breaking change.  You add this to the list of proposed 2.0 changes if you want.

PR #249, opened 2020-12-15
Integration tests setup

kiran-chaudhary:
### Description of Change ###

Added integration tests for alerts and devices.

All other integration tests PR would be raised against this branch. 

In `devices` tests that change state of the system were failing with 403 error are marked to be skipped, the one with 404 error need to find the right device-id.

### PR Checklist ###


integration/test_devices.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@pytest.mark.skip("Failing with 403.")
+def test_deauthorize(connection):
+    response = connection.devices.deauthorize(891804948033320117)
+    assert response.status_code == 201
+
+
+@pytest.mark.skip("Failing with 403.")
+def test_reactivate(connection):
```
unparalleled-js, line 49: These use the legacy IDs, that is probably why there are failing for you

integration/test_devices.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+
+
+@pytest.mark.skip("Failing with 403.")
```
unparalleled-js, line 4: Use legacy computer ID and this won't fail

run_integration.py
```diff
@@ -0,0 +1,13 @@
+import os
+import sys
+
+if __name__ == "__main__":
+    if sys.argv[1] and sys.argv[2]:
+        os.environ["C42_USER"] = sys.argv[1]
+        os.environ["C42_PW"] = sys.argv[2]
+        rc = os.system("pytest ./integration -v -rsxX -l --tb=short --strict")
+        sys.exit(rc)
+    else:
```
unparalleled-js, line 10: I would explicitly `exit(1)` so that if this ever runs on a pipeline, it will purposely fail

unparalleled-js: Or just raise an exception

`raise Exception("Error: Missing username and password")

unparalleled-js, line 10: This should fail, so you can raise an exception like `raise Exception("Error: missing Code42 username and password parameters")`

unparalleled-js (changes requested):
* Use legacy IDs for computers
* Raise exception if missing creds

integration/test_devices.py
```diff
@@ -0,0 +1,69 @@
...
+    response = connection.devices.get_agent_full_disk_access_state(891804948033320117)
+    assert response.status_code == 200
+
+
+def test_get_by_guid(connection):
+    response = connection.devices.get_by_guid(891804948033320117)
+    assert response.status_code == 200
+
+
+def test_get_settings(connection):
```
kiran-chaudhary, line 21: This test passes intermittently, at times returning `Py42Response`, 
`>>> sdk.devices.get_settings(891804948033320117)
<Py42Response [status=200, data={'SUPPORT_ACCESS': {'id': '2686', 'locked': False, 'scope': 'ORG', 'value': 'false'}, 'c42.msa.acceptance': {'id': '2686', 'locked': True, 'scope': 'ORG', 'value': '890854304274...T22:22:53:922'}, 'c42_indexing_enabled': {'locked': False, 'scope': 'SYSTEM', 'value': 'false'}, 'c42_indexing...ts_exclusions': {'locked': False, 'scope': 'SYSTEM', 'value': '3dm,3ds,3g2,...uv,z,zip,zipx'}, ...}]>`

otherwise `<DeviceSettings: guid: 891804948033320117, name: DESKTOP-3DG3V50>`. and hence assert statement fails.

run_integration.py
kiran-chaudhary: Changed.

integration/test_devices.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_block(connection):
+    response = connection.devices.block(251691)
+    assert response.status_code == 201
+
+
+@pytest.mark.skip("Failing with 404.")
+def test_get_agent_full_disk_access_state(connection):
+    response = connection.devices.get_agent_full_disk_access_state(891804948033320117)
```
unparalleled-js, line 11: The newer IDs (the longer ones) should probably be passes in as strings. That is how they get returned on responses.

The legacy IDs are numbers though.

integration/test_devices.py
unparalleled-js: Weird.. The return type is expected to be a `DeviceSettings` object.

The method `update_settings()` however does return a `Py42Response`.

This guides explains it more: https://py42docs.code42.com/en/stable/userguides/devicesettings.html

Were you perhaps calling update when you got the `Py42Response`? I am not seeing how in the code it could happen intermittently; I find that really peculiar! Can you dig in more?

unparalleled-js: @timabrmsn knows most about this feature as well

timabrmsn: Are we testing older versions of py42? `get_settings()` used to just call the [DeviceSetting](https://www.crashplan.com/apidocviewer/#DeviceSetting) endpoint, which is what that Py42Response result looks like.

timabrmsn: 1.9.0 is when this behavior changed.

run_integration.py
```diff
@@ -0,0 +1,13 @@
+import os
+import sys
+
+if __name__ == "__main__":
```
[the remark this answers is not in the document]
kiran-chaudhary: either is fine, unless there is a special need for either of them.
We have followed the same pattern done in 'extractor' I guess.

unparalleled-js:
I commented this on one of your other PRs, but what do you think of adding a helper test method that just checks that the status code is in the 200 range:

```assert response.status_code in range(200, 299)``` or something like that and just calling that for most tests

File: integration/test_devices.py (hunk 4 above)
Comment (kiran-chaudhary) [reply integration/test_devices.py:20]: changed.

File: integration/test_devices.py (hunk 5 above)
Comment (kiran-chaudhary) [reply integration/test_devices.py:11]: As discussed, skipping this tests.

**Comment by kiran-chaudhary** [comment]
> I commented this on one of your other PRs, but what do you think of adding a helper test method that just checks that the status code is in the 200 range:
> 
> `assert response.status_code in range(200, 299)` or something like that and just calling that for most tests
> 
> That will make it less important if it returns 204 vs 200 and if the mockserver accidentally uses 200 instead of 204, it shouldn't make a difference

I am planning to do it once all PR's are merged into this one, so that I can add in all of them together with another PR just having this changes.

File: run_integration.py [hunk 7]
```diff
@@ -0,0 +1,13 @@
...
+    if sys.argv[1] and sys.argv[2]:
+        os.environ["C42_USER"] = sys.argv[1]
+        os.environ["C42_PW"] = sys.argv[2]
+        rc = os.system("pytest ./integration -v -rsxX -l --tb=short --strict")
+        sys.exit(rc)
+    else:
+        raise Exception(
+            "username and password were not supplied. Integration tests will be skipped."
```
unparalleled-js, line 12: Could print a usage here.

`Usage:\n\tpython run_integration_tests.py <code42-username> <code42-password>

run_integration.py
```diff
@@ -0,0 +1,13 @@
+import os
```
unparalleled-js, line 1: I don't think this file's name is accurate.  `run_integration` sounds like it could be running like a Slack integration or something similar.

Should we call it `run_integration_tests.py` ?

integration/conftest.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+import pytest
+
+import py42.sdk as _sdk
+
+HOST_URL = "https://console.us.code42.com"
```
unparalleled-js, line 7: If we make this a parameter in `run_integration.py`, we can alternately choose to run the tests elsewhere more easily.

unparalleled-js:
I don't like that it require running a custom script to execute the integration tests.  `pytest` has capabilities to support custom markers for marking tests as a integration tests and executing them via `pytest -v -m integration`.

unparalleled-js:
We could also include custom markers for the tests that only run on-prem (or mockserver) and we can avoid using `skip`

unparalleled-js:
I think all of the tests should be testing off the sdk object

run_integration.py
unparalleled-js, line 1: So can this file be deleted?

integration/conftest.py
kiran-chaudhary: NA applicable anymore as the file is removed.

run_integration.py
kiran-chaudhary: Deleted in the other PR, where markers are added.

kiran-chaudhary: removed.

timabrmsn (approved):

PR #250, opened 2020-12-15
Integration tests alertrules

kiran-chaudhary:
### Description of Change ###

Added integration tests for alert rules.

### PR Checklist ###

### DIFF
--- a/integration/test_alertrules.py
+++ b/integration/test_alertrules.py
@@ -1,0 +1,60 @@
+import pytest
+
+
+@pytest.mark.skip("Changes state of the system.")
+def test_rules_add_user():
+    pass
+
+
+def test_rules_get_all(connection):
+    response_gen = connection.alerts.rules.get_all()
+    for response in response_gen:
+        assert response.status_code == 200
+        break
+
+
+def test_rules_get_by_observer_id(connection):
+    observer_id = "d52cbfe0-f9de-468e-afbe-3c91037322da"
+    response = connection.alerts.rules.get_by_observer_id(observer_id)
+    assert response.status_code == 200
+
+
+@pytest.mark.skip("Changes state of the system.")
+def test_rules_remove_all_users():
+    pass
+
+
+def test_rules_get_all_by_name(connection):
+    response_gen = connection.alerts.rules.get_all_by_name("Test Alerts using CLI")
+    for response in response_gen:
+        assert response.status_code == 200
+        break
+
+
+def test_rules_get_page(connection):
+    response = connection.alerts.rules.get_page()
+    assert response.status_code == 200
+
+
+@pytest.mark.skip("Changes state of the system.")
+def test_rules_remove_user():
+    pass
+
+
+def test_rules_exfiltration_get(connection):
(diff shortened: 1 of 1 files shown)


integration/test_alertrules.py
```diff
@@ -0,0 +1,68 @@
+import pytest
+
+
+@pytest.mark.skip("Failing with 404.")
```
unparalleled-js, line 4: Why is this failing?

unparalleled-js:
Did we ever figure out how we were going to ignore integrations tests for CI/CD ?

integration/test_alertrules.py
unparalleled-js, line 4: I think this is failing because you are rule id when you should be using the observer ID or visa-versa (I don't remember but I always do this).

integration/test_alertrules.py
```diff
@@ -0,0 +1,68 @@
...
+        assert response.status_code == 200
+        break
+
+
+def test_rules_get_page(connection):
+    response = connection.alerts.rules.get_page()
+    assert response.status_code == 200
+
+
+@pytest.mark.skip("Failing with 404.")
```
unparalleled-js, line 37: Same here- wrong ID, try using the other one

unparalleled-js (changes requested):
* Figure out the correct ID to use for alert rules. i think is the observer ID.
* Wait until the other PR merges first because it seems like this one depends on that one based on the fixture that gets used.

kiran-chaudhary:
> Did we ever figure out how we were going to ignore integrations tests for CI/CD ?

They are not added to the current workflow, so by default they are ignored when we execute `tox`.  We need to define when to execute them, during build-release stage or after merge to master or during creation of PR.

kiran-chaudhary:
> * Figure out the correct ID to use for alert rules. i think is the observer ID.
> * Wait until the other PR merges first because it seems like this one depends on that one based on the fixture that gets used.

I have and plan to create all other PRs against the "fixture" based PR, so all of them will merge into that PR and then that would be merged to master.

integration/test_alertrules.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def test_rules_get_page(connection):
+    response = connection.alerts.rules.get_page()
+    assert response.status_code == 200
+
+
+def test_rules_remove_user(connection):
+    response = connection.alerts.rules.remove_user(
+        "d52cbfe0-f9de-468e-afbe-3c91037322da", 912249223544144039
```
unparalleled-js, line 35: We should make the real IDs (not the legacy IDs, but IDs like this one right here) `str` because that is what the API returns and so it is more expected.

Plus I think from a memory standpoint, the long integer have predefined set of bits (64?) and strings are just a 1 byte per character, not like memory would ever be an issue, but it seems like a better use of the data types anyway

unparalleled-js: Not really a big deal though! Just more expected

unparalleled-js (approved):

integration/test_alertrules.py
kiran-chaudhary: changed.

PR #251, opened 2020-12-15
Do not expose legal hold backup sets in DeviceSettings objects

### LINKED ISSUE #248: [Bug] sdk.devices.get_settings fails if there is a single locked destination
### Description

when running `sdk.devices.get_settings()`, I encountered a stacktrace thrown by ../py42/clients/settings/device_settings.py, line 357. 

### Steps to Reproduce

1. Create a device in Code42.
2. Assign the device a single, locked, destination
3. Attempt to retrieve the device's settings using py42

### Expected Behavior
The settings data retrieves and parses correctly.

### Actual Behavior
My device has this `destinations` object in its settings JSON:

```
              "destinations": {
                "@locked": "true",
                "destination": {
                  "@id": "632540230984925185",
                  "@locked": "true"
                }
```

py42 is attempting to process it like this:

```
for d in self.data[u"destinations"]:
            guid = d[u"@id"]
            dest_name = self._manager.available_destinations[guid]
            if u"@locked" in d:
                dest_name = dest_name + u" <LOCKED>"
            destination_dict[guid] = dest_name
```

This throws a stack trace because `d[u"@id"]` does not exist; since self.data[u"destinations"] is a dict and not a list.

### Basic Information

- py42 version: 1.10.0
- python version: 3.7.4
- operating system: Mac OS 10.15.7

timabrmsn:
### Description of Change ###

The `py42.clients.settings.device_settings.BackupSet` object has a bug where it fails to handle the destinations object for legal hold backup sets, because the resolution of the `BackupSet.destinations` property is expecting a list and a legal hold set turns it into a dict so it can add the `@locked: true` property to it. 

Since legal hold backup sets are hidden from every part of the Code42 UI (both console and client), and since they should never be individually modified via API (even if some API changes are _technically_ possible to make), because of the confusion a mismatch between an individual device and the legal hold matter policy that device is part of would create, this change just hides the legal hold backup set(s) altogether from the `DeviceSettings` object's public interfaces. 

### Issues Resolved ###
- closes #248

### Testing Procedure ###
- Add a Code42 user/device to a legal hold matter
- Create a py42 `DeviceSettings` object for the device
- Verify `DeviceSettings` object creation does not throw exception like previous versions
- Verify `DeviceSettings.backup_sets` property only shows the backup set(s) visible in the console for that device

### PR Checklist ###

### DIFF
--- a/src/py42/clients/settings/device_settings.py
+++ b/src/py42/clients/settings/device_settings.py
@@ -39,11 +39,19 @@ class DeviceSettingsDefaults(UserDict, object):
             if isinstance(backup_sets, dict):  # there's only one set configured
                 return [BackupSet(self, backup_sets)]
             elif isinstance(backup_sets, list):
-                return [BackupSet(self, bs) for bs in backup_sets]
+                return [
+                    BackupSet(self, bs)
+                    for bs in backup_sets
+                    if not _backup_set_is_legal_hold(bs)
+                ]
             else:
                 raise Py42Error("Unable to extract backup sets: {}".format(backup_sets))
         else:
-            return [BackupSet(self, bs) for bs in backup_sets]
+            return [
+                BackupSet(self, bs)
+                for bs in backup_sets
+                if not _backup_set_is_legal_hold(bs)
+            ]
 
     @property
     def available_destinations(self):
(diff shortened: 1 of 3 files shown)


src/py42/clients/settings/device_settings.py
```diff
@@ -500,3 +508,15 @@ def pop(self, index=-1):
     def remove(self, value):
         self.data.remove(value)
         self.register_change()
+
+
+def _backup_set_is_legal_hold(bs):
```
unparalleled-js, line 513: Could there ever be other reasons for this besides legal hold?

The criteria for the conditional seems generic, that is why I ask.

`_backup_set_has_locked_destinations()` ?

unparalleled-js (approved):

src/py42/clients/settings/device_settings.py
timabrmsn: Nope, for both the individual device and the org device defaults, it's only possible to lock individual destinations in the UI. Locking "destinations" as a parent category only happens for legal hold. 

It's maybe possible to craft custom json to set the destinations parent as locked?

PR #252, opened 2020-12-16
Integration tests detectionlists

kiran-chaudhary:
### Description of Change ###

Added integration tests for detectionlists.

### PR Checklist ###

### DIFF
--- a/integration/test_detectionlists.py
+++ b/integration/test_detectionlists.py
@@ -2,10 +2,10 @@ from datetime import datetime
 from datetime import timedelta
 
 
-new_user = "integraiton_" + str(int(datetime.today().timestamp())) + "@test.com"
+new_user = "integration_" + str(int(datetime.now().timestamp())) + "@test.com"
 existing_user = "test1-test@test.com"
 alias_user = "test_user@test.com"
-user_departure_date = datetime.today() + timedelta(days=10)
+user_departure_date = datetime.now() + timedelta(days=10)
 user_uid = 977335752891390447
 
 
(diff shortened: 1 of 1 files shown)


integration/test_detectionlists.py
```diff
@@ -0,0 +1,125 @@
+from datetime import datetime
+from datetime import timedelta
+
+
+new_user = "integraiton_" + str(int(datetime.today().timestamp())) + "@test.com"
```
unparalleled-js, line 5: `integration`

unparalleled-js: The date here does not need to be localized, therefore it makes more sense to use the `now()` function rather than `today()`

https://stackoverflow.com/questions/32517248/what-is-the-difference-between-python-functions-datetime-now-and-datetime-t

unparalleled-js: For the sake of a random generated ID, it does not to be localized, so you can use the `now()` function over the `today()` function off `datetime`.

For the departure date, it does make sense for it to be localized though

kiran-chaudhary: changed both and fixed spelling.

integration/test_detectionlists.py
```diff
@@ -0,0 +1,125 @@
+from datetime import datetime
+from datetime import timedelta
+
+
+new_user = "integration_" + str(int(datetime.now().timestamp())) + "@test.com"
```
unparalleled-js, line 5: this one should probably be utcnow

unparalleled-js (approved):

integration/test_detectionlists.py
kiran-chaudhary: changed.

PR #255, opened 2020-12-18
Integration tests users

kiran-chaudhary:
### Description of Change ###

Added integration tests for users, user context and serveradmin features.

### Issues Resolved ###

- closes #

### PR Checklist ###

- [ Na] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/integration/test_serveradmin.py
+++ b/integration/test_serveradmin.py
@@ -1,7 +1,7 @@
 import pytest
 
 
-@pytest.mark.skip("Failing, requires administrator credentials.")
+@pytest.mark.skip("Requires on prem connection.")
 def test_get_alert_log(connection):
     response = connection.serveradmin.get_alert_log()
     assert response.status_code == 200
@@ -12,7 +12,7 @@ def test_get_current_tenant(connection):
     assert response.status_code == 200
 
 
-@pytest.mark.skip("Failing, requires administrator credentials.")
+@pytest.mark.skip("Requires on prem connection..")
 def test_get_diagnostics(connection):
     response = connection.serveradmin.get_diagnostics()
     assert response.status_code == 200
--- a/integration/test_serveradmin.py
+++ b/integration/test_serveradmin.py
@@ -1,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.skip("Failing with 403.")
+def test_get_alert_log(connection):
+    response = connection.serveradmin.get_alert_log()
+    assert response.status_code == 200
+
+
+def test_get_current_tenant(connection):
+    response = connection.serveradmin.get_current_tenant()
+    assert response.status_code == 200
+
+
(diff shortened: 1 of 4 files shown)


integration/test_serveradmin.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+
+@pytest.mark.skip("Failing, requires administrator credentials.")
```
kiran-chaudhary, line 4: The reason to skip this, is to highlight the case that in case we ever execute tests for real (not against mock server) then what kind of user credentials should we pass. 
To make sure all the necessary conversations take place.

integration/test_usercontext.py
```diff
@@ -0,0 +1,3 @@
+def test_get_current_tenant(connection):
+    response = connection.usercontext.get_current_tenant_id()
+    assert response == "1d71796f-af5b-4231-9d8e-df6434da4663"
```
kiran-chaudhary, line 3: All though this is a test system, Does exposing the `tenant-id` have any risk, given that Py42 is a publicly accessible?

This endpoint does not return `Py42Response`.

integration/test_serveradmin.py
unparalleled-js: I think this is an on-prem-only endpoint. So you could put `Required on-prem connection` and we can save it for the mock server

integration/test_usercontext.py
unparalleled-js: You could do

```
assert type(response) == str
assert len(response) == 36
```

That should cover any tenant ID

integration/test_users.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def test_get_scim_data_by_uid(connection):
+    response = connection.users.get_scim_data_by_uid(user_uid)
+    assert response.status_code == 200
+
+
+def test_change_org_assignment(connection):
+    response = connection.users.change_org_assignment(user_id, update_org_id)
+    assert response.status_code == 204
```
unparalleled-js, line 66: Might be worth having a helper method `assert_successful_response(response)` that does this

```
assert response.status_code >= 200
assert response.status_code < 300
```

unparalleled-js: I don't think it matter that it is 204, it just matter that it was successful

kiran-chaudhary: and I was thinking to raise a concern that we should make a note for our API standardisation, I have come across `create` aka `post` methods that return 200 but they should be 201.

integration/test_usercontext.py
kiran-chaudhary: Changed.

integration/test_serveradmin.py
kiran-chaudhary: Changed.

PR #256, opened 2020-12-18
Integration tests legalhold

kiran-chaudhary:
### Description of Change ###

Added integration tests for legalhold feature.

### PR Checklist ###

### DIFF
--- a/integration/test_legalhold.py
+++ b/integration/test_legalhold.py
@@ -1,16 +1,47 @@
+from datetime import datetime
+
 import pytest
 
-SKIP_TEST_MESSAGE = "Changes system state."
+user_uid = 977335752891390447
+policy_uid = 894163190167359978941631901673599755
+timestamp = str(int(datetime.now().timestamp()))
+matter_name = "integration test matter {}".format(timestamp)
+policy_name = "integration test policy {}".format(timestamp)
 
 
-@pytest.mark.skip(SKIP_TEST_MESSAGE)
-def test_add_to_matter():
-    pass
+@pytest.fixture(scope="module")
+def policy(connection):
+    response = connection.legalhold.create_policy(policy_name)
+    assert response.status_code == 200
+    return response["legalHoldPolicyUid"]
 
 
-@pytest.mark.skip("Test failing")
-def test_get_all_matter_custodians(connection):
-    response_gen = connection.legalhold.get_all_matter_custodians()
+@pytest.fixture(scope="module")
+def matter(connection, policy):
+    response = connection.legalhold.create_matter(matter_name, policy)
+    assert response.status_code == 201
+    return response["legalHoldUid"]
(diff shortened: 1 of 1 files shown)


kiran-chaudhary:
Created draft PR because of failing tests.

integration/test_legalhold.py
```diff
@@ -0,0 +1,85 @@
...
+
+user_uid = 977335752891390447
+policy_uid = 894163190167359975
+timestamp = str(int(datetime.now().timestamp()))
+matter_name = "integration test matter {}".format(timestamp)
+policy_name = "integration test policy {}".format(timestamp)
+
+
+@pytest.fixture(scope='module')
+def test_create_policy(connection):
```
kiran-chaudhary, line 12: Should this be renamed to `legalholduid`? Needed this to make sure a new policy is created for every test execution. 

Had to create it as a fixture to get the new id in other methods. However, pytest doesn't consider a fixture as a test.

integration/test_legalhold.py
```diff
@@ -0,0 +1,85 @@
...
+
+user_uid = 977335752891390447
+policy_uid = 894163190167359978941631901673599755
+timestamp = str(int(datetime.now().timestamp()))
+matter_name = "integration test matter {}".format(timestamp)
+policy_name = "integration test policy {}".format(timestamp)
+
+
+@pytest.fixture(scope="module")
+def policy(connection):
```
kiran-chaudhary, line 13: This covers the create_policy test case.

kiran-chaudhary:
added some test cases as fixtures, so that we can use the return values in other tests.

integration/test_legalhold.py
```diff
@@ -0,0 +1,85 @@
...
+
+def test_get_all_matters(connection):
+    response_gen = connection.legalhold.get_all_matters()
+    for response in response_gen:
+        assert response.status_code == 200
+        break
+
+
+@pytest.mark.skip("Failing with 400 bad request error")
+def test_get_custodians_page(connection):
```
unparalleled-js, line 68: This is not a client method...

We should probably only be testing the client methods for integration tests

unparalleled-js, line 68: So this method needs either of one of it's parameters - it does do global search for some reason.

It needs one of these:
`legalHoldUid`
`userUid`
`user` (username)

unparalleled-js (changes requested):
have to use 1 of the params for the failing test

integration/test_legalhold.py
kiran-chaudhary: Fixed, can we make a note that necessary action can to be taken here. Either we document it or we fix the behaviour, or may be in version 2 we make one of them compulsory.

unparalleled-js (approved):

PR #259, opened 2020-12-21
Integration tests orgs

kiran-chaudhary:
### Description of Change ###

Add orgs integration tests.

### PR Checklist ###

### DIFF
--- a/integration/test_orgs.py
+++ b/integration/test_orgs.py
@@ -1,20 +1,29 @@
+from datetime import datetime
+
 import pytest
 
-SKIP_TEST_MESSAGE = "Changes system state."
+from py42.exceptions import Py42ForbiddenError
+from py42.exceptions import Py42InternalServerError
 
 
-@pytest.mark.skip(SKIP_TEST_MESSAGE)
-def test_block():
-    pass
+timestamp = str(int(datetime.now().timestamp()))
+new_org = "integration test org {}".format(timestamp)
+org_id = 2689
+org_uid = 890854247383106706
+
+
+def test_create_org(connection):
+    with pytest.raises(Py42ForbiddenError):
+        connection.orgs.create_org(new_org)
 
 
 def test_get_agent_full_disk_access_states(connection):
-    response = connection.orgs.get_agent_full_disk_access_states(2689)
+    response = connection.orgs.get_agent_full_disk_access_states(org_id)
     assert response.status_code == 200
 
 
 def test_get_by_id(connection):
-    response = connection.orgs.get_by_id(2689)
+    response = connection.orgs.get_by_id(org_id)
     assert response.status_code == 200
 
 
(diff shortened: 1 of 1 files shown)


integration/test_orgs.py
```diff
@@ -0,0 +1,73 @@
+from datetime import datetime
+
+import pytest
+
+from py42.exceptions import Py42ForbiddenError
+from py42.exceptions import Py42InternalServerError
+
+
+timestamp = str(int(datetime.now().timestamp()))
```
unparalleled-js, line 9: We should use `utcnow()` instead of `now()`, sorry I know I said `now()` on another PR but I meant `utcnow()`

integration/test_orgs.py
```diff
@@ -0,0 +1,73 @@
...
+    response = connection.orgs.get_agent_state(org_id, "fullDiskAccess")
+    assert response.status_code == 200
+
+
+def test_get_by_uid(connection):
+    response = connection.orgs.get_by_uid(org_uid)
+    assert response.status_code == 200
+
+
+def test_deactivate(connection):
```
unparalleled-js, line 45: This is a good way to deal with this... But I am just wondering if we should have a special marker for the operations that we don't expect to work for the prod environment but that we can make work in the server environment, we could have real responses there...

timabrmsn: Yeah, this is pretty dependent on the role of the user running the tests. If you happen to pass in the credentials of an org_admin or higher, this deactivate (and the `create()` above will actually succeed). 

Hard-coding with a specific org_id/org_uid also means if we want to run against a different env we'll need to manually change the code. 

I wonder if the integration tests should assume an admin user (fail fast if not), and allow for the `create()` to succeed first, then use the ID of that org from the response for any future actions?

integration/test_orgs.py
kiran-chaudhary: changed.

integration/test_orgs.py
kiran-chaudhary: - The reason it was not assumed that all the tests will be run with `administrator` like privileges is because for such special cases the needed communication takes place, so that we can cover all cases possible. Yesterday discussing with @unparalleled-js we talked about `pytest markers` and may be we can use that in this case as well. While executing the script appropriate markers will be specified  and the tests will be executed with apt credentials against the markers. Before going ahead with this , we need to make sure the automation pipeline will handle that.

- on hard coded ids,   in all other features, the "create resource" task is done in a fixture and that id is used in subsequent tests, same can be done here if all the subsequent tests are executed with a user with `administrator` privileges or we create a user that can then call the subsequent tests, both approaches have different trade-offs to consider.

- Now that we are discussing the special cases, even in archives there is a hard-coded path, either we invest in creating setup/environment that can be standalone or we can mark certain tests to be executed in specific environments.

kiran-chaudhary: So looking at all other hardcoded values, user_id is the most common, this can be fixed by creating a fixture that creates user and that fixture can be used in all other tests. Still we won't be able to pass the administrative privileges to the new user. However some tests in archives, orgs and security-data might need environment setup or privileges.

integration/test_orgs.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+timestamp = str(int(datetime.utcnow().timestamp()))
+new_org = "integration test org {}".format(timestamp)
+org_id = 2689
+org_uid = 890854247383106706
+
+
+def test_create_org(connection):
+    with pytest.raises(Py42ForbiddenError):
```
kiran-chaudhary, line 16: For a long time I assumed this test was failing because the user I was using didn't have administrator privileges. However that is not the case. From my analysis, it seems I can create the organisation when I pass `parent_org_id`. I would like to understand is that ok, before I proceed and make the changes?

integration/test_orgs.py
```diff
@@ -0,0 +1,73 @@
...
+
+timestamp = str(int(datetime.utcnow().timestamp()))
+new_org = "integration test org {}".format(timestamp)
+
+
+@pytest.fixture(scope="module")
+def org(connection):
+    orgs_gen = connection.orgs.get_all()
+    orgs = next(orgs_gen)
+    # Assumption: Parent org always exists.
```
kiran-chaudhary, line 13: Even the parent org has a `parent_org_id` and `parent_org_uid`, which one would assume it won't exist for the first org, for some reason they do and hence had to make the assumption.

integration/test_orgs.py
unparalleled-js: I was only referring to his actual name, not the user ids

unparalleled-js: It is a matter of courtesy

unparalleled-js (approved):

timabrmsn (approved):

PR #262, opened 2020-12-22
Chore/clarify legal hold get custodians doc

timabrmsn:
### Description of Change ###

The backend for `sdk.legalhold.get_custodians_page()` requires at least one of the optional params `legal_hold_membership_uid`, `legal_hold_uid`, `user_uid`, `user` or else it returns a 400 error. 

This change documents that requirement as well as raises a `Py42Error` when none of the the options are passed. 

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- `sdk.legalhold.get_custodians_page()` now raises `Py42Error` when missing on of the required options.
+
 ### Added
 
 - Python 3.9 support.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,9 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
-- `sdk.legalhold.get_custodians_page()` now raises `Py42Error` when missing on of the required options.
+### Changed
+
+- `sdk.legalhold.get_custodians_page()` now raises `Py42LegalHoldCriteriaMissingError` when missing one of the required options.
 
 ### Added
 
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -180,7 +180,7 @@ class Py42LegalHoldCriteriaMissingError(Py42BadRequestError):
     def __init__(self, exception):
         super(Py42LegalHoldCriteriaMissingError, self).__init__(
             exception,
-            u"At least one criteria must be specified; hold_membership_uid, hold_uid, user_uid, or user",
+            u"At least one criteria must be specified; hold_membership_uid, hold_uid, user_uid, or user.",
         )
 
 
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -174,6 +174,16 @@ class Py42LegalHoldNotFoundOrPermissionDeniedError(Py42ForbiddenError):
         )
 
 
+class Py42LegalHoldCriteriaMissingError(Py42BadRequestError):
+    """An exception raised when a bad request was made to a Legal Hold endpoint."""
+
+    def __init__(self, exception):
+        super(Py42LegalHoldCriteriaMissingError, self).__init__(
+            exception,
+            u"At least one criteria must be specified; hold_membership_uid, hold_uid, user_uid, or user",
+        )
+
+
 class Py42InvalidRuleOperationError(Py42HTTPError):
     """An exception raised when trying to add or remove users to a system rule."""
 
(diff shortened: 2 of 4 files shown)


src/py42/services/legalhold.py
```diff
@@ -215,6 +219,10 @@ def get_custodians_page(
         Returns:
             :class:`py42.response.Py42Response`:
         """
+        if not any((legal_hold_membership_uid, legal_hold_uid, user_uid, user)):
```
unparalleled-js, line 222: Is there an HTTP error that occurs that is specific when none of these are received?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ## Unreleased
 
+- `sdk.legalhold.get_custodians_page()` now raises `Py42Error` when missing on of the required options.
```
unparalleled-js, line 13: Missing a second level header... `Changed` ?

src/py42/services/legalhold.py
unparalleled-js: I ask only because it might nice to let it occur and use it as the base class so it works similarly to before, but if that is not the case, this is probably better

timabrmsn: The error response is `Py42BadRequestError` and does have a specific error message:

```
'[{"name":"SYSTEM","description":"Invalid request: SYSTEM com.code42.core.BuilderException: At least one criteria must be specified; holdMembershipUid, holdUid, userUid, or userSearch"}]'
```

I reworked it to be a custom exception inheriting from Py42BadRequestError so anyone depending on that behavior won't run into problems.

src/py42/services/legalhold.py
```diff
@@ -227,7 +232,13 @@ def get_custodians_page(
             u"pgSize": page_size,
         }
         uri = u"/api/LegalHoldMembership"
-        return self._connection.get(uri, params=params)
+        try:
+            return self._connection.get(uri, params=params)
+        except Py42BadRequestError as ex:
+            if u"At least one criteria must be specified" in ex.response.text:
+                raise Py42LegalHoldCriteriaMissingError(ex)
+            else:
```
unparalleled-js, line 240: else is not needed

src/py42/exceptions.py
```diff
@@ -174,6 +174,16 @@ def __init__(self, exception, matter_id):
         )
 
 
+class Py42LegalHoldCriteriaMissingError(Py42BadRequestError):
+    """An exception raised when a bad request was made to a Legal Hold endpoint."""
+
+    def __init__(self, exception):
+        super(Py42LegalHoldCriteriaMissingError, self).__init__(
+            exception,
+            u"At least one criteria must be specified; hold_membership_uid, hold_uid, user_uid, or user",
```
unparalleled-js, line 183: missing .

unparalleled-js (approved):
couple of minor things but looks goood

tests/services/test_legalhold.py
```diff
@@ -123,7 +124,7 @@ def test_get_all_matter_custodians_calls_get_expected_number_of_times(
             mock_get_all_matter_custodians_response,
             mock_get_all_matter_custodians_empty_response,
         ]
-        for _ in service.get_all_matter_custodians():
+        for _ in service.get_all_matter_custodians(user="test"):
```
kiran-chaudhary, line 127: for some reason I am still not convinced how this works, when we call `get_all_matter_custodians` without any optional argument, it passes but individually calling `get_custodians_page` fails, now if `get_custodians_page` needs one of the optional field so `get_all_matter_custodians` too need to have same behaviour.

kiran-chaudhary: I mean it passes in the integration tests, and here the change has been done on `get_all_matter_custodians` method instead of `get_custodians_page`.

timabrmsn: The `user="test"` param is no longer strictly necessary in this test, as initially I had been raising an exception when all of the criteria were `None` (before any http requests were made), so here we needed to pass in _some_ option. 

I haven't looked at the integration test yet, but running `next(sdk.legalhold.get_all_matter_custodians())` in my test org raises the 400 BadRequestError (which we've now replaced with the custom `Py42LegalHoldCriteriaMissingError`.

timabrmsn: Ah, your test does pass in two arguments: 

```
def test_get_all_matter_custodians(connection, policy, matter):
     response_gen = connection.legalhold.get_all_matter_custodians(policy, matter)
     for response in response_gen:
         assert response.status_code == 200
         break
```

Which if they're valid input values, will do a search for those Ids. It returns a 200, but no results:

```
next(sdk.legalhold.get_all_matter_custodians("1", "1"))
<Py42Response [status=200, data={'legalHoldMemberships': []}]>
```

PR #263, opened 2020-12-23
Integration marker and configure data

kiran-chaudhary:
### Description of Change ###

Made tests environment independent and/or configurable.
Added integration test marker.

### PR Checklist ###

- [NA ] Add unit tests to verify this change


integration/test_alertrules.py
```diff
@@ -1,57 +1,54 @@
-rule_id = "1cae9f92-5fd7-4504-b363-9bc45015adaa"
-observer_id = "d52cbfe0-f9de-468e-afbe-3c91037322da"
-user_uid = "912249223544144039"
+import pytest
 
 
-def test_rules_add_user(connection):
```
kiran-chaudhary, line 6: all the methods that are removed here are put into a class.

integration/conftest.py
```diff
@@ -1,14 +1,69 @@
...
+    return request.config.getini("observer_rule_id")
+
+
+@pytest.fixture(scope="session")
+def connection(host):
+    return _sdk.from_local_account(host, os.environ["C42_USER"], os.environ["C42_PW"])
+
+
+@pytest.fixture(scope='session')
+def timestamp():
```
kiran-chaudhary, line 42: had to create this fixture as py2.7 was complaining.

kiran-chaudhary:
Reviewing this would be complex due to tab/space changes. Major changes are in `tox.ini` and `conftest.py`.

integration/test_detectionlists.py
```diff
@@ -1,125 +1,120 @@
...
-    assert response.status_code == 200
+    def test_departing_employee_update_departure_date(self, connection, new_user):
+        response = connection.detectionlists.departing_employee.update_departure_date(
+            new_user["userUid"], user_departure_date
+        )
+        assert response.status_code == 200
 
+    def test_add_user_risk_tags(self, connection, new_user):
+        response = connection.detectionlists.add_user_risk_tags(
+            new_user["userUid"], "Flight_Risk"
```
kiran-chaudhary, line 78: Is `Flight_Risk` a system specific data, if so then need to make it configurable, please confirm.

integration/conftest.py
```diff
@@ -1,14 +1,69 @@
 import os
+from datetime import datetime
 
 import pytest
 
 import py42.sdk as _sdk
 
-HOST_URL = "https://console.us.code42.com"
+
+def pytest_addoption(parser):
```
unparalleled-js, line 9: This looks like `adoption` with two `d`s... Can say `add_option` ?

tox.ini
```diff
@@ -18,7 +18,8 @@ commands =
     # -l: show locals in tracebacks
     # --tb=short: short traceback print mode
     # --strict: marks not registered in configuration file raise errors
-    pytest --cov=py42 --cov-report xml --cov-report term -v -rsxX -l --tb=short --strict
+    # -m="not integration": exclude tests that are marked as integration
+    pytest --cov=py42 --cov-report xml --cov-report term -v -rsxX -l --tb=short --strict -m"not integration"
```
unparalleled-js, line 22: Do you need a space between the `m` and the `"not integration"` ?

integration/test_detectionlists.py
unparalleled-js: We have the enum values defined in `py42.clients.detectionlists.RiskTags`, I think it's better to use the enum values.

I don't think it needs to be configurable, but this method accepts `str or list of str`, so we can give it multiple in one go, if you want to try it out.

integration/conftest.py
kiran-chaudhary: This is a pytest hook, and the name has to be `pytest_addoption`. Otherwise it won't read the config file.

tox.ini
kiran-chaudhary: Changed.

integration/test_detectionlists.py
kiran-chaudhary: Changed.

integration/conftest.py
timabrmsn: Can we check in an ini file with these values blank so it's clear where to put it?

Maybe also add a section to the contributing guide about integration tests, how to run them, how to skip them, etc.

kiran-chaudhary: - All values are already added to ini file. 

- Also the change to skip integration tests is also covered in ini file. So right now unit test execution will behave as it behaves in master. i.e integration tests will be skipped.

-I will make a note in contributinging guide to add username and password in appropriate environment variables.

timabrmsn: Ah, sorry. I was looking for a pytest.ini file, and hadn't yet got to the tox.ini file part of the review.

kiran-chaudhary: added notes to contributing guide.

timabrmsn (approved):

PR #264, opened 2021-01-04
Feature cases

kiran-chaudhary:
### Description of Change ###

Add support for cases API.

### Issues Resolved ###

- closes INTEG-1191

### Testing Procedure ###

- `sdk.cases` methods:
    - `sdk.cases.create()`
    - `sdk.cases.get_all()`
    - `sdk.cases.export_summary()`
    - `sdk.cases.get`
    - `sdk.case.update()`

- `sdk.cases.file_events` methods for managing cases file events:
    - `sdk.cases.file_events.add`
    - `sdk.cases.file_events.delete`
    - `sdk.cases.file_events.get`
    - `sdk.cases.file_events.get_all_events`

### PR Checklist ###

- [ Y] Add unit tests to verify this change
- [ Y] Add an entry to CHANGELOG.md describing this change
- [ Y] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,10 +22,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.cases.create()`
     - `sdk.cases.get_all()`
     - `sdk.cases.export_summary()`
-    - `sdk.cases.get_case_by_case_number()`
+    - `sdk.cases.get_case()`
     - `sdk.case.update()`
 
-- `sdk.cases.file_events` methods for managing cases file events:
+- `sdk.cases.file_events` methods for managing file events associated with a given case:
     - `sdk.cases.file_events.add_event`
     - `sdk.cases.file_events.delete_event`
     - `sdk.cases.file_events.get_event`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,13 +22,13 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.cases.create()`
     - `sdk.cases.get_all()`
     - `sdk.cases.export_summary()`
-    - `sdk.cases.get_case()`
+    - `sdk.cases.get()`
     - `sdk.case.update()`
 
 - `sdk.cases.file_events` methods for managing file events associated with a given case:
-    - `sdk.cases.file_events.add_event()`
-    - `sdk.cases.file_events.delete_event()`
-    - `sdk.cases.file_events.get_event()`
+    - `sdk.cases.file_events.add()`
+    - `sdk.cases.file_events.delete()`
+    - `sdk.cases.file_events.get()`
     - `sdk.cases.file_events.get_all()`
 
 ## 1.10.1 - 2020-12-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,10 +26,10 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
     - `sdk.case.update()`
 
 - `sdk.cases.file_events` methods for managing file events associated with a given case:
-    - `sdk.cases.file_events.add_event`
-    - `sdk.cases.file_events.delete_event`
-    - `sdk.cases.file_events.get_event`
-    - `sdk.cases.file_events.get_all_events`
+    - `sdk.cases.file_events.add_event()`
+    - `sdk.cases.file_events.delete_event()`
+    - `sdk.cases.file_events.get_event()`
+    - `sdk.cases.file_events.get_all()`
 
 ## 1.10.1 - 2020-12-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - Python 3.9 support.
 
+- `sdk.cases` methods:
+    - `sdk.cases.create()`
+    - `sdk.cases.get_all()`
+    - `sdk.cases.export_summary()`
+    - `sdk.cases.get_case_by_case_number()`
+    - `sdk.case.update()`
+
+- `sdk.cases.file_events` methods for managing cases file events:
+    - `sdk.cases.file_events.add_event`
+    - `sdk.cases.file_events.delete_event`
+    - `sdk.cases.file_events.get_event`
+    - `sdk.cases.file_events.get_all_events`
+
 ## 1.10.1 - 2020-12-16
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 - `sdk.cases` methods:
     - `sdk.cases.create()`
     - `sdk.cases.get_all()`
+    - `sdk.cases.get_page()`
     - `sdk.cases.export_summary()`
     - `sdk.cases.get()`
     - `sdk.case.update()`
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -1,5 +1,7 @@
 from enum import Enum
 
+from py42.util import parse_timestamp_to_milliseconds_precision
+
 
 class CaseStatus(Enum):
     OPEN = u"OPEN"
@@ -48,12 +50,22 @@ class CasesClient(object):
             findings=findings,
         )
 
+    @staticmethod
+    def _make_range(begin_time, end_time):
+        if not begin_time or not end_time:
+            return None
+        end = parse_timestamp_to_milliseconds_precision(end_time)
+        start = parse_timestamp_to_milliseconds_precision(begin_time)
+        return "{}/{}".format(start, end)
+
     def get_all(
         self,
         name=None,
         status=None,
-        created_at=None,
-        updated_at=None,
+        created_at_begin_time=None,
+        created_at_end_time=None,
+        updated_at_begin_time=None,
+        updated_at_end_time=None,
         subject=None,
         assignee=None,
         page_number=1,
@@ -68,10 +80,14 @@ class CasesClient(object):
             name (str, optional): Filter results by case name, matches partial names. Defaults to None.
             status (enum, optional): Filter results by case status. Defaults to None.
                 e.g `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
-            created_at (str, optional): Filter results by case creation time range, format ISO interval.
-                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
-            updated_at (str, optional): Filter results by last updated time range, format ISO interval.
-                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            created_at_begin_time (str or int or float or datetime, optional): Filter results by case creation time, start time.
+                Defaults to None. Begin time should be specified with end time. str format %Y-%m-%d %H:%M:%S
+            created_at_end_time (str, optional): Filter results by case creation time, end time.
+                Defaults to None. End time should be specified with begin time. str format %Y-%m-%d %H:%M:%S
+            updated_at_begin_time (str, optional): Filter results by last updated time, start time.
+                Defaults to None. Begin time should be specified with end time. str format %Y-%m-%d %H:%M:%S
+            updated_at_end_time (str, optional): Filter results by last updated time, end time.
+                Defaults to None. End time should be specified with begin time. str format %Y-%m-%d %H:%M:%S
             subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
             assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
             page_number (int, optional): Page number of the results. Defaults to 1.
@@ -85,6 +101,9 @@ class CasesClient(object):
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
             that each contain a page of cases.
         """
+        created_at = CasesClient._make_range(created_at_begin_time, created_at_end_time)
+        updated_at = CasesClient._make_range(updated_at_begin_time, updated_at_end_time)
+
         return self._cases_service.get_all(
             name=name,
             status=status,
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -52,16 +52,19 @@ class CasesClient(object):
 
         Args:
             name (str, optional): Filter results by case name, matches partial names. Defaults to None.
-            status (str, optional): Filter results by case status, "OPEN" or "CLOSED". Defaults to None.
-            created_at (str, optional): Filter results by case creation time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
-            updated_at (str, optional): Filter results by last updated time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            status (str, optional): Filter results by case status, `OPEN` or `CLOSED`. Defaults to None.
+            created_at (str, optional): Filter results by case creation time range, format ISO interval.
+            Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            updated_at (str, optional): Filter results by last updated time range, format ISO interval.
+            Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
             subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
             assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
             page_number (int, optional): Page number of the results. Defaults to 1.
             page_size (int, optional): Number of results to return per page. Defaults to 100.
-            sort_direction (str, optional): The direction on which to sort the response, based on the corresponding sort key. "asc" or "desc" Defaults to "asc".
-            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number". Available options are
-             name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername
+            sort_direction (str, optional): The direction on which to sort the response,
+            based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
+            Available options are name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername.
 
         Returns:
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -17,15 +17,17 @@ class CasesClient(object):
         """
         return self._file_events
 
-    def create(self, name, subject, assignee, description, findings):
+    def create(
+        self, name, subject=None, assignee=None, description=None, findings=None
+    ):
         """Creates a new case.
 
         Args:
             name (str): Name of the case.
-            subject (str): User UID of a subject of a case.
-            assignee (str): User UID of the assignee.
-            description (str): Description of the case
-            findings (str): Observations of the case.
+            subject (str, optional): User UID of a subject of a case.
+            assignee (str, optional): User UID of the assignee.
+            description (str, optional): Description of the case
+            findings (str, optional): Observations of the case.
 
         Returns
             :class:`py42.response.Py42Response`
@@ -95,7 +97,13 @@ class CasesClient(object):
         return self._cases_service.export_summary(case_number)
 
     def update(
-        self, case_number, name="", subject="", assignee="", description="", findings=""
+        self,
+        case_number,
+        name=None,
+        subject=None,
+        assignee=None,
+        description=None,
+        findings=None,
     ):
         """Updates case details for the given case number.
 
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -1,8 +1,9 @@
 class CasesClient(object):
     """A client to expose cases API.
 
-    `Rest documenation <https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases>`__ 
+    `Rest documenation <https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases>`__
     """
+
     def __init__(self, cases_service, cases_file_event_service):
         self._cases_service = cases_service
         self._file_events = cases_file_event_service
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -33,7 +33,11 @@ class CasesClient(object):
             :class:`py42.response.Py42Response`
         """
         return self._cases_service.create(
-            name, subject, assignee, description, findings
+            name,
+            subject=subject,
+            assignee=assignee,
+            description=description,
+            findings=findings,
         )
 
     def get_all(
@@ -116,6 +120,7 @@ class CasesClient(object):
         assignee=None,
         description=None,
         findings=None,
+        status=None,
     ):
         """Updates case details for the given case number.
 
@@ -126,10 +131,16 @@ class CasesClient(object):
             assignee (str, optional): User UID of the assignee. Defaults to empty string.
             description (str, optional): Description of the case. Defaults to empty string.
             findings (str, optional): Notes on the case. Defaults to empty string.
-
+            status (str, optional): Status of the case.
         Returns:
             :class:`py42.response.Py42Response`
         """
         return self._cases_service.update(
-            case_number, name, subject, assignee, description, findings
+            case_number,
+            name=name,
+            subject=subject,
+            assignee=assignee,
+            description=description,
+            findings=findings,
+            status=status,
         )
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -67,6 +67,65 @@ class CasesClient(object):
         start = parse_timestamp_to_milliseconds_precision(begin_time)
         return "{}/{}".format(start, end)
 
+    def get_page(
+        self,
+        page_num,
+        name=None,
+        status=None,
+        min_create_time=None,
+        max_create_time=None,
+        min_update_time=None,
+        max_update_time=None,
+        subject=None,
+        assignee=None,
+        page_size=100,
+        sort_direction=u"asc",
+        sort_key=u"number",
+        **kwargs
+    ):
+        """Gets individual page of cases.
+
+        Args:
+            page_num (int): The page number to request.
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status. `CaseStatus.OPEN` or `CaseStatus.CLOSED` Defaults to None.
+            min_create_time (str or int or float or datetime, optional): Filter results by case creation time, start time.
+                 str format %Y-%m-%d %H:%M:%S. Defaults to None.
+            max_create_time (str or int or float or datetime, optional): Filter results by case creation time, end time.
+                 str format %Y-%m-%d %H:%M:%S. Defaults to None.
+            min_update_time (str or int or float or datetime, optional): Filter results by last updated time, start time.
+                 str format %Y-%m-%d %H:%M:%S. Defaults to None.
+            max_update_time (str or int or float or datetime, optional): Filter results by last updated time, end time.
+                 str format %Y-%m-%d %H:%M:%S. Defaults to None.
+            subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
+            assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
+            page_size (int, optional): Number of results to return per page. Defaults to 100.
+            sort_direction (str, optional): The direction on which to sort the response,
+                based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
+                Available options are `name`, `number`, `createdAt`, `updatedAt`, `status`, `assigneeUsername`, `subjectUsername`.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+
+        created_at = CasesClient._make_range(min_create_time, max_create_time)
+        updated_at = CasesClient._make_range(min_update_time, max_update_time)
+
+        return self._cases_service.get_page(
+            page_num,
+            name=name,
+            status=status,
+            created_at=created_at,
+            updated_at=updated_at,
+            subject=subject,
+            assignee=assignee,
+            page_size=page_size,
+            sort_direction=sort_direction,
+            sort_key=sort_key,
+            **kwargs
+        )
+
     def get_all(
         self,
         name=None,
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -77,7 +77,6 @@ class CasesClient(object):
         max_update_time=None,
         subject=None,
         assignee=None,
-        page_number=1,
         page_size=100,
         sort_direction=u"asc",
         sort_key=u"number",
@@ -98,7 +97,6 @@ class CasesClient(object):
                  str format %Y-%m-%d %H:%M:%S. Defaults to None.
             subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
             assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
-            page_number (int, optional): Page number of the results. Defaults to 1.
             page_size (int, optional): Number of results to return per page. Defaults to 100.
             sort_direction (str, optional): The direction on which to sort the response,
                 based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
@@ -120,7 +118,6 @@ class CasesClient(object):
             updated_at=updated_at,
             subject=subject,
             assignee=assignee,
-            page_number=page_number,
             page_size=page_size,
             sort_direction=sort_direction,
             sort_key=sort_key,
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -44,8 +44,8 @@ class CasesClient(object):
         assignee=None,
         page_number=1,
         page_size=100,
-        sort_direction="asc",
-        sort_key="number",
+        sort_direction=u"asc",
+        sort_key=u"number",
         **kwargs
     ):
         """Gets all cases.
@@ -54,17 +54,17 @@ class CasesClient(object):
             name (str, optional): Filter results by case name, matches partial names. Defaults to None.
             status (str, optional): Filter results by case status, `OPEN` or `CLOSED`. Defaults to None.
             created_at (str, optional): Filter results by case creation time range, format ISO interval.
-            Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
             updated_at (str, optional): Filter results by last updated time range, format ISO interval.
-            Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
             subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
             assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
             page_number (int, optional): Page number of the results. Defaults to 1.
             page_size (int, optional): Number of results to return per page. Defaults to 100.
             sort_direction (str, optional): The direction on which to sort the response,
-            based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
+                based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
             sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
-            Available options are name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername.
+                Available options are name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername.
 
         Returns:
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
@@ -105,7 +105,7 @@ class CasesClient(object):
             subject (str, optional): A subject of the case. Defaults to empty string.
             assignee (str, optional): User UID of the assignee. Defaults to empty string.
             description (str, optional): Description of the case. Defaults to empty string.
-            findings (str, optional): Observations of the case. Defaults to empty string.
+            findings (str, optional): Notes on the case. Defaults to empty string.
 
         Returns:
             :class:`py42.response.Py42Response`
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -1,0 +1,36 @@
+
+class CasesClient(object):
+
+    def __init__(self, cases_service, cases_file_event_client):
+        self._cases_service = cases_service
+        self._file_events = cases_file_event_client
+
+    @property
+    def file_events(self):
+        """A collection of method for managing cases file events.
+
+        Returns :
+            :class:`py42.services.casesfileevents.FileEventsService`
+        """
+        self.file_events = self._file_events
+
+
+    def create():
+        self._cases_service.create()
+
+
+    def get_all():
+        self._cases_service.get_all()
+
+
+    def get_case_by_name():
+        self._cases_service.get_case_by_name()
+
+
+    def export():
+        self._cases_service.export()
+
+
+    def update():
+        self._cases_service.update()
+
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -127,7 +127,7 @@ class CasesClient(object):
             **kwargs
         )
 
-    def get_case(self, case_number):
+    def get(self, case_number):
         """Retrieve case details by case number.
 
         Args:
@@ -136,7 +136,7 @@ class CasesClient(object):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        return self._cases_service.get_case(case_number)
+        return self._cases_service.get(case_number)
 
     def export_summary(self, case_number):
         """Provides case summary to download as a PDF file.
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -79,7 +79,7 @@ class CasesClient(object):
             sort_direction (str, optional): The direction on which to sort the response,
                 based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
             sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
-                Available options are name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername.
+                Available options are `name`, `number`, `createdAt`, `updatedAt`, `status`, `assigneeUsername`, `subjectUsername`.
 
         Returns:
             generator: An object that iterates over :class:`py42.response.Py42Response` objects
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -1,3 +1,11 @@
+from enum import Enum
+
+
+class CaseStatus(Enum):
+    OPEN = u"OPEN"
+    CLOSED = u"CLOSED"
+
+
 class CasesClient(object):
     """A client to expose cases API.
 
@@ -58,7 +66,8 @@ class CasesClient(object):
 
         Args:
             name (str, optional): Filter results by case name, matches partial names. Defaults to None.
-            status (str, optional): Filter results by case status, `OPEN` or `CLOSED`. Defaults to None.
+            status (enum, optional): Filter results by case status. Defaults to None.
+                e.g `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
             created_at (str, optional): Filter results by case creation time range, format ISO interval.
                 Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
             updated_at (str, optional): Filter results by last updated time range, format ISO interval.
@@ -131,7 +140,8 @@ class CasesClient(object):
             assignee (str, optional): User UID of the assignee. Defaults to empty string.
             description (str, optional): Description of the case. Defaults to empty string.
             findings (str, optional): Notes on the case. Defaults to empty string.
-            status (str, optional): Status of the case.
+            status (enum, optional): Status of the case. Defaults to None.
+                `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
         Returns:
             :class:`py42.response.Py42Response`
         """
--- a/src/py42/clients/cases.py
+++ b/src/py42/clients/cases.py
@@ -14,23 +14,91 @@ class CasesClient(object):
         """
         self.file_events = self._file_events
 
+    def create(self, name, subject, assignee, description, findings):
+        """Creates a new case.
 
-    def create():
-        self._cases_service.create()
+        Args:
+            name (str): Name of the case. 
+            subject (str): A subject of the case.
+            assignee (str): User UID of the assignee.
+            description (str): Description of the case
+            findings (str): Observations of the case.
 
+        Returns
+            :class:`py42.response.Py42Response`
+        """
 
-    def get_all():
+        self._cases_service.create(name, subject, assignee, description, findings)
+
+    def get_all(
+        self,
+        name=None,
+        status=None,
+        created_at=None,
+        updated_at=None,
+        subject=None,
+        assignee=None,
+        page_number=1,
+        page_size=100,
+        sort_direction="asc",
+        sort_key="number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status, "OPEN" or "CLOSED". Defaults to None.
+            created_at (str, optional): Filter results by case creation time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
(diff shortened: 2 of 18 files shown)


src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+            page_number (int, optional): Page number of the results. Defaults to 1.
+            page_size (int, optional): Number of results to return per page. Defaults to 500.
+            sort_direction (str, optional): The direction on which to sort the response, based on the corresponding sort key. "asc" or "desc" Defaults to "asc".
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number". Available options are
+             name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        param_dict = OrderedDict()
```
alanag13, line 70: the `get` method has an optional argument called `params` that you can pass a dict, and it will automatically handle creating the query string. It will also automatically exclude any item that has `None` assigned as its value.

CHANGELOG.md
```diff
@@ -18,6 +18,19 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
 - Python 3.9 support.
 
+- `sdk.cases` methods:
+    - `sdk.cases.create()`
+    - `sdk.cases.get_all()`
+    - `sdk.cases.export_summary()`
+    - `sdk.cases.get_case_by_case_number()`
+    - `sdk.case.update()`
+
+- `sdk.cases.file_events` methods for managing cases file events:
```
alanag13, line 28: `file events associated with a given case`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
+class CasesClient(object):
+    """A client to expose cases API.
+
+    `Rest documenation <https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases>`__
```
alanag13, line 4: misspelling of `documentation`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+    """
+
+    def __init__(self, cases_service, cases_file_event_service):
+        self._cases_service = cases_service
+        self._file_events = cases_file_event_service
+
+    @property
+    def file_events(self):
+        """A collection of method for managing cases file events.
```
alanag13, line 13: `A collection of methods for managing cases`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+            :class:`py42.services.casesfileevents.FileEventsService`
+        """
+        return self._file_events
+
+    def create(self, name, subject, assignee, description, findings):
+        """Creates a new case.
+
+        Args:
+            name (str): Name of the case.
+            subject (str): A subject of the case.
```
alanag13, line 25: the subject is the uid of a user.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+    @property
+    def file_events(self):
+        """A collection of method for managing cases file events.
+
+        Returns :
+            :class:`py42.services.casesfileevents.FileEventsService`
+        """
+        return self._file_events
+
+    def create(self, name, subject, assignee, description, findings):
```
alanag13, line 20: only `name` should be a required argument here.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number". Available options are
+             name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response` objects
+            that each contain a page of cases.
+        """
+        return self._cases_service.get_all()
+
+    def get_case_by_case_number(self, case_number):
```
alanag13, line 72: why not just `get_case`?

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def update(
+        self, case_number, name="", subject="", assignee="", description="", findings=""
+    ):
+        """Update case details for the given case number.
+
+        Args:
+            case_number (int): Case number of the case.
+            name (str, optional): Name of the case. Defaults to empty string.
+            subject (str, optional): A subject of the case. Defaults to empty string.
```
alanag13, line 178: subject is a user uid

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._connection.get(
+            "{}{}".format(self._uri_prefix.format(case_number), event_id)
+        )
+
+    def get_all_events(self, case_number):
+        """Gets all events associated to the case.
```
alanag13, line 40: associated _with_

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Args:
+            case_number (int): Case number of the case.
+            event_id (str): Event id to add to the case.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._connection.post(
+            "{}{}".format(self._uri_prefix.format(case_number), event_id)
```
unparalleled-js, line 22: Missing `u`

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Args:
+            case_number (int): Case number of the case.
+            event_id (str): Event id to fetch from the case.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._connection.get(
+            "{}{}".format(self._uri_prefix.format(case_number), event_id)
```
unparalleled-js, line 36: Missing u

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+            assignee (str, optional): User UID of the assignee. Defaults to empty string.
+            description (str, optional): Description of the case. Defaults to empty string.
+            findings (str, optional): Observations of the case. Defaults to empty string.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+
+        data = {
```
unparalleled-js, line 187: Missing `u`'s

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+            case_number (int): Case number of the case.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        uri_prefix = "{}/{}/{}".format(self._uri_prefix, case_number, "export")
+        return self._connection.get(uri_prefix)
+
+    def update(
+        self, case_number, name="", subject="", assignee="", description="", findings=""
```
unparalleled-js, line 171: Will this overwrite existing case details? If that is the case (pun intended), we should first check for existing details and use those if the user does not specify here.. We do similar things in other methods that overwrite on update.

I specifically ask because you are defaulting to `""` here. If possible, we should default to `None`. Is that possible? Or does this api actually require all values to be `""` when not specified.

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+        sort_direction="asc",
+        sort_key="number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status, "OPEN" or "CLOSED". Defaults to None.
+            created_at (str, optional): Filter results by case creation time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
```
unparalleled-js, line 116: Some of these lines are quite long

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+        page_size=items_per_page,
+        sort_direction="asc",
+        sort_key="number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status, "OPEN" or "CLOSED". Defaults to None.
```
unparalleled-js, line 115: Probably will look better in the docs if we do ```OPEN``` and ```CLOSED```

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+            created_at (str, optional): Filter results by case creation time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            updated_at (str, optional): Filter results by last updated time range, format ISO interval. Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
+            assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
+            page_number (int, optional): Page number of the results. Defaults to 1.
+            page_size (int, optional): Number of results to return per page. Defaults to 100.
+            sort_direction (str, optional): The direction on which to sort the response, based on the corresponding sort key. "asc" or "desc" Defaults to "asc".
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number". Available options are
+             name, number, createdAt, updatedAt, status, assigneeUsername, subjectUsername
```
unparalleled-js, line 64: missing period

src/py42/clients/cases.py
unparalleled-js, line 4: Wait we should not link to Core Int. Is there one in `console.us.cloud` we can use?

src/py42/clients/cases.py
```diff
@@ -0,0 +1,112 @@
...
+
+        Args:
+            case_number (int): Case number of the case.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+        return self._cases_service.export_summary(case_number)
+
+    def update(
```
unparalleled-js, line 94: Why default to ""? Are we not able to use null with this API? And will this overwrite existing when not specified?

This comment also applies here: https://github.com/code42/py42/pull/264#discussion_r551553835

src/py42/sdk/__init__.py
```diff
@@ -199,6 +202,15 @@ def auditlogs(self):
         """
         return self._clients.auditlogs
 
+    @property
+    def cases(self):
+        """A collection of methods and properties for managing cases and cases events.
```
unparalleled-js, line 207: `cases events` should be `case events` or `cases' events`

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
+from collections import OrderedDict
+
+from py42.services import BaseService
+from py42.services.util import get_all_pages
+from py42.settings import items_per_page
+
+
+class CasesService(BaseService):
+    """`Rest documentation https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases`__ ."""
```
unparalleled-js, line 9: Should prob not reference Core int

tests/clients/test_cases.py
```diff
@@ -0,0 +1,62 @@
...
+        cases_client = CasesClient(mock_cases_service, mock_cases_file_event_service)
+        cases_client.get_all()
+        mock_cases_service.get_all.call_count == 1
+
+    def test_get_case_by_case_number_calls_cases_service_with_expected_params(
+        self, mock_cases_service, mock_cases_file_event_service
+    ):
+        cases_client = CasesClient(mock_cases_service, mock_cases_file_event_service)
+        cases_client.get_case_by_case_number(123456)
+        mock_cases_service.get_case.assert_called_once_with(123456)
```
unparalleled-js, line 46: This should be a global variable,like `_TEST_CASE_NUM`

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+        page_size=items_per_page,
+        sort_direction="asc",
+        sort_key="number",
+        **kwargs
+    ):
+        """Gets an individual page of cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status, "OPEN" or "CLOSED". Defaults to None.
```
unparalleled-js, line 56: We should consider including an enum `CaseStatus`

src/py42/services/cases.py
```diff
@@ -0,0 +1,194 @@
...
+    ):
+        """Update case details for the given case number.
+
+        Args:
+            case_number (int): Case number of the case.
+            name (str, optional): Name of the case. Defaults to empty string.
+            subject (str, optional): A subject of the case. Defaults to empty string.
+            assignee (str, optional): User UID of the assignee. Defaults to empty string.
+            description (str, optional): Description of the case. Defaults to empty string.
+            findings (str, optional): Observations of the case. Defaults to empty string.
```
unparalleled-js, line 181: `Observations` made me think it was similar to `alert` observations, which are essentially file events. Instead, this seems to just be a string of notes... 

Can we not call these observations?

src/py42/services/cases.py
unparalleled-js, line 9: Also, the link is not correct.. It would belike:

```
`REST Documentation <https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases>`__
```

tests/services/test_cases.py
```diff
@@ -0,0 +1,119 @@
...
+        cases_service = CasesService(mock_connection)
+        mock_connection.get.side_effect = [
+            DUMMY_GET_ALL_TEST_RESPONSE,
+            DUMMY_GET_ALL_TEST_RESPONSE,
+        ]
+        for _ in cases_service.get_all():
+            continue
+
+        mock_connection.get.call_count == 2
+        assert (
```
unparalleled-js, line 52: Alan mentioned this, but take care to also verify that the params given is the expected dict. 
Here is an example: https://github.com/code42/py42/blob/master/tests/services/test_archive.py#L109-L115

tests/services/test_cases.py
```diff
@@ -0,0 +1,119 @@
...
+            mock_connection.get.call_args[0][0]
+            == u"/api/v1/case?pgNum=1&pgSize=500&srtDir=asc&srtKey=number"
+        )
+
+    def test_get_all_called_with_expected_url_and_params(self, mock_connection):
+        cases_service = CasesService(mock_connection)
+        mock_connection.get.side_effect = [
```
unparalleled-js, line 59: I generally prefer mock generators are actually generators because it is had bitten me in the past by them not being so.

There two data types _are_ different, even though they are both iterable.

tests/services/test_cases.py
```diff
@@ -0,0 +1,119 @@
...
+        cases_service = CasesService(mock_connection)
+        cases_service.export_summary(123456)
+        assert mock_connection.get.call_args[0][0] == u"/api/v1/case/123456/export"
+
+    def test_get_case_by_case_number_called_with_expected_url_and_params(
+        self, mock_connection
+    ):
+        cases_service = CasesService(mock_connection)
+        cases_service.get_case(123456)
+        assert mock_connection.get.call_args[0][0] == u"/api/v1/case/123456"
```
unparalleled-js, line 106: These can also use the _TEST_CASE_NUM global param

src/py42/services/cases.py
kiran-chaudhary: Changed.

CHANGELOG.md

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Changed. I earlier planned to implement the filter `get_case_by_name` but then that filter moved to `get_all` .

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/services/casesfileevents.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Waiting for access to test API behaviour.

src/py42/services/casesfileevents.py
kiran-chaudhary: added.

src/py42/services/casesfileevents.py
kiran-chaudhary: added.

src/py42/services/cases.py
kiran-chaudhary: added.

src/py42/services/cases.py
kiran-chaudhary: Waiting for access to api to confirm the behaviour.

src/py42/services/cases.py
kiran-chaudhary: changed.

src/py42/clients/cases.py
kiran-chaudhary: added.

src/py42/clients/cases.py
kiran-chaudhary: Once I get access I will verify and update the same.

src/py42/clients/cases.py
kiran-chaudhary: Yes, all fields should not be updated, waiting for access to confirm the behaviour.

src/py42/sdk/__init__.py
kiran-chaudhary: Changed.

tests/clients/test_cases.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: Updated.

tests/services/test_cases.py
kiran-chaudhary: I still see fields passed with None in tests.

tests/services/test_cases.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/sdk/__init__.py
```diff
@@ -192,13 +195,23 @@ def alerts(self):
 
     @property
     def auditlogs(self):
-        """A collections of methods for retrieving audit logs.
+        """A collection of methods for retrieving audit logs.
```
kiran-chaudhary, line 198: observed a typo, not related to this feature, but creating a new PR for this wasn't worth it.

tests/services/test_cases.py
alanag13: they will be passed with `None` but when `requests` actually makes the http request, the query string will only contain non-None items. (The part that removes `None` items is in `requests` code, not our code).

tests/services/test_cases.py
unparalleled-js: Oops, I thought this was a get-all method, disregard

unparalleled-js: `        items = [
            mock_case_response,
            mock_case_response,
        ]
        def gen():
            yield(next(items))

CHANGELOG.md
```diff
@@ -18,6 +18,19 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
 
+- `sdk.cases` methods:
+    - `sdk.cases.create()`
+    - `sdk.cases.get_all()`
+    - `sdk.cases.export_summary()`
+    - `sdk.cases.get_case()`
+    - `sdk.case.update()`
+
+- `sdk.cases.file_events` methods for managing file events associated with a given case:
+    - `sdk.cases.file_events.add_event`
```
unparalleled-js, line 29: missing `()` at the end of the file event methods

CHANGELOG.md
```diff
@@ -18,6 +18,19 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
...
+    - `sdk.cases.file_events.delete_event`
+    - `sdk.cases.file_events.get_event`
+    - `sdk.cases.file_events.get_all_events`
```
unparalleled-js, line 32: `get_all_event()` could just be `get_all()` I think

src/py42/clients/cases.py
```diff
@@ -0,0 +1,135 @@
...
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            updated_at (str, optional): Filter results by last updated time range, format ISO interval.
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
+            subject (str, optional): Filter results based on User UID of a subject of a case. Defaults to None.
+            assignee (str, optional): Filter results based on User UID of an assignee of a case. Defaults to None.
+            page_number (int, optional): Page number of the results. Defaults to 1.
+            page_size (int, optional): Number of results to return per page. Defaults to 100.
+            sort_direction (str, optional): The direction on which to sort the response,
+                based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
```
unparalleled-js, line 68: For the available options , can we put them in backticks so they stand out more on the doc site?

src/py42/services/cases.py
```diff
@@ -0,0 +1,195 @@
+from py42.services import BaseService
+from py42.services.util import get_all_pages
+from py42.settings import items_per_page
+
+
+class CasesService(BaseService):
+    """`Rest documentation <https://default-cases.core-int.cloud.code42.com/swagger-ui.html#/Cases>`__ ."""
```
unparalleled-js, line 7: Not a big deal, but I don't think this service needs class/method documentation.

unparalleled-js: Only the client and the file events service do show up in the rst file and are directly called by the user

src/py42/services/cases.py
```diff
@@ -0,0 +1,195 @@
...
+
+        data = {
+            u"assignee": assignee,
+            u"description": description,
+            u"findings": findings,
+            u"name": name,
+            u"subject": subject,
+        }
+        return self._connection.put(
+            u"{}/{}".format(self._uri_prefix, case_number), data
```
unparalleled-js, line 194: Can you specify which kwarg `data` is for `put()`,

Probably `json=data`, unless these are query params, then I think it's `params=data`

unparalleled-js: Look for other examples in py42!

tests/clients/test_cases.py
```diff
@@ -0,0 +1,64 @@
...
+        self, mock_cases_service, mock_cases_file_event_service
+    ):
+        cases_client = CasesClient(mock_cases_service, mock_cases_file_event_service)
+        assert cases_client.file_events is mock_cases_file_event_service
+
+    def test_create_calls_cases_service_with_expected_params(
+        self, mock_cases_service, mock_cases_file_event_service
+    ):
+        cases_client = CasesClient(mock_cases_service, mock_cases_file_event_service)
+        cases_client.create("name", "subject", "assignee", "description", "observation")
```
unparalleled-js, line 31: A helpful thing to do in python (that I have recently adopted) is to always use key-value arg specification (kwargs) when they are optional args. This tells the code-reader that they are optional.

Example:

```
cases_client.create("name", subject="subject", assignee="assignee", description="description", "observation")

mock_cases_service.create.assert_called_once_with(
            "name", subject="subject", assignee="assignee", description="description", observation="observation"
        )
```

src/py42/clients/cases.py
```diff
@@ -0,0 +1,135 @@
...
+            name (str): Name of the case.
+            subject (str, optional): User UID of a subject of a case.
+            assignee (str, optional): User UID of the assignee.
+            description (str, optional): Description of the case
+            findings (str, optional): Observations of the case.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._cases_service.create(
```
unparalleled-js, line 35: Will need to make the change her as well:

https://github.com/code42/py42/pull/264/files#r552785803

src/py42/services/cases.py
```diff
@@ -0,0 +1,196 @@
...
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+
+        data = {
+            u"name": name,
+            u"assignee": assignee,
+            u"description": description,
+            u"findings": findings,
+            u"subject": subject,
```
unparalleled-js, line 192: Talked offline - missing `status` param

src/py42/services/cases.py
```diff
@@ -0,0 +1,196 @@
...
+        }
+        return self._connection.put(
```
unparalleled-js, line 194: talked offline - handle when trying to update a closed case via something like 

```
      except Py42BadRequestError as err:
            if u"NO_EDITS_ONCE_CLOSED" in err.response.text:
                raise Exception("No edits once closed")
            raise
```

Except using a custom exception

src/py42/clients/cases.py
```diff
@@ -0,0 +1,135 @@
...
+            name, subject, assignee, description, findings
+        )
+
+    def get_all(
+        self,
+        name=None,
+        status=None,
+        created_at=None,
+        updated_at=None,
```
kiran-chaudhary, line 44: @unparalleled-js @alanag13  @timabrmsn 
`created_at` and `updated_at` are ranges separated by '\', wondering whether we should make it simpler for user by breaking it into 4 fields, `created_at_from` , `created_at_to`  and `updated_at_from`, `updated_at_to`.

The names are just suggestions, open to other naming options.
Also the `datetime` format expected by the API is ISO format, do we want to follow that ?

src/py42/clients/cases.py
kiran-chaudhary: Changed

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: As discussed offline, changed link to https://developer.code42.com/api/#tag/Cases

src/py42/services/cases.py
kiran-chaudhary: Enums were not available until python3.4, I have added the changes but py2.7 is complaining `enum not found`. 
One of the solution is to add dependency of `enum34` module for py2.7. Do we want to add it as dependency?

CHANGELOG.md
kiran-chaudhary: added.

src/py42/clients/cases.py
kiran-chaudhary: added.

src/py42/services/cases.py
kiran-chaudhary: Removed link from `service/cases`

src/py42/services/cases.py
kiran-chaudhary: Changed.

tests/clients/test_cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: Added.

src/py42/services/cases.py
kiran-chaudhary: Added.

CHANGELOG.md
kiran-chaudhary: changed.

tests/integration/test_cases.py
```diff
@@ -0,0 +1,40 @@
+import pytest
```
kiran-chaudhary, line 1: Need to fix this tests.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,175 @@
+from enum import Enum
```
unparalleled-js, line 1: This import statement is causing the build to fail on python 2.7 - must import full path

unparalleled-js: Oh I see now... This just is not available in python2, but I don't think it is needed anyway.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,175 @@
...
+
+from py42.util import parse_timestamp_to_milliseconds_precision
+
+
+class CaseStatus(Enum):
```
unparalleled-js, line 6: This does not need to inherit the Enum class, just look at other enums in py42 for examples.

unparalleled-js: Here is an example: https://github.com/code42/py42/blob/master/src/py42/services/detectionlists/departing_employee.py#L16

src/py42/clients/cases.py
```diff
@@ -0,0 +1,175 @@
...
+            name,
+            subject=subject,
+            assignee=assignee,
+            description=description,
+            findings=findings,
+        )
+
+    @staticmethod
+    def _make_range(begin_time, end_time):
+        if not begin_time or not end_time:
```
unparalleled-js, line 55: I am wondering if it would make sense to have `end_time` default to the current time if not provided, so that way it is easy to get all cases from a point back in time until now. Does that make sense?

src/py42/clients/cases.py
```diff
@@ -0,0 +1,175 @@
...
+        sort_direction=u"asc",
+        sort_key=u"number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (enum, optional): Filter results by case status. Defaults to None.
+                e.g `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
```
unparalleled-js, line 82: Can just list the strings here. Defaults to `OPEN`. The enums I refer to in python are just basically constants that are accessible via an easy to import class... Still strings though..

src/py42/clients/cases.py
```diff
@@ -0,0 +1,175 @@
...
+        page_size=100,
+        sort_direction=u"asc",
+        sort_key=u"number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (enum, optional): Filter results by case status. Defaults to None.
```
unparalleled-js, line 81: The type should be `str`, not `enum`

unparalleled-js: Sorry for the misunderstand, I just meant use the enums like we do in py42 in other places, not importh this python package

src/py42/exceptions.py
```diff
@@ -255,6 +255,13 @@ def __init__(self, exception, rule_id):
         super(Py42NotFoundError, self).__init__(exception, message)
 
 
+class Py42InvalidActionError(Py42BadRequestError):
```
unparalleled-js, line 258: This is custom error is no different than the bad request error and does virtually nothing... This is not what I meant in my other comment!

It should be something like this:

```
class Py42UpdateClosedCaseError(Py42BadRequestError):
"""An error raised when the the operation/action is not permitted."""

  def __init__(exception):
    msg = "Cannot update a closed case."
    super(Py42InvalidActionError, self).__init__(exception, msg)
```

The message is part of the init func.  See other errors in this file as an example.

src/py42/services/cases.py
```diff
@@ -0,0 +1,207 @@
...
+            u"subject": subject or current_case_data.get(u"subject"),
+            u"status": status or current_case_data.get("status"),
+        }
+        try:
+            return self._connection.put(
+                u"{}/{}".format(self._uri_prefix, case_number), json=data
+            )
+        except Py42BadRequestError as err:
+            if u"NO_EDITS_ONCE_CLOSED" in err.response.text:
+                raise Py42InvalidActionError(
```
unparalleled-js, line 205: See https://github.com/code42/py42/pull/264/files#r553466324

This is what should be:

```
if u"NO_EDITS_ONCE_CLOSED" in err.response.text:
  raise Py42UpdateClosedCaseError()
raise
```

src/py42/services/cases.py
```diff
@@ -0,0 +1,207 @@
...
+        sort_key=u"number",
+        **kwargs
+    ):
+        """Gets an individual page of cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status, `OPEN` or `CLOSED`. Defaults to None.
+            created_at (str, optional): Filter results by case creation time range, format ISO interval.
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
```
unparalleled-js, line 59: `Defaults to None.` should be the closing part of the parameter doc, not in the middle

unparalleled-js: This happens in other spots as well

src/py42/services/cases.py
```diff
@@ -0,0 +1,207 @@
...
+            updated_at (str, optional): Filter results by last updated time range, format ISO interval.
+                Defaults to None. e.g 2020-08-31T11:00:00Z/2020-09-01T15:30:00Z
```
unparalleled-js, line 61: Missing periods as well

src/py42/services/cases.py
unparalleled-js: Actually.... Might even be preferred to just remove all the docs from this service since this is not a public service. That way, less stuff to manage and make sure is correct

tests/services/test_cases.py
```diff
@@ -0,0 +1,215 @@
...
+            "assignee": "string",
+            "description": None,
+            "status": "OPEN",
+            "findings": u"x",
+        }
+        mock_connection.put.assert_called_once_with(
+            u"/api/v1/case/{}".format(_TEST_CASE_NUMBER), json=data
+        )
+
+    def test_update_failure_raised_appropriate_custom_exception(
```
unparalleled-js, line 204: Missing a test that makes sure `update` uses current values of the case when not given new values

src/py42/services/cases.py
alanag13: agree here, only the client part that the users actually interact with need docstrings.

src/py42/clients/cases.py
alanag13: I would agree with this approach, you could similarly use the epoch time if the user only supplied an end time and no begin time.

src/py42/clients/cases.py
alanag13: the pattern that we have in the rest of the codebase would mean that these fields should be named `min_create_time`, `max_create_time`, etc, but I agree that this is a good idea.

We should allow the string representation of whatever the API expects to be used.

src/py42/clients/cases.py
alanag13: I'm fairly certain that this APIs behavior is that it does _not_ support partial updates, that is, if you pass an empty string, the field will be updated to be an empty string. This should be documented if it is the case.

src/py42/services/cases.py
kiran-chaudhary: Removed docs from service.

tests/services/test_cases.py
kiran-chaudhary: The test above this test does cover that case.

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/exceptions.py
kiran-chaudhary: Changed. 

In this expected style of writing a new error class is created for every case. I personally feel it is too much work.

We could create more generic error name classes and respective modules can pass their appropriate messages. That is why I renamed it. `InvalidActionError` or `InvalidOperationError` can be reused in other features as well and all those module need to send is the appropriate message respective to its feature.  Its just a different style of doing the same thing.

src/py42/clients/cases.py
kiran-chaudhary: Added.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Done.

src/py42/clients/cases.py
kiran-chaudhary: Addressed.

src/py42/clients/cases.py
kiran-chaudhary: Done.

src/py42/clients/cases.py
kiran-chaudhary: Filters should have defaults to None, as default to `OPEN` will filter out the closed cases.

src/py42/clients/cases.py
kiran-chaudhary: changed.

src/py42/exceptions.py
unparalleled-js: So say I am writing a program that does stuff with cases and handles the error for when trying to update a closed case... 

Which error makes more sense to catch?  If you just catch one that says `InvalidOperationError`, than how do you know what the invalid operation was unless you also check the error message.  By having more specific errors, it makes so that the users know the condition for which the error is raised and can handle appropriately, rather than having to figure it out on their own.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,196 @@
...
+class CaseStatus(object):
+    OPEN = u"OPEN"
+    CLOSED = u"CLOSED"
+
+    @staticmethod
+    def choices():
+        return get_attribute_keys_from_class(CaseStatus)
+
+
+class Py42InvalidCaseStatus(Exception):
```
unparalleled-js, line 16: This should be moved to `errors.py`

unparalleled-js, line 16: Custom error classes should end in the word `Error`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,196 @@
...
+            sort_direction (str, optional): The direction on which to sort the response,
+                based on the corresponding sort key. `asc` or `desc`. Defaults to `asc`.
+            sort_key (str, optional): Values on which the response will be sorted. Defaults to "number".
+                Available options are `name`, `number`, `createdAt`, `updatedAt`, `status`, `assigneeUsername`, `subjectUsername`.
+
+        Returns:
+            generator: An object that iterates over :class:`py42.response.Py42Response` objects
+            that each contain a page of cases.
+        """
+        if status and status.upper() not in CaseStatus.choices():
```
unparalleled-js, line 122: There is no need to do this... This should be handled on the server.  What if they add another status?  Now we have to go update py42... Let the user pass in a `str` for the value... We provide enums as a guidance, not as a the law.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,196 @@
...
+        Args:
+            case_number (int): Case number of the case.
+            name (str, optional): Name of the case. Defaults to empty string.
+            subject (str, optional): A subject of the case. Defaults to empty string.
+            assignee (str, optional): User UID of the assignee. Defaults to empty string.
+            description (str, optional): Description of the case. Defaults to empty string.
+            findings (str, optional): Notes on the case. Defaults to empty string.
+            status (enum, optional): Status of the case. Defaults to None.
+                `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
+        Returns:
```
unparalleled-js, line 185: Missing space between `Returns:` and the args

src/py42/clients/cases.py
```diff
@@ -0,0 +1,196 @@
...
+        """Updates case details for the given case number.
+
+        Args:
+            case_number (int): Case number of the case.
+            name (str, optional): Name of the case. Defaults to empty string.
+            subject (str, optional): A subject of the case. Defaults to empty string.
+            assignee (str, optional): User UID of the assignee. Defaults to empty string.
+            description (str, optional): Description of the case. Defaults to empty string.
+            findings (str, optional): Notes on the case. Defaults to empty string.
+            status (enum, optional): Status of the case. Defaults to None.
```
unparalleled-js, line 183: The type of `status` should be `str`.

unparalleled-js, line 183: The `Defaults to None.` should be at the end of the doc str.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,196 @@
...
+                `CaseStatus.OPEN.value` or `CaseStatus.CLOSED.value`.
```
unparalleled-js, line 184: These should just be `CaseStatus.OPEN and `CaseStatus.CLOSED`.... No `.value`...

src/py42/exceptions.py
kiran-chaudhary: I understand your view. Just for the sake of argument, yes `messages` convey the needed information. From my perspective the user would read the message. imo, that is the intended purpose of the messages. Here instead of conveying the information through messages we are conveying through error class names.

tests/services/test_cases.py
unparalleled-js: Oh I see. I guess it was not obvious based on the test signature

src/py42/clients/cases.py
unparalleled-js: Or removed, see comments below.

kiran-chaudhary: aah, I always miss this. I am on it.

src/py42/clients/cases.py
kiran-chaudhary: That is the reason why I was hesitant to use it, so now if we have the enum and user passes some another string and we allow that then what is the point of using enum?

alanag13: the point is convenience -- it makes it so that the user doesn't have to memorize the list available strings for this parameter since they are available in code.

This does mean that if a new value becomes possible that a user _would_ have to just know it in order to use it, but that is better than having them try it our method erroring.

src/py42/exceptions.py
alanag13: You cannot (well okay, you can, but shouldn't) handle errors based on what the error message is. The point of an exception is not only to tell the user what went wrong, but also to give them a path to handling that specific error in their own code.

src/py42/clients/cases.py
kiran-chaudhary: I beg to differ, we have documentation to help so that they don't have to memorize. 
However its 2 against 1, so I will keep the enum and allow string.

src/py42/exceptions.py
kiran-chaudhary: I see your point and does makes sense in context when there can be multiple reasons of a failure, its better to provide different errror(classes). 
However this is very relative and differs on how we see it, in the code itself we look at the message like "EDITS_NOT_ON_CLOSED" and decide what action to take. So both ways works. I guess,  our team's preference way is  to go with new error class.

src/py42/clients/cases.py
kiran-chaudhary: Removed.

src/py42/clients/cases.py
kiran-chaudhary: added.

src/py42/clients/cases.py
kiran-chaudhary: Moved.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Removed.

src/py42/clients/cases.py
kiran-chaudhary: Changes done.

src/py42/exceptions.py
alanag13: We inspect the message because that's what we _have_ to do in order create the exception, but the aim here is to make this a better experience for users.

src/py42/services/cases.py
```diff
@@ -0,0 +1,122 @@
...
+            assignee=assignee,
+            subject=subject,
+            page_number=page_number,
+            page_size=page_size,
+            sort_direction=sort_direction,
+            sort_key=sort_key,
+            **kwargs
+        )
+
+    def get_case(self, case_number):
```
alanag13, line 88: can we make this just `get`?

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class CasesFileEventsService(BaseService):
+
+    _uri_prefix = u"/api/v1/case/{0}/fileevent/"
+
+    def __init__(self, connection):
+        super(CasesFileEventsService, self).__init__(connection)
+
+    def add_event(self, case_number, event_id):
```
alanag13, line 11: can we make this just `add`?

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+            event_id (str): Event id to add to the case.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._connection.post(
+            u"{}{}".format(self._uri_prefix.format(case_number), event_id)
+        )
+
+    def get_event(self, case_number, event_id):
```
alanag13, line 25: can we make this just `get`?

src/py42/services/casesfileevents.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Args:
+            case_number (int): Case number of the case.
+
+        Returns
+            :class:`py42.response.Py42Response`
+        """
+        return self._connection.get(self._uri_prefix.format(case_number))
+
+    def delete_event(self, case_number, event_id):
```
alanag13, line 50: can we make this just `delete`?

src/py42/services/casesfileevents.py
kiran-chaudhary: Changed.

src/py42/services/casesfileevents.py
kiran-chaudhary: Changed.

src/py42/services/casesfileevents.py
kiran-chaudhary: Changed.

src/py42/services/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
```diff
@@ -0,0 +1,185 @@
...
+        """Creates a new case.
+
+        Args:
+            name (str): Name of the case.
+            subject (str, optional): User UID of a subject of a case.
+            assignee (str, optional): User UID of the assignee.
+            description (str, optional): Description of the case
+            findings (str, optional): Observations of the case.
+
+        Returns
```
unparalleled-js, line 47: Missing `:` after `Returns`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,185 @@
...
+
+    def __init__(self, cases_service, cases_file_event_service):
+        self._cases_service = cases_service
+        self._file_events = cases_file_event_service
+
+    @property
+    def file_events(self):
+        """A collection of methods for managing file events associated with a given case.
+
+        Returns :
```
unparalleled-js, line 30: Extra space between `Returns` and `:`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,185 @@
...
+            :class:`py42.services.casesfileevents.FileEventsService`
```
unparalleled-js, line 31: This link is not working. I believe it should be `py42.services.casesfileevents.CasesFileEventsService`

src/py42/clients/cases.py
```diff
@@ -0,0 +1,185 @@
...
+        sort_direction=u"asc",
+        sort_key=u"number",
+        **kwargs
+    ):
+        """Gets all cases.
+
+        Args:
+            name (str, optional): Filter results by case name, matches partial names. Defaults to None.
+            status (str, optional): Filter results by case status. `CaseStatus.OPEN` or `CaseStatus.CLOSED` Defaults to None.
+            min_create_time (str or int or float or datetime, optional): Filter results by case creation time, start time.
```
unparalleled-js, line 91: `begin_create_time` and `end_create_time` would follow a more familiar pattern

src/py42/clients/cases.py
```diff
@@ -0,0 +1,185 @@
...
+            status=status,
+            created_at=created_at,
+            updated_at=updated_at,
+            subject=subject,
+            assignee=assignee,
+            page_number=page_number,
+            page_size=page_size,
+            sort_direction=sort_direction,
+            sort_key=sort_key,
+            **kwargs
```
unparalleled-js, line 127: I don't think any of the values from `kwargs` make it to the request, from what I can.. Is this needed?

kiran-chaudhary: thanks, missed adding it in `.services.get_page`. 
Added `kwargs` because API supports filter by other fields which we have not supported, like 'isAssigned', `lastModifiedBy`, etc.

Also do we need a `get_page` method for cases?

src/py42/clients/cases.py
kiran-chaudhary: Changed.

src/py42/clients/cases.py
kiran-chaudhary: Removed.

src/py42/clients/cases.py
kiran-chaudhary: added.

src/py42/clients/cases.py
unparalleled-js: Ah just noticing that we don't have one. Yeah, it'd be nice to have, as the generators don't work too well in certain applications. Do you want to add it now or just create a story for it and add it later?

unparalleled-js (approved):
Expose `get_page` now or later.

Yay, we have Cases now!

src/py42/clients/cases.py
kiran-chaudhary: Added.

alanag13 (approved):

PR #268, opened 2021-01-19
Add custom exceptions in cases

kiran-chaudhary:
### Description of Change ###

Added custom exceptions for cases:

- Py42CaseNameExistsError: In create case when case name already exists in the system.
- Py42DescriptionLimitExceededError: In create case when description is too long.
- Py42DescriptionLimitExceededError: In update case when description is too long.
- Py42CaseAlreadyHasEventError: When file event is associated multiple times to a case.
- Py42UpdateClosedCaseError: While adding file event to a closed case.
- Py42UpdateClosedCaseError: While deleting file event from a closed case.

### Issues Resolved ###

- closes # INTEG-1456

### Testing Procedure ###

```
try:
    sdk.cases.create("Duplicate name")
except Py42CaseNameExistsError as err:
```### PR Checklist ###

- [ ] Add unit tests to verify this change 
- [ ] Add an entry to CHANGELOG.md describing this change 
- [ ] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -263,6 +263,22 @@ class Py42UpdateClosedCaseError(Py42BadRequestError):
         super(Py42UpdateClosedCaseError, self).__init__(exception, msg)
 
 
+class Py42CaseNameExistsError(Py42BadRequestError):
+    """An error raised when trying to create a case with a name that already exists."""
+
+    def __init__(self, exception, case_name):
+        msg = "Case name '{}' already exists, please set another name".format(case_name)
+        super(Py42CaseNameExistsError, self).__init__(exception, msg)
+
+
+class Py42DescriptionLimitExceededError(Py42BadRequestError):
+    """An error raised when description of a case exceeds allowed char length limit."""
+
+    def __init__(self, exception):
+        msg = "Description limit exceeded, max 254 characters allowed."
+        super(Py42DescriptionLimitExceededError, self).__init__(exception, msg)
+
+
 def raise_py42_error(raised_error):
     """Raises the appropriate :class:`py42.exceptions.Py42HttpError` based on the given
     HTTPError's response status code.
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -279,6 +279,14 @@ class Py42DescriptionLimitExceededError(Py42BadRequestError):
         super(Py42DescriptionLimitExceededError, self).__init__(exception, msg)
 
 
+class Py42CaseAlreadyHasEventError(Py42BadRequestError):
+    """An error raised when event is already associated to the case."""
+
+    def __init__(self, exception):
+        msg = "Event is already associated to the case."
+        super(Py42CaseAlreadyHasEventError, self).__init__(exception, msg)
+
+
 def raise_py42_error(raised_error):
     """Raises the appropriate :class:`py42.exceptions.Py42HttpError` based on the given
     HTTPError's response status code.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by kiran-chaudhary** [comment]
Did not add entry into change-log considering this feature is not released yet! open to adding it if needed.

File: src/py42/exceptions.py [hunk 1]
```diff
@@ -263,6 +263,22 @@ def __init__(self, exception):
...
+class Py42CaseNameExistsError(Py42BadRequestError):
+    """An error raised when trying to create a case with a name that already exists."""
+
+    def __init__(self, exception, case_name):
+        msg = "Case name '{}' already exists, please set another name".format(case_name)
+        super(Py42CaseNameExistsError, self).__init__(exception, msg)
+
+
+class Py42DescriptionLimitExceededError(Py42BadRequestError):
+    """An error raised when description of a case exceeds allowed char length limit."""
```
unparalleled-js, line 275: when _the_ description of a case... (missing the)

src/py42/services/cases.py
```diff
@@ -22,7 +24,14 @@ def create(
             u"name": name,
             u"subject": subject,
         }
-        return self._connection.post(self._uri_prefix, json=data)
+        try:
+            return self._connection.post(self._uri_prefix, json=data)
+        except Py42BadRequestError as err:
+            if u"NAME_EXISTS" in err.response.text:
+                raise Py42CaseNameExistsError(err, name)
+            if u"DESCRIPTION_TOO_LONG" in err.response.text:
```
unparalleled-js, line 32: even though programmatically has no effect, it is still better to use `elif`

Here is why:  Say this code changes and no longer raised the error now, we may forget to change it to `elif` then.  Having it now is more readable as well.

src/py42/services/cases.py
```diff
@@ -118,4 +127,6 @@ def update(
         except Py42BadRequestError as err:
             if u"NO_EDITS_ONCE_CLOSED" in err.response.text:
                 raise Py42UpdateClosedCaseError(err)
+            if u"DESCRIPTION_TOO_LONG" in err.response.text:
```
unparalleled-js, line 130: Same goes here - `elif` instead of `if`.  

Here is why: Say this code changes and no longer raised the error now, we may forget to change it to elif then. Having it now is more readable as well.

tests/services/test_cases.py
```diff
@@ -213,3 +236,49 @@ def test_update_failure_raised_appropriate_custom_exception(
...
+        mock_connection.put.side_effect = Py42BadRequestError(
+            mock_description_too_long_response
+        )
+        with pytest.raises(Py42DescriptionLimitExceededError) as e:
+            cases_service.update(_TEST_CASE_NUMBER, description=u"supposedly too long")
+
+        assert (
+            e.value.args[0]
+            == u"Description limit exceeded, max 254 characters allowed."
+        )
```
unparalleled-js, line 284: Could add a test that makes sure that the 400 still gets raised when none of the scriptions match any expected custom-exception triggers.

Not super important, but code coverage wise, it is a missing covered section of the execution tree.

unparalleled-js:
Some small suggestions but overall looks great!

tests/services/test_cases.py
kiran-chaudhary: Added.

src/py42/services/cases.py
kiran-chaudhary: Added.

src/py42/services/cases.py
kiran-chaudhary: Added.

kiran-chaudhary:
Added remaining custom errors as discussed during grooming.

src/py42/exceptions.py
kiran-chaudhary: Done.

unparalleled-js (approved):
Some missing `the`'s in the Changelog but looks good.

e.g. when _the_ event is added ...
when _the_ description ...

PR #275, opened 2021-01-26
Integ test adjustments

unparalleled-js:
### Description of Change ###

Fixes small things and improves the flow of using the mock server while using running py42 integration tests.

* Use default arguments that are the expected host/port of the mockserver when environment variables are not set.
* Corrects the name of said environment variable in the CONTRIBUTING guide as well as update the CONTRIBUTING guide 
to include part about default variables.
* Checks for successful status code range instead of specific status code in integration tests
* Fails integ tests early when SDK doesn't get created (thanks @timabrmsn )
*** Removes `Content-Type application/json` when no request data.**
* Make URLs easier to parse; remove unneeded class variables
* Some formatting and spelling fixes
* Makes test py2 compat

### Issues Resolved ###
n/a

### Testing Procedure ###
Now, when testing the mock server PRs, you don't have to worry about setting the environment variables.

### PR Checklist ###

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,7 +156,12 @@ def test_add_one_and_one_equals_two():
 
 #### Integration tests
 
-Set environment variables C42_USERNAME and C42_PW with CCA credentials. To execute integration tests:
+If not using the mock server, set the environment variables `C42_USER` and `C42_PW`
+with CCA credentials. Otherwise, the integration tests default to using
+`127.0.0.1:4200`, which is the same address that the mock server is set to run on.
+
+To execute integration tests:
+
 ```bash
 $ pytest -m integration
 ```
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -158,7 +158,7 @@ def test_add_one_and_one_equals_two():
 
 If not using the mock server, set the environment variables `C42_HOST`, `C42_USER`,
 and `C42_PW` with CCA credentials. Otherwise, the integration tests default to using
-`127.0.0.1:4200`, which is the same address that the mock server is set to run on.
+`http://127.0.0.1:4200`, which is the same address that the mock server is set to run on.
 
 To execute integration tests:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,8 +156,8 @@ def test_add_one_and_one_equals_two():
 
 #### Integration tests
 
-If not using the mock server, set the environment variables `C42_USER` and `C42_PW`
-with CCA credentials. Otherwise, the integration tests default to using
+If not using the mock server, set the environment variables `C42_HOST`, `C42_USER`,
+and `C42_PW` with CCA credentials. Otherwise, the integration tests default to using
 `127.0.0.1:4200`, which is the same address that the mock server is set to run on.
 
 To execute integration tests:
--- a/src/py42/sdk/queries/alerts/alert_query.py
+++ b/src/py42/sdk/queries/alerts/alert_query.py
@@ -31,13 +31,16 @@ class AlertQuery(BaseQuery):
         groups_string = u",".join(
             str(group_item) for group_item in self._filter_group_list
         )
-        json = u'{{"tenantId": null, "groupClause":"{0}", "groups":[{1}], "pgNum":{2}, "pgSize":{3}, "srtDirection":"{4}", "srtKey":"{5}"}}'.format(
-            self._group_clause,
-            groups_string,
-            self.page_number,
-            self.page_size,
-            self.sort_direction,
-            self.sort_key,
+        json = (
+            u'{{"tenantId": null, "groupClause":"{0}", "groups":[{1}], "pgNum":{2}, '
+            u'"pgSize":{3}, "srtDirection":"{4}", "srtKey":"{5}"}}'.format(
+                self._group_clause,
+                groups_string,
+                self.page_number,
+                self.page_size,
+                self.sort_direction,
+                self.sort_key,
+            )
         )
         return json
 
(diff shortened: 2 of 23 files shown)


src/py42/sdk/queries/alerts/alert_query.py
```diff
@@ -31,13 +31,16 @@ def __str__(self):
...
             str(group_item) for group_item in self._filter_group_list
         )
-        json = u'{{"tenantId": null, "groupClause":"{0}", "groups":[{1}], "pgNum":{2}, "pgSize":{3}, "srtDirection":"{4}", "srtKey":"{5}"}}'.format(
-            self._group_clause,
-            groups_string,
-            self.page_number,
-            self.page_size,
-            self.sort_direction,
-            self.sort_key,
+        json = (
```
unparalleled-js, line 34: just makes a long line not as long

src/py42/services/alertrules.py
```diff
@@ -7,10 +7,7 @@
 
 class AlertRulesService(BaseService):
     """A service to manage Alert Rules."""
-
-    _version = u"v1"
```
unparalleled-js, line 11: these class vars did not seem needed and makes it card to copy the url

src/py42/services/alertrules.py
```diff
@@ -7,10 +7,7 @@
...
-    _resource = u"Rules/"
-    _api_prefix = u"/svc/api/{}/{}".format(_version, _resource)
+    _api_prefix = u"/svc/api/v1/Rules/"
```
unparalleled-js, line 10: Renamed to `_resource`, as it is not prefix in this case

tests/integration/conftest.py
```diff
@@ -41,12 +42,14 @@ def observer_id(request):
...
-    return _sdk.from_local_account(host, os.environ["C42_USER"], os.environ["C42_PW"])
+    user = os.environ.get("C42_USER") or "test.user@example.com"
+    pw = os.environ.get("C42_PW") or "password"
+    return _sdk.from_local_account(host, user, pw)
 
 
 @pytest.fixture(scope="session")
 def timestamp():
-    return str(int(datetime.utcnow().timestamp()))
+    return convert_datetime_to_epoch(datetime.utcnow())
```
unparalleled-js, line 52: Makes py2 compat

tests/integration/conftest.py
```diff
@@ -41,12 +42,14 @@ def observer_id(request):
 
 @pytest.fixture(scope="session")
 def connection(host):
-    return _sdk.from_local_account(host, os.environ["C42_USER"], os.environ["C42_PW"])
+    user = os.environ.get("C42_USER") or "test.user@example.com"
```
unparalleled-js, line 45: user/pwd does not really matter for the mock server but they need to be set

src/py42/services/alertrules.py
```diff
@@ -98,14 +93,11 @@ def get(self, rule_id):
             :class:`py42.response.Py42Response`
         """
         data = {u"tenantId": self._tenant_id, u"ruleIds": [rule_id]}
-        return self._connection.post(self._api_prefix, json=data)
+        return self._connection.post(self._endpoint, json=data)
 
 
 class ExfiltrationService(BaseService):
-
```
unparalleled-js, line 105: IGNORE THIS COMMENT

@annie-payseur TEST, this is what it will look like

tests/integration/conftest.py
```diff
@@ -41,12 +42,14 @@ def observer_id(request):
...
+    pw = os.environ.get("C42_PW") or "password"
+    return _sdk.from_local_account(host, user, pw)
```
timabrmsn, line 47: Can we fail fast here so we don't have to wait out all the failures if the mockserver isn't up or we provide bad creds? Something like:

```
    try:
        return _sdk.from_local_account(host, user, pw)
    except Exception as e:
        pytest.exit("Failed to init sdk for integration tests: {e}".format(e))
```

CONTRIBUTING.md
```diff
@@ -156,7 +156,12 @@ def test_add_one_and_one_equals_two():
 
 #### Integration tests
 
-Set environment variables C42_USERNAME and C42_PW with CCA credentials. To execute integration tests:
+If not using the mock server, set the environment variables `C42_USER` and `C42_PW`
+with CCA credentials. Otherwise, the integration tests default to using
+`127.0.0.1:4200`, which is the same address that the mock server is set to run on.
```
timabrmsn, line 161: Should we make host be an env variable too? Or at least call out where you'd change the host from the default here.

unparalleled-js: Done

tests/integration/conftest.py
unparalleled-js: Done

tests/integration/conftest.py
```diff
@@ -40,13 +35,26 @@ def observer_id(request):
...
+    user = os.environ.get("C42_USER") or "test.user@example.com"
+    pw = os.environ.get("C42_PW") or "password"
+    return _get_sdk(host, user, pw)
+
+
+def _get_sdk(host, user, pw):
+    try:
+        return _sdk.from_local_account(host, user, pw)
+    except Exception as err:
+        pytest.exit(
```
unparalleled-js, line 49: `echo $?` seems to prove this works, but is kind of odd since the tests show as green

unparalleled-js:
I am going to work out something regarding empty JSON in requests

tests/integration/conftest.py
timabrmsn: Are you running all tests or just integration when it shows green? For me running `pytest -m integration` highlights yellow since it failed before anything actually ran.

unparalleled-js: I was running `pytest tests/test_users/py`.
I see the yellow when running all the tests, but I think that is because of other warnings

unparalleled-js: Or perhaps some race condition. It seems to work though! In both cases

timabrmsn (approved):

PR #276, opened 2021-01-26
make header not persist

unparalleled-js:
### Description of Change ###

Fixed bug from last PR where the header would persist if getting added once.
Now, the header will get removed, as you can see with the unit test.
Also, finally my Prism tests are working, so that is a really good sign.

### Testing Procedure ###
Unit test
make sure py42 still works (it does but I will continue to verify)

### PR Checklist ###

### DIFF
--- a/src/py42/services/_connection.py
+++ b/src/py42/services/_connection.py
@@ -226,6 +226,7 @@ class Connection(object):
         if json is not None:
             data = json_lib.dumps(json)
 
+        headers = headers or {}
         headers.update(self._headers)
         if data:
             headers.update({u"Content-Type": u"application/json"})
--- a/src/py42/services/_connection.py
+++ b/src/py42/services/_connection.py
@@ -183,7 +183,6 @@ class Connection(object):
                 auth=auth,
                 hooks=hooks,
             )
-
             response = self._session.send(
                 request,
                 stream=stream,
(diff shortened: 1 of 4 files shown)


src/py42/services/_connection.py
```diff
@@ -244,6 +245,14 @@ def _prepare_request(
         _print_request(method, url, params=params, data=data)
         return self._session.prepare_request(request)
 
+    def _set_content_header(self, data):
```
alanag13, line 248: This isn't thread-safe -- if we were hitting this code in one of our batch cli processes for example, we could be removing/adding  a header right before a request that was supposed to have the opposite happen was sent.

alanag13:
@unparalleled-js I think this would be better implemented by modifying https://github.com/code42/py42/blob/b75e72073f60b0144fd260765ef3e990e51d137a/src/py42/services/_connection.py#L187.

I believe the `send` function will add any `headers` passed to it to the headers that are in the session, so something like:
```python
headers = {"Content-Type": "application/json"}  if data else None
self._session.send(..., headers=headers)
```

this way the functionality isn't dependent on any held state.

unparalleled-js:
Would there be a changelog entry for this?  I am not sure any bug would have been present to the user

src/py42/services/_connection.py
```diff
@@ -227,12 +226,14 @@ def _prepare_request(
         if json is not None:
             data = json_lib.dumps(json)
 
-        user_headers = _create_user_headers(headers, data)
-        self._headers.update(user_headers)
+        headers = _create_user_headers(headers, data)
```
alanag13, line 229: The user headers should be added last so that a method can manually override `Content-Type` (or any of the items in `self._headers`).

alanag13 (approved):

PR #277, opened 2021-01-27
Fix content type param header issue

unparalleled-js:
### Description of Change ###

Fixes issue that removed Content-Type header when it was given as a parameter

### Issues Resolved ###
N/a

### Testing Procedure ###
Before, the key value store (which adds */* content type header) was not succeeding in adding it.
Verify that it now adds it (using -d in CLI?)

### PR Checklist ###


alanag13 (approved):

alanag13:
nice catch.

PR #278, opened 2021-01-28
conditionally set Accept header when not in headers arg

unparalleled-js:
### Description of Change ###
Conditionally sets Accept header for when user client passes in, we don't overwrite

### Issues Resolved ###
Helps Prism work in the Mock Server

### Testing Procedure ###
Run integration tests for alerts using mock server

### PR Checklist ###

### DIFF
--- a/src/py42/services/_connection.py
+++ b/src/py42/services/_connection.py
@@ -26,7 +26,6 @@ ROOT_SESSION = Session()
 ROOT_SESSION.mount(u"https://", SESSION_ADAPTER)
 ROOT_SESSION.mount(u"http://", SESSION_ADAPTER)
 ROOT_SESSION.headers = {
-    u"Accept": u"application/json",
     u"Accept-Encoding": u"gzip, deflate",
     u"Connection": u"keep-alive",
 }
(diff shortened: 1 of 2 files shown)


src/py42/services/_connection.py
```diff
@@ -230,6 +229,8 @@ def _prepare_request(
         headers.update(self._headers)
         if data and u"Content-Type" not in headers:
             headers.update({u"Content-Type": u"application/json"})
+        if u"Accept" not in headers:
```
alanag13, line 232: it's starting to look like these updates are being applied in the wrong order -- we shouldn't have to make an exception here for every header that a user passes in. I probably gave some incorrect guidance on this whole thing earlier, sorry about that.

I'm thinking what we want is something that ensure that `headers` gets applied last so that no matter what, whatever the function submitted would take precedence, like:

```python
final_headers = {}
final_headers.update(self._headers)
headers = _create_user_headers(headers)
final_headers.update(headers)
request = Request(... headers=final_headers)
```

alanag13 (approved):

PR #282, opened 2021-02-01
Test/integ stream to dest

unparalleled-js:
### Description of Change ###

* Adds missing integration test for push restore
* Adds missing unit test for `get_web_restore_info()`
* Sets `octet-stream` `Accept` header for `stream_from_backup()`.
* Removed expected test failure (fails in our Partner org but we don't expect a failure in the mock server)
* Addresses python2 warnings
* Prettifies a test-JSON blob
* Simplifies the path parameter used for web/push restore in `tox.ini`

Some of these changes are needed for the mocker server `test_archive` integration test

### Issues Resolved ###
n/a

### Testing Procedure ###
* Web restore regression test
* Run integration test with mock server

### PR Checklist ###


alanag13:
I still need to regression test web restore, but visually, this looks good.

kiran-chaudhary (approved):
The major change in `src` is passing the header, all other changes are in tests, so didn't test it out , LGTM.

unparalleled-js:
@alanag13 For what it's worth, it seems to still work

unparalleled-js:
@alanag13 

Do you plan to test that web restore still works from adding that header?  We could alternatively make a small ticket and let QA do a regression test.

(I have also made sure it still works myself and feel confident).

alanag13 (approved):
QA testing is fine, I'm pretty confident in this too, sorry for the delay.

PR #286, opened 2021-02-03
Fix integration tests in security data

peterbriggs42:
### Description of Change ###

Fix two integration tests that were not properly consuming the output of `get_all_plan_security_events`. The output of that function is now a tuple with the first element being the Py42Response object and the second term being a cursor.

### Issues Resolved ###
N/A

### Testing Procedure ###
Run the tests in `./tests/integration/test_securitydata.py` against Partner.

### PR Checklist ###

- [N/A ] Add unit tests to verify this change
- [N/A ] Add an entry to CHANGELOG.md describing this change
- [N/A ] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -108,8 +108,8 @@ class SecurityDataClient(object):
                 str format "yyyy-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
 
         Returns:
-            generator: An object that iterates over :class:`py42.response.Py42Response` objects
-            that each contain a page of events.
+            generator: An object that iterates over tuples whose first element is a :class:`py42.response.Py42Response` object
+            containing a page of events, and whose second element is a cursor.
         """
         return self._get_security_detection_events(
             plan_storage_info,
@@ -158,8 +158,8 @@ class SecurityDataClient(object):
                 str format "yyyy-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
 
         Returns:
-            generator: An object that iterates over :class:`py42.response.Py42Response` objects
-            that each contain a page of events.
+            generator: An object that iterates over tuples whose first element is a :class:`py42.response.Py42Response` object
+            containing a page of events, and whose second element is a cursor.
         """
         security_plan_storage_infos = self.get_security_plan_storage_info_list(user_uid)
         return self._get_security_detection_events(
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -109,7 +109,7 @@ class SecurityDataClient(object):
 
         Returns:
             generator: An object that iterates over tuples whose first element is a :class:`py42.response.Py42Response` object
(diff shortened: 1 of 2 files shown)


kiran-chaudhary (approved):
LGTM

tests/integration/test_securitydata.py
```diff
@@ -39,13 +39,13 @@ class TestSecurityData:
     def test_get_all_plan_security_events(self, connection, plan_info):
         response_gen = connection.securitydata.get_all_plan_security_events(plan_info)
         for response in response_gen:
```
unparalleled-js, line 41: Is `response_gen` not actually a generator?

unparalleled-js: If this works like the other generators, I would expect `response` to be of type `Py42Response`

peterbriggs42: At the end of `clients/securitydata.py::_get_security_detection_events()` there's this block:
```                if response.text:
                    cursor = response.data.get(u"cursor")
                    # if there are no results, we don't get a cursor and have reached the end
                    if cursor:
                        yield response, cursor
```
So the Py42Response is the first element in that tuple and the cursor is the second. My guess is that we updated the function without modifying the test but I haven't delved into the history to confirm that.

unparalleled-js (approved):

tests/integration/test_securitydata.py
unparalleled-js: Good catch!  we should also update the documentation of that method

src/py42/clients/securitydata.py
```diff
@@ -108,8 +108,8 @@ def get_all_plan_security_events(
                 str format "yyyy-MM-DD HH:MM:SS" or a datetime instance. Defaults to None.
 
         Returns:
-            generator: An object that iterates over :class:`py42.response.Py42Response` objects
-            that each contain a page of events.
+            generator: An object that iterates over tuples whose first element is a :class:`py42.response.Py42Response` object
+            containing a page of events, and whose second element is a cursor.
```
unparalleled-js, line 112: do you happen to know the type of this cursor?

src/py42/clients/securitydata.py
```diff
@@ -109,7 +109,7 @@ def get_all_plan_security_events(
 
         Returns:
             generator: An object that iterates over tuples whose first element is a :class:`py42.response.Py42Response` object
-            containing a page of events, and whose second element is a cursor.
+            containing a page of events, and whose second element is a string cursor.
```
unparalleled-js, line 112: `a string cursor representing the timestamp of the event` or `... with the format of a timestamp` might be good to know, but not a big deal, they will probably figure that out.

unparalleled-js: I think it's a timestamp anyway

peterbriggs42: There's a description of the cursor in the parameter docstring for this function so I think it's probably fine as is.

PR #291, opened 2021-02-16
Added or updated rest documentation link to api gateway url

kiran-chaudhary:
### Description of Change ###

Change link in documentation to point to API Gateway instead of actual micorservice.

### Issues Resolved ###

- closes #INTEG-1424

### Testing Procedure ###

Execute `tox -e docs`.
Verify changes in `alerts`, `alertrules`, `auditlogs`, `detectionlists`, `users` link at `class` level changed/added to link to API Gateway url.
Verify change in `securitydata` at `method` level. 

### PR Checklist ###

### DIFF
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -2,6 +2,11 @@ from py42.sdk.queries.alerts.filters import AlertState
 
 
 class AlertsClient(object):
+    """A client to expose alert API.
+
+    `Rest Documentation <https://developer.code42.com/api/#tag/Alerts>`__
+    """
+
     def __init__(self, alert_service, alert_rules_client):
         self._alert_service = alert_service
         self._alert_rules_client = alert_rules_client
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -23,6 +23,8 @@ class AlertsClient(object):
     def search(self, query):
         """Searches alerts using the given :class:`py42.sdk.queries.alerts.alert_query.AlertQuery`.
 
+        `Rest Documentation <https://developer.code42.com/api/#operation/Alerts_QueryAlert>`__
+
         Args:
             query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
                 See the :ref:`Executing Searches User Guide <anchor_search_alerts>` to learn more
@@ -37,6 +39,8 @@ class AlertsClient(object):
     def get_details(self, alert_ids):
         """Gets the details for the alerts with the given IDs, including the file event query that,
         when passed into a search, would result in events that could have triggered the alerts.
+
+        `Rest Documentation <https://developer.code42.com/api/#operation/Alerts_QueryAlertDetails>`__
 
         Args:
             alert_ids (iter[str]): The identification numbers of the alerts for which you want to
--- a/src/py42/services/detectionlists/high_risk_employee.py
+++ b/src/py42/services/detectionlists/high_risk_employee.py
@@ -62,6 +62,7 @@ class HighRiskEmployeeService(BaseService):
 
     def set_alerts_enabled(self, enabled=True):
         """Enables alerts.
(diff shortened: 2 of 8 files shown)


src/py42/services/users.py
```diff
@@ -11,6 +11,8 @@
 class UserService(BaseService):
     """A service for interacting with Code42 user APIs. Use the UserService to create and retrieve
     users. You can also use it to block and deactivate users.
+
+    `REST Documentation <https://developer.code42.com/api/#tag/User>`__
     """
 
     def create_user(
```
kiran-chaudhary, line 18: Did not change the link in `create_user` doc as that is not exposed on the gateway.

src/py42/clients/alerts.py
```diff
@@ -2,6 +2,11 @@
 
 
 class AlertsClient(object):
+    """A client to expose alert API.
```
unparalleled-js, line 5: Any reason why we do method level docs at some points and class level at other places?

Like couldn't `search()` have one of its own?

kiran-chaudhary: There are multiple reasons:

- Some clients do not have method's equivalent definition in API Gateway. e.g. `create_user` in `UserClient`

- Some clients consume multiple micro services. e.g. `SecurityData`

- Some classes are still not added in API Gateway. e.g. `Devices`, `ArchiveClient`

So at this point I left some of the existing swagger docs link as is, based on my understanding.

unparalleled-js: If you to py42docs.code42.com, start searching through the method documentation, find the method you want, you won't see the REST documentation link without scrolling around to the top of the client.

kiran-chaudhary: Need to understand whether we want to remove all the swagger docs in this PR or change as and when we add other docs to API Gateway.

unparalleled-js: Probably add other docs as they come in to the gateway

src/py42/services/users.py
```diff
@@ -11,6 +11,8 @@
 class UserService(BaseService):
     """A service for interacting with Code42 user APIs. Use the UserService to create and retrieve
     users. You can also use it to block and deactivate users.
+
+    `REST Documentation <https://developer.code42.com/api/#tag/User>`__
```
timabrmsn, line 15: I wonder if we should wait to change this one since the gateway has a different meaning for `userId` than what py42 is currently using?

src/py42/clients/alerts.py
kiran-chaudhary: Added rest documentation link to header considering all the links refer to the same page. As requested, added links to few (not all) method documentation to make sure user can find the link in the method if they miss out on the top of the page.

src/py42/services/users.py
kiran-chaudhary: Removed. Thanks.

src/py42/services/detectionlists/high_risk_employee.py
```diff
@@ -41,6 +41,7 @@ def add(self, user_id):
 
         Raises a :class:`Py42BadRequestError` when a user already exists in the High Risk Employee
         detection list.
+        `REST Documentation <https://developer.code42.com/api/#operation/HighRiskEmployeeControllerV2_AddEmployee>`__
 
         Args:
             user_id (str or int): The Code42 userUid of the user you want to add to the High Risk
```
unparalleled-js, line 47: Could we add the method links for High Risk Employee as well?

src/py42/services/users.py
```diff
@@ -60,7 +61,6 @@ def create_user(
 
     def get_by_id(self, user_id, **kwargs):
```
unparalleled-js, line 62: Could we add the users ones?

kiran-chaudhary: Added.

src/py42/services/detectionlists/high_risk_employee.py
kiran-chaudhary: added to almost all methods.

unparalleled-js (approved):

PR #295, opened 2021-02-22
Fix doc links

kiran-chaudhary:
### Description of Change ###
Fixes doc links. Refers to [PR](https://github.com/code42/py42/pull/291)

### PR Checklist ###

- [ n/a] Add unit tests to verify this change
- [ n/a] Add an entry to CHANGELOG.md describing this change
- [ n/a] Add docstrings for any new public parameters / methods / classes

### DIFF
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -173,7 +173,7 @@ class SecurityDataClient(object):
 
     def search_file_events(self, query):
         """Searches for file events, returns up to the first 10,000 events.
-        `REST Documentation <https://developer.code42.com/api/#tag/File-Events>`__
+        `REST Documentation <https://developer.code42.com/api/#operation/searchEventsUsingPOST>`__
 
         Args:
             query (str or :class:`py42.sdk.queries.fileevents.file_event_query.FileEventQuery`):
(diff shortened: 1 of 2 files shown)


unparalleled-js:
We are still missing some.

`add_user_risk_tags` is missing one https://developer.code42.com/api/#operation/UserControllerV2_AddRiskFactors

All of the `cases` APIs are missing.

`audit-log` is missing: `https://developer.code42.com/api/#tag/Audit-Log`.

timabrmsn (approved):

kiran-chaudhary:
> https://developer.code42.com/api/#operation/UserControllerV2_AddRiskFactors

Do we want to have a one-one linking for each method? 

My approach in adding these links has been as:

- Replace all the existing links with their respective "developer portal" link.

- Add new links wherever missing ( Here I preferred to link under the class doc against main heading ) . I am aware in earlier PR you requested to add to methods as well, I added few more then but I feel we don't need to document all methods unless the methods are out of order in different classes, etc.... `Cases` methods can be easily mapped and `audit-logs` has just two methods `get_page` and `get_all` so added the link in the main class.

unparalleled-js:
> > https://developer.code42.com/api/#operation/UserControllerV2_AddRiskFactors
> 
> Do we want to have a one-one linking for each method?
> 
> My approach in adding these links has been as:
> 
> * Replace all the existing links with their respective "developer portal" link.
> * Add new links wherever missing ( Here I preferred to link under the class doc against main heading ) . I am aware in earlier PR you requested to add to methods as well, I added few more then but I feel we don't need to document all methods unless the methods are out of order in different classes, etc.... `Cases` methods can be easily mapped and `audit-logs` has just two methods `get_page` and `get_all` so added the link in the main class doc.

When we link to specific methods, it makes easier to find the link from using the search feature on the docs site

PR #300, opened 2021-03-08
Chore/remove detectionlist create_if_not_exists

timabrmsn:
### Description of Change ###

Removes the logic for creating detection list profiles if they don't exist, since they now automatically get created on core user creation in the backend services.

### Issues Resolved ###
None

### Testing Procedure ###
Regression test methods for adding to Departing Employee, High Risk Employee, and Alert Rules lists. 

### PR Checklist ###

- [x] Add unit tests to verify this change (removed related unit tests)
- [ ] Add an entry to CHANGELOG.md describing this change
- [x] Updated docstrings for any public parameters / methods / classes

### DIFF
--- a/docs/userguides/departingemployee.md
+++ b/docs/userguides/departingemployee.md
@@ -27,7 +27,6 @@ except Py42UserAlreadyAddedError:
 .. important::
     If the user is already in the Departing Employees list, you will get an `py42.exceptions.Py42UserAlreadyAddedError`.
 
-    If a detection list user profile doesn't exist yet for this user, one will automatically be created before adding the user to the Departing Employees list.
 ```
 
 To remove a user from the Departing Employees list:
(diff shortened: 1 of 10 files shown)


unparalleled-js:
In detectionlists.py, do we still need the `create_user()` method?  removing it by technically be considered a breaking change,  but maybe we can add some deprecation note there?

unparalleled-js:
That flaky cases test is failing again.. I think it can be fixed by changing the precision to be more lenient

timabrmsn:
> In detectionlists.py, do we still need the create_user() method?

I thought about this, but there _might_ be a case where the normal user creation path fails leaving us without a detectionlist profile, so I'd say let's keep it around for now.

unparalleled-js (approved):

PR #304, opened 2021-03-19
added update user method

ceciliastevens:
### Description of Change ###

Currently, py42 can create a user, but cannot update a user that already exists. This change adds a method to allow updates to existing users.

### Issues Resolved ###
n/a, this is a prerequesite for work on parts of https://github.com/code42/code42cli/issues/259

### Testing Procedure ###
1. Create a user in Code42.
2. Create an SDK object using py42.
3. Run `sdk.users.update_user(user_id, username="new_username@test.com", email="new_username@test.com", first_name="new", last_name="name")`
4. Verify using the Code42 console that changes have applied as expected.

### PR Checklist ###

### DIFF
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -346,7 +346,7 @@ class UserService(BaseService):
         `REST Documentation <https://console.us.code42.com/apidocviewer/#User-put>`__
 
         Args:
-            user_uid (int): An ID for the user.
+            user_uid (int): The unique identifier for the user.
             username (str, optional): The username to which the user's username will be changed. Defaults to None.
             email (str, optional): The email to which the user's email will be changed. Defaults to None.
             password (str, optional): The password to which the user's password will be changed. Defaults to None.
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -353,6 +353,7 @@ class UserService(BaseService):
             first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
             last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
             note (str, optional): Descriptive information about the user. Defaults to None.
+            quota (int, optional): The quota in bytes that limits the user's archive size.
 
         Returns:
             :class:`py42.response.Py42Response`
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -339,7 +339,7 @@ class UserService(BaseService):
         password=None,
         first_name=None,
         last_name=None,
-        note=None,
+        notes=None,
         quota=None,
     ):
         """Updates an existing user.
@@ -352,7 +352,7 @@ class UserService(BaseService):
             password (str, optional): The password to which the user's password will be changed. Defaults to None.
             first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
             last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
-            note (str, optional): Descriptive information about the user. Defaults to None.
+            notes (str, optional): Descriptive information about the user. Defaults to None.
             quota (int, optional): The quota in bytes that limits the user's archive size. Defaults to None.
 
         Returns:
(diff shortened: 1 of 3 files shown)


src/py42/services/users.py
```diff
@@ -330,3 +330,42 @@ def remove_role(self, user_id, role_name):
...
+    def update_user(
+        self,
+        user_id,
+        username=None,
+        email=None,
+        password=None,
+        first_name=None,
+        last_name=None,
+        note=None,
+        quota=None,
```
unparalleled-js, line 343: Missing doc str for quote

src/py42/services/users.py
```diff
@@ -330,3 +330,42 @@ def remove_role(self, user_id, role_name):
...
+    ):
+        """Updates an existing user.
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#User-put>`__
+
+        Args:
+            user_id (int): An ID for the user.
```
unparalleled-js, line 349: This looks like the `userUid`.  In the users service, we still refer to `user_uid` since that will match the response, and we don't want to confuse with the legacy ID.

ceciliastevens: the REST endpoint can accept user_uid or user_id, I used user_id for consistency with the other methods in py42, would you like me to change it to user_uid instead?

src/py42/services/users.py
ceciliastevens: Updated.

src/py42/services/users.py
unparalleled-js: I think we tend to use `user_uid` whenever possible, so yes I would prefer it there.  The methods that located _near_ this method happen to use the legacy ID because they have to, but other methods in this service use the `user_uid`.

src/py42/services/users.py
```diff
@@ -353,6 +353,7 @@ def update_user(
             first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
             last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
             note (str, optional): Descriptive information about the user. Defaults to None.
+            quota (int, optional): The quota in bytes that limits the user's archive size.
```
unparalleled-js, line 356: Missing `Defaults to None.`

src/py42/services/users.py
unparalleled-js: Obligation _or_ mistake, in case there are any that slipped through when not necessary

src/py42/services/users.py
ceciliastevens: updated

src/py42/services/users.py
ceciliastevens: Makes sense. I've updated it to use uid.

src/py42/services/users.py
```diff
@@ -330,3 +330,43 @@ def remove_role(self, user_id, role_name):
...
+        `REST Documentation <https://console.us.code42.com/apidocviewer/#User-put>`__
+
+        Args:
+            user_uid (int): A Code42 user UID.
+            username (str, optional): The username to which the user's username will be changed. Defaults to None.
+            email (str, optional): The email to which the user's email will be changed. Defaults to None.
+            password (str, optional): The password to which the user's password will be changed. Defaults to None.
+            first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
+            last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
+            note (str, optional): Descriptive information about the user. Defaults to None.
```
unparalleled-js, line 355: Can we make `note` plural?  `notes`

That is how it is in the `create_user()` method

ceciliastevens: Done!

src/py42/services/users.py
```diff
@@ -330,3 +330,43 @@ def remove_role(self, user_id, role_name):
...
+            first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
+            last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
+            notes (str, optional): Descriptive information about the user. Defaults to None.
+            quota (int, optional): The quota in bytes that limits the user's archive size. Defaults to None.
+
+        Returns:
+            :class:`py42.response.Py42Response`
+        """
+
+        uri = u"/api/User/{}?idType=uid".format(user_uid)
```
unparalleled-js, line 362: I think this is fine, but usually we let the `params=` argument take care of query parameters

unparalleled-js (approved):

PR #307, opened 2021-03-25
Fix arg name

unparalleled-js:
### Description of Change ###

Seems like the arg name for `quota` is incorrect. This fixes it as well as makes the public facing one more apparent at what it is 

### Issues Resolved ###

- closes #

### PR Checklist ###

### DIFF
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -340,13 +340,13 @@ class UserService(BaseService):
         first_name=None,
         last_name=None,
         notes=None,
-        quota=None,
+        archive_size_quota_bytes=None,
     ):
         """Updates an existing user.
         `REST Documentation <https://console.us.code42.com/apidocviewer/#User-put>`__
 
         Args:
-            user_uid (int): A Code42 user UID.
+            user_uid (str or int): A Code42 user UID.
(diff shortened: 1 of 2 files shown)


src/py42/services/users.py
```diff
@@ -367,6 +367,6 @@ def update_user(
             u"firstName": first_name,
             u"lastName": last_name,
             u"notes": notes,
-            u"quota": quota,
+            u"quotaInBytes": archive_size_quota_bytes,
```
unparalleled-js, line 370: @ceciliastevens I believe this is supposed to be `quotaInBytes`, correct?  `quota` was not updating anything whereas this property does.

src/py42/services/users.py
```diff
@@ -340,13 +340,13 @@ def update_user(
         first_name=None,
         last_name=None,
         notes=None,
-        quota=None,
+        archive_size_quota_bytes=None,
```
alanag13, line 343: isn't this a breaking change?

alanag13: oh wait, this isn't released yet. carry on.

alanag13 (approved):

PR #308, opened 2021-03-25
Fix arg doc

unparalleled-js:
### Description of Change ###

I forgot to update the arg name in the doc str so this does that

### Issues Resolved ###

- closes #

### PR Checklist ###

### DIFF
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -353,7 +353,7 @@ class UserService(BaseService):
             first_name (str, optional): The first name to which the user's first name will be changed. Defaults to None.
             last_name (str, optional): The last name to which the user's last name will be changed. Defaults to None.
             notes (str, optional): Descriptive information about the user. Defaults to None.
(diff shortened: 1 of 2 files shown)


kiran-chaudhary (approved):
Change LGTM, though I would not mention the unit 'bytes' in the arg name!  Reason being, unit is explained in the docstring and in future if the implementation detail changes to KB, MB etc we need not have to rename the argument name. so `archive_size_quota` or `quota_archive_size` whichever suits best seems more apt.

Approved, otherwise.

EDIT: Need to add/change a test case

kiran-chaudhary:
Also a unit test should be changed if a function argument has changed, I see a change only in one src file!

kiran-chaudhary (changes requested):
Missing unit test!?

timabrmsn:
The existing unit test has the correct arg name, the only change here was the docstring which was incorrect.

timabrmsn (approved):

kiran-chaudhary (approved):

PR #313, opened 2021-04-01
Custom err for username != email and fix unraised 500s

unparalleled-js:
### Description of Change ###

* Adds customer error handling for `update_user()` when trying to update the username to a non-email.
* Fixes bug in `create_user()` method where the 500 errors would not raise
* Fixed test that would not testing anything
* Adds missing tests

### Issues Resolved ###
n/a

### Testing Procedure ###

* Try updating a user's username to something that is not an email in a cloud env and notice the custom error

### PR Checklist ###

### DIFF
--- a/src/py42/services/users.py
+++ b/src/py42/services/users.py
@@ -3,6 +3,7 @@ from py42._compat import quote
 from py42.exceptions import Py42BadRequestError
 from py42.exceptions import Py42InternalServerError
 from py42.exceptions import Py42UserAlreadyExistsError
+from py42.exceptions import Py42UsernameMustBeEmailError
 from py42.services import BaseService
 from py42.services import handle_active_legal_hold_error
 from py42.services.util import get_all_pages
@@ -58,6 +59,7 @@ class UserService(BaseService):
         except Py42InternalServerError as err:
             if u"USER_DUPLICATE" in err.response.text:
                 raise Py42UserAlreadyExistsError(err)
+            raise
 
     def get_by_id(self, user_id, **kwargs):
         """Gets the user with the given ID.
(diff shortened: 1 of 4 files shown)


src/py42/services/users.py
```diff
@@ -58,6 +59,7 @@ def create_user(
         except Py42InternalServerError as err:
             if u"USER_DUPLICATE" in err.response.text:
                 raise Py42UserAlreadyExistsError(err)
+            raise
```
unparalleled-js, line 62: This is the fix for what is listed in the CHANGELOG

tests/services/test_users.py
```diff
@@ -61,12 +64,20 @@ def put_api_mock_response(self, mocker):
...
-    def post_api_mock_error_response(self, mocker):
-        response = mocker.MagicMock(spec=Response)
-        response.status_code = 500
-        response.encoding = "utf-8"
-        response.text = MOCK_ERROR_text
-        return Py42Response(response)
+    def internal_server_error(self, mocker):
+        base_error = mocker.MagicMock(spec=HTTPError)
+        base_error.response = mocker.MagicMock(spec=Response)
+        return Py42InternalServerError(base_error)
```
unparalleled-js, line 70: The error was never getting raised in the methods, and the tests were not really correctly checking so anyway (false positives)

src/py42/services/users.py
timabrmsn: wouldn't the `Py42UsernameMustBeEmailError` also make sense here?

unparalleled-js: Hm yes I would think so, let me see

unparalleled-js: Just checked -- we get a slightly different and less-specific error message in the response.

```
Response content: [{"name":"USERNAME_INVALID","description":"An error has occurred. See server logs for more information.","objects":[]}]
```

So it is not clear of the exact reasons from the error message alone

kiran-chaudhary (approved):
LGTM

PR #320, opened 2021-04-15
Log response when debug level is set

kiran-chaudhary:
### Description of Change ###

Log response when debug level is set.

### Issues Resolved ###

- closes #INTEG-988

### Testing Procedure ###

Regression tests and test when log level is set to debug level or above, response is printed in the logs. Earlier only requests were getting logged.

### PR Checklist ###

- [NA ] Add an entry to CHANGELOG.md describing this change


src/py42/services/_connection.py
```diff
@@ -275,6 +275,7 @@ def _handle_error(method, url, response):
         msg = u"No response was returned for {} request to {}.".format(method, url)
         raise Py42Error(msg)
 
+    debug.logger.debug(u"Response\n", str(response))
```
kiran-chaudhary, line 278: would this suffice?!!

unparalleled-js: Looking at the ticket, it says we want to log every time debug is turned on.
This logging should happen outside of the `_handle_error()` method

unparalleled-js: Line 192/193 area would work

unparalleled-js: Do you get any output from this though? When I do

```
            debug.logger.debug("TEST")
            debug.logger.debug(u"Response\n", str(response))
            debug.logger.debug("TEST")
```

The output from that part is just :

```
TEST
TEST
```

kiran-chaudhary: it did work for me.

kiran-chaudhary: "_it says we want to log every time debug is turned on_"
As per logging module implementation, we need not check log level explicitly, for all level "debug and above" all debug messages would be displayed.

kiran-chaudhary: I just missed the return statement for success cases and was under the impression that `handle_error` is called always irrespective of whether there is an error or not, and does nothing if there is no error.

Changed now.

unparalleled-js: I still am not seeing any Response logging.
I see the requests getting logged, as they did before, but the specific line you added is not showing up for me

unparalleled-js: I think it is because you are passing multiple args to the `debug` method. You only need to pass it a string (1 arg).

Also, the __str__ of a response is just `<Response [200]>`. It does not contain any of the response data. We want to include both the status code _and_ the response data.

src/py42/services/_connection.py
```diff
@@ -195,6 +195,8 @@ def request(
                 # setting this manually speeds up read times
                 response.encoding = u"utf-8"
 
+            debug.logger.debug(u"Response\n {0}".format(str(response)))
```
timabrmsn, line 198: Can this be like this instead? No need for a newline, and gives the status and text of response. 
```
debug.logger.debug(u"Response: status={0}, data={1}".format(response.status_code, response.text))
```

alanag13:
@kiran-chaudhary @unparalleled-js @timabrmsn  I think the best way to handle this will be to change when `_print_request` gets called to be after the response is received instead of before. This allows us to include other useful info such as the status code.

src/py42/services/_connection.py
timabrmsn: I suppose we should also first check if the response is streamed and not use `.text` in that case, so it doesn't block and consume the stream just to log.

alanag13:
reecheckcla

unparalleled-js:
@alanag13 re: moving the `_print_request()` movement suggestion

My streaming thoughts:

If a request times out, would we still log the request data? I think that is an http error so therefore yes.

If the process itself fails while waiting for a response, the request would never been logged.

My gut says that if we can log something earlier, let's not hold off and therefore let's log request and responses separately.

src/py42/services/_connection.py
```diff
@@ -195,6 +195,15 @@ def request(
                 # setting this manually speeds up read times
                 response.encoding = u"utf-8"
 
+            if response is not None:
+                debug.logger.debug(
+                    u"Response: status={}, data={}".format(
+                        response.status_code, response.text
+                    )
+                )
```
unparalleled-js, line 203: The part from lines 207-208 should be moved here because we don't need to check if the `response is not None` twice.

```
            if response is not None:
                debug.logger.debug(
                    u"Response: status={}, data={}".format(
                        response.status_code, response.text
                    )
                )
                if 200 <= response.status_code <= 399:
                  return Py42Response(response)
           else:
              ...
```

alanag13:
> If a request times out, would we still log the request data? I think that is an http error so therefore yes.
> 
> If the process itself fails while waiting for a response, the request would never been logged.
> 
> My gut says that if we can log something earlier, let's not hold off and therefore let's log request and responses separately.

I'm not opposed to keeping both, but if we do, they should probably at different levels of verbosity, otherwise logs will _usually_ have the double the amount of lines that someone actually needs.

imo, it's more likely that users will want response info, so I'd error toward that being the lower level of verbosity.

src/py42/services/_connection.py
kiran-chaudhary: refactored. Similar check was done even in next few lines as well, encompassed all of them into a single block.

kiran-chaudhary:
> > If a request times out, would we still log the request data? I think that is an http error so therefore yes.
> > If the process itself fails while waiting for a response, the request would never been logged.
> > My gut says that if we can log something earlier, let's not hold off and therefore let's log request and responses separately.
> 
> I'm not opposed to keeping both, but if we do, they should probably be at different levels of verbosity, otherwise logs will _usually_ have double the amount of lines that someone actually needs.
> 
> imo, it's more likely that users will want response info, so I'd error toward that being the lower level of verbosity.

Changed response logging to `error`.

src/py42/services/_connection.py
```diff
@@ -191,16 +191,25 @@ def request(
...
-                if isinstance(self._auth, C42RenewableAuth):
-                    self._auth.clear_credentials()
+            if response is not None:
+                if not stream:
+                    # setting this manually speeds up read times
+                    response.encoding = u"utf-8"
+
+                debug.logger.error(
+                    u"Response: status={}, data={}".format(
+                        response.status_code, response.text
```
timabrmsn, line 201: If the response is a stream, accessing response.text here blocks to consume the stream, defeating the purpose of streaming.

kiran-chaudhary: Refactored, logs response when response is not a stream! Do we want to log stream response too?

timabrmsn: Yeah we should log the response line whether it's a stream or not, but if it is a stream, instead of accessing `response.text` and consuming the stream, we should swap it out with `<streamed>` or something to indicate that it's a streamed response that may not be consumed yet.

kiran-chaudhary: added.

src/py42/services/_connection.py
```diff
@@ -191,16 +191,31 @@ def request(
...
-
-            if response is not None and response.status_code == 401:
-                if isinstance(self._auth, C42RenewableAuth):
-                    self._auth.clear_credentials()
+            if response is not None:
+                if not stream:
+                    # setting this manually speeds up read times
+                    response.encoding = u"utf-8"
+
+                    debug.logger.error(
```
timabrmsn, line 199: These should be logged at debug level, not error. We don't want _every_ response sent to stderr by default, only when the user has enabled debug logging.

kiran-chaudhary: based on this [conversation](https://github.com/code42/py42/pull/320#issuecomment-825029146) I set it to error.

unparalleled-js (approved):

timabrmsn:
@unparalleled-js @alanag13 

One last change, I decided logging the response data @ info level was still too high (since things like v3 tokens would then also get logged). This logs the response status at info but moves the data to debug.

So at debug a full request's log would look like:

```
>>> sdk.users.get_current()
GET     https://www.crashplan.com/api/User/my
Response status: 200
Response data: {"data":{"<data here>"}}
```

alanag13 (approved):

PR #322, opened 2021-04-21
Prevent alert note clearing

unparalleled-js:
### Description of Change ###

Ignores `note` parameter for `alerts.update_state()` when given `None`. Before, it would default to the empty string and therefore clear the existing note.

### Issues Resolved ###
no filed issue

### Testing Procedure ###

Before:
Let `alert_id` refer to the ID of an alert in Code42 where the `note.message` is set.
Call `sdk.alerts.update_state("PENDING", [alert_id])`
Notice now that the alert's note.message was cleared.

After:
Let `alert_id` refer to the ID of an alert in Code42 where the `note.message` is set.
Call `sdk.alerts.update_state("PENDING", [alert_id])`
Notice now that the alert's note.message was not affected.

### PR Checklist ###


CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## 1.14.0 - 2021-04-21
```
unparalleled-js, line 11: I'd like to release this right away

unparalleled-js:
I imagine changing the default value is considered a breaking change, but in this case, I think the intent has not changed - we want to default to nothing and we want the request to succeed. Plus, fixing a bug.

alanag13 (approved):

kiran-chaudhary (approved):

PR #323, opened 2021-04-21
switch to debug logs only

unparalleled-js:
### Description of Change ###

Changing the `.error` method to `.debug` for logging http errors because it looks really bad when doing bulk operations and depending on CLI output for progress, like we do in the CLI.

Here is what the output looks like (no `-d` was given for the CLI - this always happens no matter what)

```
Updating alerts:  [########################------------]   66%  00:00:01  2 succeeded, 0 failed out of 3Failure in HTTP call 404 Client Error: Not Found for url: https://alert-service-east.us.code42.com/svc/api/v1/update-state. Response content: {
  "pop-bulletin": {
    "Details": []
  }
}
Updating alerts:  [####################################]  100%  2 succeeded, 1 failed out of 3 
```

### Issues Resolved ###
No official issue

### Testing Procedure ###
Use this in the CLI as its py42 dependency. Do something erroronous.  Rely on logs instead of stdout.

### PR Checklist ###

- [?] Add an entry to CHANGELOG.md describing this change

### DIFF
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -60,7 +60,7 @@ class Py42HTTPError(Py42ResponseError):
             message = u"Failure in HTTP call {}. {}".format(
                 str(exception), response_content
             )
-            debug.logger.error(message)
+            debug.logger.debug(message)
 
         super(Py42HTTPError, self).__init__(exception.response, message)
 


unparalleled-js:
I don't enjoy seeing `pop-bulletin` and a huge JSON blob for CLI errors.
If we really want this in our py42 scripts, I can try to come up with a way to parametrize silencing them.
I only wish to see these in the error log file, not in stdout, for scripts / CLI.

src/py42/exceptions.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, exception, message=None):
             message = u"Failure in HTTP call {}. {}".format(
                 str(exception), response_content
             )
-            debug.logger.error(message)
+            debug.logger.debug(message)
```
alanag13, line 63: yeah, this level of logging seems more appropriate because us hitting this line does indicate a problem with `py42` itself.

alanag13 (approved):

PR #327, opened 2021-05-07
Preparation/update roles api

unparalleled-js:
### Description of Change ###

Uses the newer API for getting roles.

### Testing Procedure ###
Make a call with the new API. Save the response text. Make a call with the old API. Save the response text. Verify they are the same.

Go into the console, go to Administration -> Users. Select a user, select that gear, pretend you are about to add a role with dev tools open, notice the API call used to get the roles, make sure it is the same as the one I am using here.

### PR Checklist ###


unparalleled-js:
Failing because I need to update the mock servers... BRB

PR #328, opened 2021-06-04
make create just call get instead

unparalleled-js:
### Description of Change ###

Instead of returning a noop 200 response, we could return the response from `get()`. That way, it will contain the same response data as before.

One issue with this is that getting a detection list profile by username is more delayed than getting a profile by ID, for newly created users. However, anybody's script that was calling `create` for newly created users would have already been failing from the API, so this is fine imo.

### Issues Resolved ###
n/a

### Testing Procedure ###
Call the method `sdk.detectionlists.create_user()` and make sure the response is 200 for users that exist.

### PR Checklist ###

### DIFF
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -281,9 +281,10 @@ class Py42UnableToCreateProfileError(Py42BadRequestError):
 
     def __init__(self, exception, username):
         message = (
-            u"Detection-list profiles are now created automatically on the server."
+            u"Detection-list profiles are now created automatically on the server. "
             u"Unable to find a detection-list profile for '{}'. "
-            u"It is possibly still being created.".format(username)
+            u"It is possibly still being created if you just recently created the "
+            u"Code42 user.".format(username)
         )
         super(Py42UnableToCreateProfileError, self).__init__(exception, message)
 
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -277,7 +277,7 @@ class Py42UnableToCreateProfileError(Py42BadRequestError):
     """An error raised when trying to call the method for creating a detection-list
     user when the user does not exist or is currently awaiting the profile to get
     created on the back-end. Note: you are no longer able to create detection-list
-    profiles using the API and the py42 only returns already existing profiles."""
+    profiles using the API and py42 only returns already existing profiles."""
 
     def __init__(self, exception, username):
         message = (
--- a/src/py42/exceptions.py
+++ b/src/py42/exceptions.py
@@ -273,6 +273,21 @@ class Py42UserNotOnListError(Py42NotFoundError):
         super(Py42NotFoundError, self).__init__(exception, message)
 
 
+class Py42UnableToCreateProfileError(Py42BadRequestError):
+    """An error raised when trying to call the method for creating a detection-list
+    user when the user does not exist or is currently awaiting the profile to get
+    created on the back-end. Note: you are no longer able to create detection-list
+    profiles using the API and the py42 only returns already existing profiles."""
+
+    def __init__(self, exception, username):
+        message = (
+            u"Detection-list profiles are now created automatically on the server."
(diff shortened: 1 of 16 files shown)


timabrmsn:
Should we also delete the create method from the `user_profile_service`?

timabrmsn:
Also, what about having a custom exception raised if the user isn't found in the create call, as that could be confusing if someone tries to create a profile for a non-existent and/or typo'd username and just gets a `Could not find user` error response.

alanag13:
> Should we also delete the create method from the `user_profile_service`?

I don't think that service gets exposed directly by the SDK object in any way, so that should be safe

tests/services/detectionlists/test_user_profile.py
```diff
@@ -28,56 +26,30 @@ def mock_user_client(self, mock_connection, user_context, py42_response):
 
     @pytest.fixture
     def mock_get_by_id_fails(self, mocker, mock_connection):
-        response = mocker.MagicMock(spec=Response)
-        response.status_code = 400
-        exception = mocker.MagicMock(spec=HTTPError)
-        exception.response = response
-        mock_connection.post.side_effect = Py42BadRequestError(exception)
```
unparalleled-js, line 35: Just trying out the new helper method in a fe more spots

tests/services/detectionlists/test_user_profile.py
```diff
@@ -28,56 +26,30 @@ def mock_user_client(self, mock_connection, user_context, py42_response):
...
         user_client = UserService(mock_connection)
-        response = mocker.MagicMock(spec=Response)
-        response.status_code = 400
-        response.text = CLOUD_ALIAS_LIMIT_EXCEEDED_ERROR_MESSAGE
-        exception = mocker.MagicMock(spec=HTTPError)
-        exception.response = response
-        mock_connection.post.side_effect = Py42BadRequestError(exception)
-        return user_client
-
-    def test_create_posts_expected_data(
```
unparalleled-js, line 63: deleted test as well, from deleted method

src/py42/exceptions.py
```diff
@@ -273,6 +273,21 @@ def __init__(self, exception, user_id, list_name):
         super(Py42NotFoundError, self).__init__(exception, message)
 
 
+class Py42UnableToCreateProfileError(Py42BadRequestError):
+    """An error raised when trying to call the method for creating a detection-list
+    user when the user does not exist or is currently awaiting the profile to get
+    created on the back-end. Note: you are no longer able to create detection-list
+    profiles using the API and the py42 only returns already existing profiles."""
```
timabrmsn: unneeded "the" in `using the API and the py42`

src/py42/exceptions.py
```diff
@@ -273,6 +273,21 @@ def __init__(self, exception, user_id, list_name):
...
+
+    def __init__(self, exception, username):
+        message = (
+            u"Detection-list profiles are now created automatically on the server."
+            u"Unable to find a detection-list profile for '{}'. "
+            u"It is possibly still being created.".format(username)
```
timabrmsn, line 286: Maybe `It is possibly still being created if you just recently created the Code42 user`

src/py42/exceptions.py
```diff
@@ -273,6 +273,21 @@ def __init__(self, exception, user_id, list_name):
...
 
+class Py42UnableToCreateProfileError(Py42BadRequestError):
+    """An error raised when trying to call the method for creating a detection-list
+    user when the user does not exist or is currently awaiting the profile to get
+    created on the back-end. Note: you are no longer able to create detection-list
+    profiles using the API and the py42 only returns already existing profiles."""
+
+    def __init__(self, exception, username):
+        message = (
+            u"Detection-list profiles are now created automatically on the server."
```
timabrmsn, line 284: Could use a space after the period here.

timabrmsn (approved):
Just a few comments on the exception text but otherwise LGTM

PR #336, opened 2021-06-18
Honor microseconds in Alert DateObserved query  + fix py2 timestamp unicode bug

unparalleled-js:
### Description of Change ###

I noticed this bug while working on Phantom.
Alert Search output contains `createdAt` with microsecond precision, such as `"2021-04-18T10:02:36.3198680Z"`.
However, when I would plug an equivalent float timestamp into the Action for search alerts, it would shorten it unnecessarily to `"2021-04-18T10:02:36.319Z"`. The bug appears to be in py42 and this PR addresses it. The API supports this level of precision, so we should allow it in our SDK, or we could potentially be missing alerts.

Additionallty, fixed a PR in py42 when using python2.7 that forced users to have to use unicode strings when crafting queries

A CLI Pr will need to be merged as well, opening that next to prevent failing nightly builds.

### Testing Procedure ###
Make a filter and notice it is using microsecond precision:

First, open the code42cli shell:

```bash
code42 shell
```

```python
test_epoch = 1599653541.001002
test_filter = DateObserved.on_or_after(test_epoch)
print(test_filter)

>>> {"filterClause":"AND", "filters":[{"operator":"ON_OR_AFTER", "term":"createdAt", "value":"2020-09-09T12:12:21.001002Z"}]}


# Create a filter and execute the search and notice that it works and stuff
test_query = AlertQuery(test_filter)
sdk.alerts.search(test_query)
```

### PR Checklist ###

### DIFF
--- a/src/py42/_compat.py
+++ b/src/py42/_compat.py
@@ -16,7 +16,9 @@ if is_py2:
     from urlparse import urljoin
     from urlparse import urlparse
 
+    original_str = str
     str = unicode
+    str_options = (original_str, str)
 
     import repr as reprlib
 
@@ -33,6 +35,7 @@ else:
     from urllib.parse import urlencode
 
     str = str
+    str_options = str
 
     import reprlib
 
--- a/src/py42/_compat.py
+++ b/src/py42/_compat.py
@@ -16,9 +16,7 @@ if is_py2:
     from urlparse import urljoin
     from urlparse import urlparse
 
-    _original_str = str
     str = unicode
-    str_options = (_original_str, str)
 
     import repr as reprlib
 
@@ -35,7 +33,6 @@ else:
     from urllib.parse import urlencode
 
     str = str
-    str_options = (str,)
 
     import reprlib
 
--- a/src/py42/_compat.py
+++ b/src/py42/_compat.py
@@ -16,9 +16,9 @@ if is_py2:
     from urlparse import urljoin
     from urlparse import urlparse
 
-    original_str = str
+    _original_str = str
     str = unicode
-    str_options = (original_str, str)
+    str_options = (_original_str, str)
 
     import repr as reprlib
 
--- a/src/py42/_compat.py
+++ b/src/py42/_compat.py
@@ -35,7 +35,7 @@ else:
     from urllib.parse import urlencode
 
     str = str
-    str_options = str
+    str_options = (str,)
 
     import reprlib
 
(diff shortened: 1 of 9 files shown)


src/py42/_compat.py
```diff
@@ -16,7 +16,9 @@
     from urlparse import urljoin
     from urlparse import urlparse
 
+    original_str = str
     str = unicode
+    str_options = (original_str, str)
```
unparalleled-js, line 21: Doing this because of the following flow:

1.) py42 module would import `str` from compat, which would really be unicode in python2.7
2.) Then, a client module would create a str (normally, not unicode) in their python2.7 script and they would pass that str into a query filter.
3.) That query filter would turn their perfectly normal str into the value `None` because it was not of an expected type (datetime, float, or unicode)... It was non intuitive of what was going wrong.

Now, the new flow is for our pythong module to import `str_options` in this case for using `isinstance()` to verify that argument is one of the two str types.

src/py42/util.py
```diff
@@ -3,9 +3,10 @@
 import json
 from datetime import datetime
 
-from py42._compat import str
+from py42._compat import string_type
```
unparalleled-js, line 6: This was the py2 releated bug culprit

src/py42/util.py
```diff
@@ -3,9 +3,10 @@
...
 
-_MICROSECOND_FORMAT = u"%Y-%m-%dT%H:%M:%S.%fZ"
+MILLISECOND_FORMAT = u"%Y-%m-%dT%H:%M:%S.%f"
+MICROSECOND_FORMAT = u"%Y-%m-%dT%H:%M:%S.%fZ"
```
peterbriggs42, line 9: My understanding is that the `Z` signifies that the timestamp is in UTC, rather than suggesting anything about the precision of the seconds value.

unparalleled-js: Yeah it is confusingly being used here...
The millisecond one truncates itself in the places it gets used

unparalleled-js: I can try and find a way to reword this or re-do something somewhere

unparalleled-js: Just pushed something ... does this wok? I am a little nervous changing this now.

timabrmsn: Yeah, I think that should work fine

timabrmsn (approved):

PR #341, opened 2021-06-29
Escape unescaped double-quotes in tokens

timabrmsn:
### Description of Change ###

Automatically escapes unescaped double-quote characters in page_tokens, allowing end-users to just pass the received token from previous request without having to worry about modifying it themselves. As some eventIds (which is what the `nextPgToken` value is) contain double-quote characters. 

### Testing Procedure ###
Build an FFS query that returns more than 10k results (or lower page size to allow for response to generate a `nextPgToken` value. Run subsequent requests until a `nextPgToken` value contains a `"` character.  Example eventId:

```
1XpE25X_jybWQVvWDPYlRLDdB5qbV0sET_"iIuugeBvajEpS7jVVV6bKPtQvJb9NR0eDfu6E-hDxr8/jE0vdQ411emJ4__ipmdtOfyGiwQ"
```

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Fixed
+
+- `sdk.securitydata.search_all_file_events()` automatically escapes double-quote characters in the `page_token` param when passed.
+
 ## 1.15.1 - 2021-06-22
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Fixed
 
-- `sdk.securitydata.search_all_file_events()` automatically escapes double-quote characters in the `page_token` param when passed.
+- `sdk.securitydata.search_all_file_events()` now automatically escapes double-quote characters in the `page_token` param when passed.
 
 ## 1.15.1 - 2021-06-22
 
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -471,7 +471,7 @@ class PlanStorageInfo(object):
         return self._node_guid
 
 
-def _escape_quote_chars_in_token(string):
+def _escape_quote_chars_in_token(token):
     """
     The `nextPgToken` returned in Forensic Search requests with > 10k results is the eventId
     of the last event returned in the response. Some eventIds have double-quote chars in
@@ -482,5 +482,5 @@ def _escape_quote_chars_in_token(string):
     return re.sub(
         pattern=unescaped_quote_pattern,
         repl=lambda match: match.group().replace('"', r'\"'),
-        string=string
+        string=token
     )
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -481,6 +481,6 @@ def _escape_quote_chars_in_token(token):
 
     return re.sub(
         pattern=unescaped_quote_pattern,
-        repl=lambda match: match.group().replace(u'"', r'\"'),
-        string=token
+        repl=lambda match: match.group().replace(u'"', r"\""),
+        string=token,
     )
--- a/src/py42/clients/securitydata.py
+++ b/src/py42/clients/securitydata.py
@@ -1,3 +1,4 @@
+import re
 from threading import Lock
 
 from py42.exceptions import Py42ChecksumNotFoundError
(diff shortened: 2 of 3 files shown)


src/py42/clients/securitydata.py
```diff
@@ -468,3 +469,18 @@ def destination_guid(self):
...
+
+def _escape_quote_chars_in_token(string):
+    """
+    The `nextPgToken` returned in Forensic Search requests with > 10k results is the eventId
+    of the last event returned in the response. Some eventIds have double-quote chars in
+    them, which need to be escaped when passing the token in the next search request.
+    """
+    unescaped_quote_pattern = r'[^\\]"'
+
+    return re.sub(
```
unparalleled-js, line 482: Why do we need the `match.group()`, why can't just use the replace call on its own?

unparalleled-js: Like, 

```python
x = "asdfsdf\"asdf\"asdf335425'"
x.replace('"', r'\"')
```

outputs `'asdfsdf\\"asdf\\"asdf335425\''`, wouldn't that suffice?

unparalleled-js: Oh I see it is weird, it like adds an extra slash... ok.

tests/clients/test_securitydata.py
```diff
@@ -1384,3 +1384,33 @@ def test_search_all_file_events_calls_search_with_expected_params_when_pg_token_
             data='{"groupClause":"AND", "groups":[], "pgToken":"abc", "pgSize":10000}',
         )
         assert response is successful_response
+
+    @pytest.mark.parametrize(
```
unparalleled-js, line 1388: Why parametrize the test if there is only a single test case?

src/py42/clients/securitydata.py
```diff
@@ -468,3 +469,18 @@ def destination_guid(self):
     def node_guid(self):
         """The GUID of the storage node containing the archive."""
         return self._node_guid
+
+
+def _escape_quote_chars_in_token(token):
+    """
+    The `nextPgToken` returned in Forensic Search requests with > 10k results is the eventId
```
unparalleled-js, line 476: Does adding events to cases work ok without having to escape the quotes or should we do the same there?

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Fixed
+
+- `sdk.securitydata.search_all_file_events()` automatically escapes double-quote characters in the `page_token` param when passed.
```
unparalleled-js, line 15: Someone might get the impression that the the automatic-escaping _was_ the bug that got fixed.

src/py42/clients/securitydata.py
```diff
@@ -468,3 +469,18 @@ def destination_guid(self):
...
+    of the last event returned in the response. Some eventIds have double-quote chars in
+    them, which need to be escaped when passing the token in the next search request.
+    """
+    unescaped_quote_pattern = r'[^\\]"'
+
+    return re.sub(
+        pattern=unescaped_quote_pattern,
+        repl=lambda match: match.group().replace(u'"', r'\"'),
```
unparalleled-js, line 484: Do we need to worry about escaped-single quotes at all?

src/py42/clients/securitydata.py
timabrmsn: oh, that's a good thought. I'll check.

CHANGELOG.md
timabrmsn: I'll clarify.

src/py42/clients/securitydata.py
timabrmsn: Probably not until we actually find `eventIds` with single-quotes. I think the double-quotes being in some IDs is likely a bug on the backend.

tests/clients/test_securitydata.py
timabrmsn: There are two cases, we want to add an escape to a string that isn't already escaped: `('1234_"abcde"', r'1234_\"abcde\"')` and leave untouched if the string is already escaped: `(r'1234_\"abcde\"', r'1234_\"abcde\"')`.

unparalleled-js: Oh whoops, I see. That might be more readable as a separate tests named `test_search_all_file_events_does_not_escape_already_escaped_quotes()`

timabrmsn: That's true, I'll make it two tests.

src/py42/clients/securitydata.py
timabrmsn: Looks like these events break the cases API when the eventId is in the URL.. gonna need to file a bug.

unparalleled-js (approved):
This seems like a good patch for now, even though we can't fix it the URL path parameters for adding events to cases.

PR #346, opened 2021-07-15
fixes/utilize-requests-json-param

unparalleled-js:
### Description of Change ###

Debug logging assumed a dictionary for data during the logging call, which messed with the output format of the logs. This fixes that and makes the data logging the same as the params logging.

### Issues Resolved ###
n/a

### Testing Procedure ###
Turn on debugging (or use `-d` in the CLI and notice the logs look nicer now for data-based requests.
You can add a print statement in `_print_request()` before these changes to verify the type of data before it is passed to json.dumps (it was a str, but json.dumps assumes it is dictionary).
After these changes, it is now a dictionary.

### PR Checklist ###

### DIFF
--- a/src/py42/services/_connection.py
+++ b/src/py42/services/_connection.py
@@ -231,6 +231,8 @@ class Connection(object):
 
         if json is not None:
             data = json_lib.dumps(json)
+        elif data is not None:
+            json = json_lib.loads(data)
 
         headers = headers or {}
         headers.update(self._headers)
@@ -251,7 +253,7 @@ class Connection(object):
             hooks=hooks,
         )
 
-        _print_request(method, url, params=params, data=data)
+        _print_request(method, url, params=params, data=json)
         return self._session.prepare_request(request)
 
     def _get_host_address(self):
--- a/src/py42/services/_connection.py
+++ b/src/py42/services/_connection.py
@@ -233,9 +233,8 @@ class Connection(object):
         if json is not None:
             data = json_lib.dumps(json)
 
-        if data is not None:
-            if isinstance(data, str):
-                data = data.encode("utf-8")
+        if isinstance(data, str):
+            data = data.encode("utf-8")
 
         headers = headers or {}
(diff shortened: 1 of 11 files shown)


src/py42/services/_connection.py
```diff
@@ -239,6 +240,12 @@ def _prepare_request(
         if u"Accept" not in headers:
             headers.update({u"Accept": u"application/json"})
         headers = _create_user_headers(headers)
+
+        _print_request(method, url, params=params, data=data)
```
unparalleled-js, line 244: Had to shuffle some of this logic around because we wanted to first check if it `str` first (even though we know it alway is), and the logging method logs them as `dicts`, and it would error on byte arrays.

This is why I said we should assume it is always a `str` for now, because we are just going to switch to always use `json` anyway and we would not have to had to mess with some of the logging logic.

Nonetheless, this seems to be working okay.

unparalleled-js:
I was looking at this and thinking about this last night, and I think from what we have learned, it makes way more sense to use `json` and `data` appropriately, so I am working on that now.

src/py42/services/fileevent.py
```diff
@@ -24,9 +24,8 @@ def search(self, query):
             :class:`py42.response.Py42Response`: A response containing the query results.
         """
         try:
-            query_str = str(query)
             uri = u"/forensic-search/queryservice/api/v1/fileevent"
-            return self._connection.post(uri, data=query_str)
+            return self._connection.post(uri, json=dict(query))
```
timabrmsn, line 28: Since our docstring above mentions `query` could be a `FileEventQuery` or `str`, we'll need to try to parse the string into a `FileEventQuery` first if we want to pass to `json`, otherwise this is a breaking change.

unparalleled-js: Thanks, it should now work with str queries. I added a test as well.

timabrmsn (approved):
LGTM, thanks for doing the json arg updates as well!

PR #348, opened 2021-07-26
Add risk indicator filters

alanag13:
### Description of Change ###

Adds RiskIndicator and and RiskSeverity filters to file event searches

### Issues Resolved ###
INTEG-1714

### Testing Procedure ###

```python
import py42
from py42.sdk.queries.fileevents.filters import *
from py42.sdk.queries.fileevents.file_event_query import FileEventQuery

# use se demo account credentials
sdk = py42.sdk.from_local_account("https://console.us.code42.com", "username", "password")

sev_query = RiskSeverity.is_in([RiskSeverity.HIGH, RiskSeverity.LOW])
sev_results = sdk.securitydata.search_file_events(sev_query)

# There are many RiskIndicator constants. You don't need to test all of them, but it's worth trying a couple different ones
# to make sure different results are returned for each one.
# see https://github.com/code42/py42/blob/361e0348543648201cca346e61b0160852b66ed2/src/py42/sdk/queries/fileevents/filters/risk_filter.py
indicator_query = RiskIndicator.is_in([RiskIndicator.ExternalDevices.REMOVABLE_MEDIA, RiskIndicator.ExternalDevices.AIRDROP])
indicator_results = sdk.securitydata.search_file_events(indicator_query)
```

verify that the results are returned without errors and the result count and general content is the same for equivalent queries made in the Forensic Search UI.

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- `RiskIndicator` and `RiskSeverity` filters class to new `py42.sdk.queries.fileevents.filters.risk_filter` module.
+
 ## 1.16.1 - 2021-07-20
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 ### Added
 
-- `RiskIndicator` and `RiskSeverity` filters class to new `py42.sdk.queries.fileevents.filters.risk_filter` module.
+- `RiskIndicator` and `RiskSeverity` filter classes to new `py42.sdk.queries.fileevents.filters.risk_filter` module.
 
 ## 1.16.1 - 2021-07-20
 
--- a/src/py42/sdk/queries/fileevents/filters/risk_filter.py
+++ b/src/py42/sdk/queries/fileevents/filters/risk_filter.py
@@ -3,7 +3,71 @@ from py42.util import get_attribute_keys_from_class
 
 
 class RiskIndicator(FileEventFilterStringField):
-    """Class that filters events by risk indicator."""
+    """Class that filters events by risk indicator.
+
+    Available options are provided as class attributes:
+        - :attr:`RiskIndicator.CloudDataExposures.PUBLIC_CORPORATE_BOX`
+        - :attr:`RiskIndicator.CloudDataExposures.PUBLIC_CORPORATE_GOOGLE_DRIVE`
+        - :attr:`RiskIndicator.CloudDataExposures.PUBLIC_CORPORATE_ONEDRIVE`
+        - :attr:`RiskIndicator.CloudDataExposures.SENT_CORPORATE_GMAIL`
+        - :attr:`RiskIndicator.CloudDataExposures.SHARED_CORPORATE_BOX`
+        - :attr:`RiskIndicator.CloudDataExposures.SHARED_CORPORATE_GOOGLE_DRIVE`
+        - :attr:`RiskIndicator.CloudDataExposures.SHARED_CORPORATE_ONEDRIVE`
+        - :attr:`RiskIndicator.CloudStorageUploads.AMAZON_DRIVE`
+        - :attr:`RiskIndicator.CloudStorageUploads.BOX`
+        - :attr:`RiskIndicator.CloudStorageUploads.DROPBOX`
+        - :attr:`RiskIndicator.CloudStorageUploads.GOOGLE_DRIVE`
+        - :attr:`RiskIndicator.CloudStorageUploads.ICLOUD`
+        - :attr:`RiskIndicator.CloudStorageUploads.MEGA`
+        - :attr:`RiskIndicator.CloudStorageUploads.ONEDRIVE`
+        - :attr:`RiskIndicator.CloudStorageUploads.ZOHO`
+        - :attr:`RiskIndicator.CodeRepositoryUploads.BITBUCKET`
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- `RiskIndicator` and `RiskSeverity` filters class to new `py42.sdk.queries.fileevents.filters.risk_filter` module.
```
unparalleled-js, line 15: Wrong plurality around the phrase `filters class`?

src/py42/sdk/queries/fileevents/filters/risk_filter.py
```diff
@@ -0,0 +1,232 @@
...
+    class CloudDataExposures(object):
+        PUBLIC_CORPORATE_BOX = "Public link from corporate Box"
+        PUBLIC_CORPORATE_GOOGLE_DRIVE = "Public link from corporate Google Drive"
+        PUBLIC_CORPORATE_ONEDRIVE = "Public link from corporate OneDrive"
+        SENT_CORPORATE_GMAIL = "Sent from corporate Gmail"
+        SHARED_CORPORATE_BOX = "Shared from corporate Box"
+        SHARED_CORPORATE_GOOGLE_DRIVE = "Shared from corporate Google Drive"
+        SHARED_CORPORATE_ONEDRIVE = "Shared from corporate OneDrive"
+
+        @staticmethod
```
unparalleled-js, line 98: I wonder why we didn't just make a base Enum class with `classmethod` `all(cls)` that all these classes inherit from (all the filter classes etc). I have used this approach myself and it works the same way as it is now without having to repeat the same method N times.

unparalleled-js:
```
class Enum:
  def all(cls):
    return get_attribute_keys_from_class(cls)
```

src/py42/sdk/queries/fileevents/filters/risk_filter.py
```diff
@@ -0,0 +1,232 @@
...
+            + RiskIndicator.ExternalDevices.choices()
+            + RiskIndicator.FileCategories.choices()
+            + RiskIndicator.MessagingServiceUploads.choices()
+            + RiskIndicator.Other.choices()
+            + RiskIndicator.SocialMediaUploads.choices()
+            + RiskIndicator.UserBehavior.choices()
+        )
+
+    class CloudDataExposures(object):
+        PUBLIC_CORPORATE_BOX = "Public link from corporate Box"
```
unparalleled-js, line 90: These are some strange enums, I am guessing this is what you meant when you said these were not going to change? I wish we could have had them be something else still.

src/py42/sdk/queries/fileevents/filters/risk_filter.py
alanag13: Not every string filter has choices (like file name, file path, username, etc).  Otherwise `choices` would just return nothing for those. 

That's probably why it got done this way originally, but we could still achieve what you're after by creating separate classes for those that do have choice vs those that don't. That's probably a good idea and would remove some cruft, I can create a ticket for it.

src/py42/sdk/queries/fileevents/filters/risk_filter.py
alanag13: Yeah that's exactly what I was talking about, I double checked  and they're not tied to the UI and are not going to change, but I agree they're strange.

src/py42/sdk/queries/fileevents/filters/risk_filter.py
```diff
@@ -0,0 +1,232 @@
...
+        - :attr:`RiskIndicator.FileCategories.SPREADSHEET`
+        - :attr:`RiskIndicator.FileCategories.VIDEO`
+        - :attr:`RiskIndicator.FileCategories.VIRTUAL_DISK_IMAGE`
+        - :attr:`RiskIndicator.FileCategories.ZIP`
+        - :attr:`RiskIndicator.MessagingServiceUploads.FACEBOOK_MESSENGER`
+        - :attr:`RiskIndicator.MessagingServiceUploads.MICROSOFT_TEAMS`
+        - :attr:`RiskIndicator.MessagingServiceUploads.SLACK`
+        - :attr:`RiskIndicator.MessagingServiceUploads.WHATSAPP`
+        - :attr:`RiskIndicator.Other.OTHER`
+        - :attr:`RiskIndicator.MessagingServiceUploads.UNKNOWN`
```
unparalleled-js, line 62: One `MessageServiceUploads` is away from the rest, maybe swap it with the one above?

unparalleled-js: Oh wait, that would should be `:attr:`RiskIndicator.Other.UNKNOWN`

src/py42/sdk/queries/fileevents/filters/risk_filter.py
unparalleled-js: All that it would be is to create base class that classes with `choices` could inherit from, and then they get the `all()` method for free

unparalleled-js (approved):

src/py42/sdk/queries/fileevents/filters/risk_filter.py
alanag13: Yep exactly. Right now there are some `FileEventFilterStringField`s that have choices and some that don't, and we don't want choices to be on the ones that don't, so we cant just add all() to `FileEventFilterStringField`. We would need to create `FileEventFilterStringOptionField` which inherits from `FileEventFilterStringField` and add the method to that, then change everything that has choices to use that class instead.

unparalleled-js: Ah I forgot that there is already a base class in this situation, that does make it a tad more complicated

PR #350, opened 2021-08-02
Feature / alert details helper

timabrmsn:
### Description of Change ###

Adds method `sdk.alerts.get_all_alert_details()` to combine `sdk.alerts.search_all_pages()` and `sdk.alerts.get_details()`, making it painless to get all alert details in chronological order. 

### Testing Procedure ###

- Create an `AlertQuery`
- Pass query to `sdk.alerts.get_all_alert_details(query)`
- Will return a generator of individual alerts with detail included, in ascending chronological order.

- Pass query to `sdk.alerts.get_all_alert_details(query, ascending=False)`
- Will return a generator of individual alerts with detail included, in descending chronological order.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -1,4 +1,8 @@
+import json
+
 from py42.sdk.queries.alerts.filters import AlertState
+from py42.sdk.queries.alerts.alert_query import AlertQuery
+from py42._compat import string_type
 
 
 class AlertsClient(object):
@@ -136,3 +140,32 @@ class AlertsClient(object):
             :class:`py42.response.Py42Response`
         """
         return self._alert_service.get_aggregate_data(alert_id)
+
+    def iter_all_alert_details(self, query, ascending=True):
+        """
+        Helper method that combines `.get_all_pages()` and `.get_details()`.
+        Returns an iterator of all alert details in chronological order.
+
+        Args:
+            query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
+            ascending (bool): Determines if sort ordering should be ascending/descending
+                based on alert creation time. Defaults to True.
+
+        Returns:
+            generator: An object that iterates over alert detail result items.
+        """
+        if isinstance(query, string_type):
+            query = AlertQuery.from_dict(json.loads(query))
+        query.page_size = 100
+        if ascending:
+            query.sort_direction = "asc"
+        pages = self._alert_service.search_all_pages(query)
+        for page in pages:
+            alert_ids = [alert["id"] for alert in page["alerts"]]
+            alert_details = self._alert_service.get_details(alert_ids)
+            for alert in sorted(
+                alert_details["alerts"],
+                key=lambda x: x["createdAt"],
+                reverse=not ascending,
+            ):
+                yield alert
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -140,7 +140,8 @@ class AlertsClient(object):
     def get_all_alert_details(self, query, ascending=True):
         """
         Helper method that combines `.get_all_pages()` and `.get_details()`.
-        Returns an iterator of all alert details in chronological order.
+        Returns an iterator of alert detail objects in chronological order by alert
+        creation date.
 
         Args:
             query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
@@ -153,6 +154,7 @@ class AlertsClient(object):
         """
         query.page_size = 100
         query.sort_direction = "asc" if ascending else "desc"
+        query.sort_key = "CreatedAt"
         pages = self._alert_service.search_all_pages(query)
         for page in pages:
             alert_ids = [alert["id"] for alert in page["alerts"]]
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -139,7 +139,7 @@ class AlertsClient(object):
 
     def get_all_alert_details(self, query, ascending=True):
         """
-        Helper method that combines `.get_all_pages()` and `.get_details()`.
+        Helper method that combines :func:`.get_all_pages()` and :func:`.get_details()`.
         Returns an iterator of alert detail objects in chronological order by alert
         creation date.
 
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -139,8 +139,13 @@ class AlertsClient(object):
 
     def get_all_alert_details(self, query):
         """
-        Helper method that combines :func:`.search_all_pages()` and :func:`.get_details()`.
+        Helper method that combines :func:`.search_all_pages()` and :func:`.get_details()`
+        methods to get alert objects with alert "observations" details populated.
         Returns an iterator of alert detail objects.
+
+        Note: automatically overrides the `page_size` property on the query object to limit
+        search to 100 results per page, as that is the max that :func:`.get_details()` can
+        request at a time.
 
         Args:
             query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
@@ -158,6 +163,6 @@ class AlertsClient(object):
             alert_ids = [alert["id"] for alert in page["alerts"]]
             alert_details = self._alert_service.get_details(alert_ids)
             for alert in sorted(
-                alert_details["alerts"], key=lambda x: x[sort_key], reverse=reverse,
+                alert_details["alerts"], key=lambda x: x.get(sort_key), reverse=reverse,
             ):
                 yield alert
--- a/src/py42/clients/alerts.py
+++ b/src/py42/clients/alerts.py
@@ -154,6 +154,8 @@ class AlertsClient(object):
         """
         query.page_size = 100
         sort_key = query.sort_key[0].lower() + query.sort_key[1:]
+        if sort_key == "alertId":
+            sort_key = "id"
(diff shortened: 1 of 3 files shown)


tests/clients/test_alerts.py
```diff
@@ -92,3 +138,58 @@ def test_alerts_client_calls_get_aggregate_data_with_expected_value_and_param(
...
         alert_client.get_aggregate_data("alert-id")
         mock_alerts_service.get_aggregate_data.assert_called_once_with("alert-id")
+
+    def test_alerts_client_get_all_alert_details_calls_get_details_for_each_page(
+        self, mock_alerts_service_with_pages, mock_alert_rules_service
+    ):
+        mock_alerts_service = mock_alerts_service_with_pages(ascending=True)
+        alert_client = AlertsClient(mock_alerts_service, mock_alert_rules_service)
+        query = AlertQuery()
+        list(alert_client.get_all_alert_details(query))
```
peterbriggs42, line 148: Ahh nice I did not know about this Python idiom for retrieving all the results of a generator! Seems very handy for debugging.

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,32 @@ def get_aggregate_data(self, alert_id):
             :class:`py42.response.Py42Response`
         """
         return self._alert_service.get_aggregate_data(alert_id)
+
+    def get_all_alert_details(self, query, ascending=True):
+        """
+        Helper method that combines `.get_all_pages()` and `.get_details()`.
```
peterbriggs42, line 142: I think this should read "Helper method that combines `.search_all_pages()` and `.get_details()`."

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,32 @@ def get_aggregate_data(self, alert_id):
...
+        Returns an iterator of alert detail objects in chronological order by alert
+        creation date.
```
unparalleled-js, line 144: We could also note that sort-related parameters are ignored on the query, just in case the user is concerned about that for some reason

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,32 @@ def get_aggregate_data(self, alert_id):
             :class:`py42.response.Py42Response`
         """
         return self._alert_service.get_aggregate_data(alert_id)
+
+    def get_all_alert_details(self, query, ascending=True):
```
peterbriggs42, line 140: Should this maybe be called `search_all_alert_details` or `search_alert_details` instead? As written might give the impression that we are retrieving **all** alert details in the database

src/py42/clients/alerts.py
unparalleled-js: Oh wait, I see a mention of it in the `ascending` doc

alanag13: Why not just support them? Could we do this if we let a user provide the `sort_key` as normal and use that in the `sorted` call?

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,32 @@ def get_aggregate_data(self, alert_id):
...
+            query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
+            ascending (bool): Determines if sort ordering should be ascending/descending
+                based on alert creation time (overrides the sort order on the query object).
+                Defaults to True.
+
+        Returns:
+            generator: An object that iterates over alert detail items.
+        """
+        query.page_size = 100
+        query.sort_direction = "asc" if ascending else "desc"
```
alanag13, line 156: why not simply honor the `page_size` and `sort_direction` that the user passes in on the `query`?

src/py42/clients/alerts.py
alanag13: `createdAt` is still a good default for this though

timabrmsn: I suppose we could now that I think of it. As long as `sorted()` doesn't order things any differently than the server then it should just work I think...

src/py42/clients/alerts.py
timabrmsn: updated to add :func: for doc formatting

timabrmsn: and correct function name..

src/py42/clients/alerts.py
timabrmsn: `get_details()` only accepts a max of 100 ids, so we're just getting only what we can process in each iteration.

src/py42/clients/alerts.py
timabrmsn: Updated to now just use whatever is on the query for sort_key/direction. We need to tweak the sort key param because the server takes capitalized strings but returns camel-case.

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,30 @@ def get_aggregate_data(self, alert_id):
...
+        """
+        query.page_size = 100
+        sort_key = query.sort_key[0].lower() + query.sort_key[1:]
+        reverse = query.sort_direction == "desc"
+        pages = self._alert_service.search_all_pages(query)
+        for page in pages:
+            alert_ids = [alert["id"] for alert in page["alerts"]]
+            alert_details = self._alert_service.get_details(alert_ids)
+            for alert in sorted(
+                alert_details["alerts"], key=lambda x: x[sort_key], reverse=reverse,
```
alanag13, line 163: does this behave as expected when the field referenced by sort_key is a timestamp? I feel like python would just treat it as a string unless we explicitly convert.

alanag13: oh I see, we're using a sane format where those happen to end up with the same result, cool.

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,30 @@ def get_aggregate_data(self, alert_id):
...
+
+    def get_all_alert_details(self, query):
+        """
+        Helper method that combines :func:`.search_all_pages()` and :func:`.get_details()`.
+        Returns an iterator of alert detail objects in chronological order by alert
+        creation date.
+
+        Args:
+            query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
+            ascending (bool): Determines if sort ordering should be ascending/descending
```
unparalleled-js, line 148: `ascending` param doc no longer needed

src/py42/clients/alerts.py
timabrmsn: Yeah, I could go either way. I was thinking of this more as a corollary to `get_details()` but without the 100 alert limitation and also not having know the ids themselves. I also thought about `iter_all_alert_details` since we're iterating over the alerts individually opposed to a "page" that all the other methods return.

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,28 @@ def get_aggregate_data(self, alert_id):
             :class:`py42.response.Py42Response`
         """
         return self._alert_service.get_aggregate_data(alert_id)
+
+    def get_all_alert_details(self, query):
+        """
+        Helper method that combines :func:`.search_all_pages()` and :func:`.get_details()`.
```
unparalleled-js, line 142: The main reason to use this method is to get access to the alert observations, I suppose?
Should we mention that at all, that if you need the observations of all alerts, then use this?  Just an idea.

unparalleled-js:
Do we know if the alerts team ever plans on making their API returns observations / details from the 1 API?
It does seem like this change should exist on the backend rather than py42, since there are many use-cases.

tests/clients/test_alerts.py
```diff
@@ -92,3 +258,101 @@ def test_alerts_client_calls_get_aggregate_data_with_expected_value_and_param(
...
+            "D",
+            "E",
+            "F",
+        ]
+
+    @pytest.mark.parametrize(
+        "sort_key",
+        [
+            "AlertId",
+            "TenantId",
```
unparalleled-js, line 284: Seems to weird to be able to sort by tenant ID, wouldn't that just always be the same value?

src/py42/clients/alerts.py
```diff
@@ -136,3 +136,28 @@ def get_aggregate_data(self, alert_id):
...
+        Returns an iterator of alert detail objects.
+
+        Args:
+            query (:class:`py42.sdk.queries.alerts.alert_query.AlertQuery`): An alert query.
+
+        Returns:
+            generator: An object that iterates over alert detail items.
+        """
+        query.page_size = 100
```
unparalleled-js, line 151: Do we still want to document that `page_size` gets overriden?
What if a user is thinking they are managing their network usage by setting it

unparalleled-js (approved):

tests/clients/test_alerts.py
timabrmsn: yep, but it's a possible sort key so I figure we should at least make sure we handle it correctly if a user attempts it.

src/py42/clients/alerts.py
timabrmsn: yeah, that's a good call. I'll add a note.

src/py42/clients/alerts.py
timabrmsn: yeah, that's a good clarification of what "details" actually means.

alanag13 (approved):

src/py42/clients/alerts.py
peterbriggs42: Took a look at some of the other py42 clients and looks like we use `get_all` to mean "get all but we also support several filtering parameters". So I think in that context this usage is accurate and consistent.

peterbriggs42 (approved):

PR #352, opened 2021-08-03
Feature/integ 1737 xfc support

alanag13:
### Description of Change ###

Allows customers that do not use Code42 preservations feature to use the `sdk.securitydata.stream_file_by_md5()` and `sdk.securitydata.stream_file_by_sha256()` functions to stream files that were captured and via an agent's exfiltration detection.

This means that a given search could pull its file from either preservation data or exfiltration data. The logic to figure out which it should is as follows:
- if there is an exact match via exfiltration data, use this.
- if there is a path match via exfiltration data, use the latest version that matches the path
- otherwise, follow the same preservation logic as previously -- check preservation data for an exact match, if there isn't one, use the latest version.

Also refactors the usage of `py42_response` in the test suite. This makes setting up mock responses much easier and more consistently mimics their behavior.

### Issues Resolved ###

- closes INTEG-1737

### Testing Procedure ###
Call both `sdk.securitydata.stream_file_by_md5()` and `sdk.securitydata.stream_file_by_sha256()` in each of the following circumstances and ensure that a file is successfully downloaded:
- hash is available via preservation data 
- hash has an exact match
- hash has a matching path
- hash falls back to searching other devices that have this hash, and is available via preservation data of one of _those_ devices.

There is a tenant in our core-int environment that has suitable events for testing, slack me for details.

### PR Checklist ###


src/py42/clients/securitydata.py
```diff
@@ -279,18 +271,15 @@ def _get_device_file_version(self, device_guid, md5_hash, sha256_hash, path):
         response = self._preservation_data_service.get_file_version_list(
             device_guid, md5_hash, sha256_hash, path
         )
-        versions = response[u"versions"]
+        versions = response.data.get(u"securityEventVersionsMatchingChecksum") \
```
alanag13, line 274: This is the main logic change. The response no longer has `versions`, and instead returns an obj that has these three props. We want to use whichever is the first of these to have any data in them.

A customer cannot have both xfc and pds so `preservationVersions` will always be empty if either of the other two are not, or vice-versa.

src/py42/clients/securitydata.py
```diff
@@ -307,8 +296,37 @@ def _get_other_file_location_version(self, md5_hash, sha256_hash):
             version = self._preservation_data_service.find_file_version(
                 md5_hash, sha256_hash, paths
             )
-            if version.status_code != 204:
-                return version
+            if version.status_code == 200:
+                return version.data
+
+    def _get_file_stream(self, version):
+        if version.get(u"edsUrl"):
```
alanag13, line 303: if the version found was in a `securityEventVersions` prop, it will have an `edsUrl` property, otherwise the prop will be named `storageNodeURL`.

src/py42/services/preservationdata.py
```diff
@@ -19,8 +19,8 @@ def find_file_version(self, file_md5, file_sha256, paths):
...
         return self._connection.post(uri, json=data)
 
-    def get_file_version_list(self, deviceGuid, file_md5, file_sha256, path):
-        params = u"fileSHA256={}&fileMD5={}&deviceGuid={}&path={}"
-        params = params.format(file_sha256, file_md5, deviceGuid, quote(path))
-        uri = u"/api/v1/FileVersionListing?{}".format(params)
+    def get_file_version_list(self, deviceUid, file_md5, file_sha256, path):
+        params = u"fileSHA256={}&fileMD5={}&deviceUid={}&filePath={}"
+        params = params.format(file_sha256, file_md5, deviceUid, quote(path))
+        uri = u"/api/v2/file-version-listing?{}".format(params)
```
alanag13, line 25: no need to call the old version of this api anymore

tests/clients/test_securitydata.py
```diff
@@ -1321,6 +1325,146 @@ def test_stream_file_by_md5_when_find_file_version_returns_204_status_code_raise
 
         assert e.value.args[0] == PDS_EXCEPTION_MESSAGE.format("mdhash")
 
+    def test_stream_file_by_md5_when_has_exact_match_calls_get_token_with_expected_params_and_streams_successfully(
```
alanag13, line 1328: New tests start here, almost everything else in this PR is a result of refactoring `py42_response`

tests/integration/test_securitydata.py
```diff
@@ -60,9 +60,11 @@ def test_search_file_events(self, connection):
         assert_successful_response(response)
 
     def test_stream_file_by_md5(self, connection, md5_hash, file_data):
-        response = connection.securitydata.stream_file_by_md5(md5_hash)
-        assert str(response) == file_data
+        pass
```
alanag13, line 63: These are failing because the url for this method changed -- I'll need to open a PR against the mock server repo

alanag13: I'm going to wait to do this because there is other py42 work going on that is currently expecting the old one, so I'd rather have this be a clean break. I'll make a ticket to make sure the mock server gets updated and these get uncommented.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 ### Added
 
 - `RiskIndicator` and `RiskSeverity` filter classes to new `py42.sdk.queries.fileevents.filters.risk_filter` module.
+- support for Incydr SKUs that do not include backup and restore functionality to `sdk.securitydata.stream_file_by_sha256()` and  `sdk.securitydata.stream_file_by_md5()`
```
unparalleled-js, line 16: Missing newline between CL entries

unparalleled-js, line 16: Missing period at end of CL entry

unparalleled-js, line 16: CL entry missing capitalization

src/py42/clients/securitydata.py
```diff
@@ -251,20 +251,10 @@ def _stream_file(self, checksum, version_info):
...
-                version[u"versionTimestamp"],
-            )
-            return pds.get_file(str(token))
+            return self._get_file_stream(version)
 
         raise Py42Error(
-            u"No file with hash {} available for download on any storage node.".format(
-                checksum
-            )
+            u"No file with hash {} available for download.".format(checksum)
```
unparalleled-js, line 257: `u` not necessary anymore

alanag13: I will inevitably have to deal with that merge conflict, so I'll handle all those changes at that time.

src/py42/services/preservationdata.py
```diff
@@ -19,8 +19,8 @@ def find_file_version(self, file_md5, file_sha256, paths):
         uri = u"/api/v1/FindAvailableVersion"
         return self._connection.post(uri, json=data)
 
-    def get_file_version_list(self, deviceGuid, file_md5, file_sha256, path):
-        params = u"fileSHA256={}&fileMD5={}&deviceGuid={}&path={}"
-        params = params.format(file_sha256, file_md5, deviceGuid, quote(path))
-        uri = u"/api/v1/FileVersionListing?{}".format(params)
+    def get_file_version_list(self, deviceUid, file_md5, file_sha256, path):
+        params = u"fileSHA256={}&fileMD5={}&deviceUid={}&filePath={}"
```
unparalleled-js, line 23: Why not just use the `params=dict` in the call to `get()` rather than build your own query string?

src/py42/services/storage/exfiltrateddata.py
```diff
@@ -0,0 +1,49 @@
...
+class ExfiltratedDataService(BaseService):
+
+    _base_uri = u"api/v1/"
+
+    def __init__(self, main_session, streaming_session):
+        super(ExfiltratedDataService, self).__init__(main_session)
+        self._streaming_session = streaming_session
+
+    def get_download_token(self, event_id, device_id, file_path, timestamp):
+        """Get PDS download token for a file.
```
unparalleled-js, line 13: Should this be `Get EDS download token for a file.`?

alanag13: yep

src/py42/services/preservationdata.py
alanag13: I just copied this from what the old endpoint did, but yeah, that would be more consistent with what we have else where.

src/py42/clients/securitydata.py
```diff
@@ -307,8 +296,35 @@ def _get_other_file_location_version(self, md5_hash, sha256_hash):
...
+            if version.status_code == 200:
+                return version.data
+
+    def _get_file_stream(self, version):
+        if version.get(u"edsUrl"):
+            return self._get_exfiltrated_file(version)
+
+        return self._get_stored_file(version)
+
+    def _get_stored_file(self, version):
```
unparalleled-js, line 308: Could swap the placement of `_get_exfiltrated_file()` and `_get_stored_file()` to match the order of usage in the method above.

src/py42/clients/securitydata.py
```diff
@@ -470,6 +486,12 @@ def node_guid(self):
         return self._node_guid
 
 
+def _get_first_matching_version(versions, md5_hash):
+    exact_match = next((x for x in versions if x[u"fileMD5"] == md5_hash), None,)
```
unparalleled-js, line 490: there are certain cases where a trailing comma in the tuple like this can cause our formatter to throw odd errors (specifically, `flake8`). It isn't now (obviously), but the comma is not needed and I am scarred from my past.

src/py42/services/preservationdata.py
```diff
@@ -19,8 +19,8 @@ def find_file_version(self, file_md5, file_sha256, paths):
         uri = u"/api/v1/FindAvailableVersion"
         return self._connection.post(uri, json=data)
 
-    def get_file_version_list(self, deviceGuid, file_md5, file_sha256, path):
-        params = u"fileSHA256={}&fileMD5={}&deviceGuid={}&path={}"
-        params = params.format(file_sha256, file_md5, deviceGuid, quote(path))
-        uri = u"/api/v1/FileVersionListing?{}".format(params)
+    def get_file_version_list(self, deviceUid, file_md5, file_sha256, path):
```
unparalleled-js, line 22: `deviceUid` should be in `snake_case`?

unparalleled-js: I am guessing it is not a breaking change? Or is it? Because it also is being renamed already from `deviceGuid` to `deviceUid`, so we should fix the casing or let it be?

src/py42/services/storage/_service_factory.py
```diff
@@ -36,6 +37,11 @@ def create_preservation_data_service(self, host_address):
         streaming_connection = Connection.from_host_address(host_address)
         return StoragePreservationDataService(main_connection, streaming_connection)
 
+    def create_exfiltrated_data_service(self, host_address):
+        main_connection = self._connection.clone(host_address)
+        streaming_connection = Connection.from_host_address(host_address)
```
unparalleled-js, line 42: I see we did this for pds as well, but what is the benefit of having separate sessions for streaming?

src/py42/services/storage/exfiltrateddata.py
```diff
@@ -0,0 +1,49 @@
...
+        }
+        resource = u"file-download-token"
+        uri = u"{}{}".format(self._base_uri, resource)
+        return self._connection.get(uri, params=params)
+
+    def get_file(self, token):
+        """Streams a file.
+
+        Args:
+            token (str): PDS Download token.
```
unparalleled-js, line 38: `EDS Download token.`

src/py42/services/preservationdata.py
alanag13: I don't expect that anyone is realistically calling this function directly and even if they are its even less likely they're doing it with a named param. We should use the proper casing, thanks.

tests/conftest.py
```diff
@@ -115,12 +115,13 @@ def unauthorized_response(mocker, http_error):
     return response
 
 
-@pytest.fixture
-def py42_response(mocker):
+def py42_response(mocker, text):
```
unparalleled-js, line 118: I don't think we should be mocking the `Py42Response`.
I think we should just use the already existing `create_mock_response()` method, which mocks the `Response` and passes it into the `Py42Response` class. It is works nicer in my experience.

src/py42/services/storage/_service_factory.py
alanag13: Streaming only works if you use anonymous auth.

tests/conftest.py
alanag13: 🤦🏼  here I was thinking that I was creating a better way of doing this when one already existed? You're totally right, fortunately the method signature are the same so this should be pretty easy to update en masse

src/py42/services/preservationdata.py
unparalleled-js: Nevermind - I remember there is some annoying reason for this

tests/conftest.py
unparalleled-js: Yeah I feel bad as you changed so many! they needed it

tests/clients/test_securitydata.py
```diff
@@ -219,6 +218,76 @@
...
+    ],
+    "securityEventVersionsMatchingChecksum": [],
+    "securityEventVersionsAtPath": []
+}"""
+
+XFC_EXACT_FILE_VERSIONS = """{
+    "preservationVersions": [],
+    "securityEventVersionsMatchingChecksum": [
+        {
+            "edsUrl": "https://host-1.com",
```
unparalleled-js, line 230: should we do `"https://host-1.example.com"` so this never resolves on accident

alanag13: I do tend to recommend using `example.com` due to the rule that it can never be anything, but don't feel it's worth searching the test suite for stuff that isn't that just to make that switch. That being said, since I was touching this code anyway,  it's probably a good thing to do.

src/py42/services/preservationdata.py
alanag13: yeah i remembered too, its the encoding on `filePath`.

tests/clients/test_securitydata.py
```diff
@@ -1321,6 +1336,166 @@ def test_stream_file_by_md5_when_find_file_version_returns_204_status_code_raise
...
+        security_service,
+        file_event_service,
+        preservation_data_service,
+        saved_search_service,
+        storage_service_factory,
+        file_event_search,
+        file_download,
+        mocker,
+    ):
+        file_version_list = py42_response(mocker, XFC_EXACT_FILE_VERSIONS)
```
unparalleled-js, line 1350: A fixture (or fixtures) and/or method for the setup of this test and the next would be nice. It'd help us manage testing states in a global way, and a familiar way if uses fixtures.

tests/clients/test_securitydata.py
```diff
@@ -1321,6 +1336,166 @@ def test_stream_file_by_md5_when_find_file_version_returns_204_status_code_raise
...
+        security_client = SecurityDataClient(
+            security_service,
+            file_event_service,
+            preservation_data_service,
+            saved_search_service,
+            storage_service_factory,
+        )
+        response = security_client.stream_file_by_md5("testmd5-2")
+        assert response == b"stream"
+        download_token_params = [
```
unparalleled-js, line 1449: The name of this param makes it sound more like an `actual` than an `expected`

tests/services/detectionlists/test_departing_employee.py
```diff
@@ -65,29 +66,27 @@ def test_choices_are_correct(self):
 
 class TestDepartingEmployeeClient(object):
     @pytest.fixture
-    def mock_get_all_response(self, py42_response):
-        py42_response.text = _GET_ALL_RESPONSE
-        return py42_response
+    def mock_get_all_response(self, mocker):
+        return py42_response(mocker, _GET_ALL_RESPONSE)
```
unparalleled-js, line 70: This is not really on you, but this `return` should probably be a `yield`, if we are being picky about types, which _can_ matter sometimes.

tests/clients/test_securitydata.py
unparalleled-js: In other spots, I ust meant `host-2` etc, in the same blob

test.py
```diff
@@ -0,0 +1,20 @@
+import py42.sdk
+
+sdk = py42.sdk.from_local_account("https://console.core-int.cloud.code42.com/", "alan.grgic+coreint_dnp@code42.com", "Nh5)2K3oaIjyKzF")
```
unparalleled-js, line 3: oops

src/py42/clients/securitydata.py
ryan-haley-code42: Ran into same issue of the two fighting and failing. Had to manually remove the trailing comma after the 3.6+ args was added.

alanag13:
@unparalleled-js okay I've addressed everything and merged in the latest changes! Making the fixture took a bit but was totally worth it, those tests are definitely easier to work with now

unparalleled-js (approved):

PR #353, opened 2021-08-05
Bugfix/fix 2fa

timabrmsn:
### Description of Change ###

- Deprecates Py42MFARequiredError
- Adds `sdk.loginconfig.get_for_user(username)` method. 
- Catches `Py42Unauthorized` exception in `py42.sdk.from_local_account` and provides more helpful exception messages.

### Testing Procedure ###

SSO User test:
- Attempt to initialize the SDKClient using `py42.sdk.from_local_account()` with a user configured for SSO auth.
- A `Py42Error` should be raised with text: ```SSO users are not supported in `from_local_account()`.```

Local & 2FA user test:
- Attempt to initialize the SDKClient using `py42.sdk.from_local_account()` with a local user (both with 2FA enabled and without), but use incorrect credentials and/or TOTP token.
- A `Py42UnauthorizedError` should be raised with text: ```SDK initialization failed, double-check username/password, and provide two-factor TOTP token if Multi-Factor Auth configured for your user. User LoginConfig: {login_type}``` where `login_type` matches that of the user you used (either `LOCAL` or `LOCAL_2FA`). 

Get Login Configuration method test:
- Initialize the SDKClient from the `SDKClient.from_local_account()` classmethod (this doesn't check credentials on initialization. Use incorrect credentials as `sdk.loginconfig.get_for_user()` should work unauthenticated:
```
from py42.sdk import SDKClient
sdk = SDKclient.from_local_account("{console_url}", "{username}", "{bad_password}")
```

- Call `sdk.loginconfig.get_for_user("{username"})`
- the `loginType` key from the response data should match that of the user login type requested (`LOCAL`, `LOCAL_2FA`, or `CLOUD_SSO`).

### PR Checklist ###


src/py42/sdk/__init__.py
```diff
@@ -133,6 +146,19 @@ def from_jwt_provider(cls, host_address, jwt_provider):
 
         return cls(main_connection, custom_auth)
 
+    def get_login_configuration_for_user(self, username):
```
unparalleled-js, line 149: `for_user` might be extra,
`get_login_configuration()` probably suffices

src/py42/services/_connection.py
```diff
@@ -127,30 +127,35 @@ def from_device_connection(cls, connection, device_guid):
     def host_address(self):
         return self._get_host_address()
 
+    def get_login_configuration_for_user(self, username):
```
unparalleled-js, line 130: Are we intentionally making this available to every type of Connection?
I don't think it would work on a microservice connection.

Would it be better to have a `LoginConfiguration` client that takes a simple `Connection` as a parameter.
Then, from the `sdk`, you would just do `sdk.login_configuration.get_for_user()` or however.

unparalleled-js: We are also missing tests for this part (just the simple `_calls_get_with_expected_url_and_params()`, but if we separate this from `Connection`, those tests would be more isolated in their spot as well, which would be nice.

timabrmsn: Hmm, yeah that makes more sense.

timabrmsn: refactored it into a LoginConfigurationClient.

CHANGELOG.md
```diff
@@ -16,10 +16,14 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 
 - New method `sdk.alerts.get_all_alert_details()` as a helper to make getting alerts with details easier (combines `sdk.alerts.search_all_pages()` and `sdk.alerts.get_details()`).
 
+- New method `sdk.loginconfig.get_for_user()` to identify if a user's login type is `LOCAL`, `LOCAL_2FA`, or `CLOUD_SSO`.
+
 ### Removed
 
 - py42 no longer supports python 2.7 or python 3.5.
 
+- Usage of `Py42MFARequiredError`. Use `sdk.get_login_configuration_for_user()` instead to check if a user is configured for two-factor authentication.
```
unparalleled-js, line 25: This line does not reference the updated method path

src/py42/clients/loginconfig.py
```diff
@@ -0,0 +1,21 @@
...
+
+class LoginConfigurationClient:
+    def __init__(self, connection):
+        self._connection = connection
+
+    def get_for_user(self, username):
+        """Retrieves login configuration for a given username. Possible `loginType` values are
+        `LOCAL`, `LOCAL_2FA`, and `CLOUD_SSO`. If username does not exist the default
+        return value is `LOCAL_2FA`.
+
```
unparalleled-js, line 12: Could include rest link:

```
`REST Documentation: <https://console.us.code42.com/swagger/index.html?urls.primaryName=v3#/Feature/get>`__
```

CHANGELOG.md
timabrmsn: done

src/py42/clients/loginconfig.py
timabrmsn: done

unparalleled-js (approved):

PR #355, opened 2021-08-09
Chore/abstract choices method

tora-kozic:
### Description of Change ###

Created the `Choices` class to contain a `choices()` method whose definition was previously duplicated in many areas of the code.  Removed this definition from those classes and then had them inherit this new class.

### Testing Procedure ###
Because no functionality was changed, only the code structure, running existing unit tests for the choices() methods should test these changes.

### PR Checklist ###


src/py42/sdk/queries/__init__.py
```diff
@@ -26,3 +27,16 @@ def any(cls, *args):
     @classmethod
     def all(cls, *args):
         return cls(*args)
+
+
+class Choices:
```
unparalleled-js, line 32: Since this gets used outside of the search queries, it should probably live in a different file, perhaps `util.py` or its own file `choices.py` at the root of the repo

src/py42/sdk/queries/fileevents/filters/risk_filter.py
```diff
@@ -77,7 +77,7 @@ def choices():
             RiskIndicator.CloudDataExposures.choices()
             + RiskIndicator.CloudStorageUploads.choices()
             + RiskIndicator.CodeRepositoryUploads.choices()
-            + RiskIndicator.EmaiServiceUploads.choices()
+            + RiskIndicator.EmailServiceUploads.choices()
```
unparalleled-js, line 80: We need to make sure to get this in the next release!  These `RiskIndicator` enums thankfully have not been released yet.

src/py42/sdk/queries/__init__.py
```diff
@@ -26,3 +27,16 @@ def any(cls, *args):
...
+    """Helper class to provide the choices() method to a class."""
```
unparalleled-js, line 33: If you put back ticks, e.g. ``choices()`` around the method name, it will look nicer in the generated-documentation page.

src/py42/sdk/queries/__init__.py
unparalleled-js: Probably it is own file, the more I think about this. Then, we don't have to include it in the generated-docs and it can be just an implementation detail that we can freely change.

src/py42/sdk/queries/fileevents/filters/risk_filter.py
alanag13: This was my doing, I'll open a separate fix for it 😞

alanag13:
Let's hold off on merging this until after the next release.

src/py42/sdk/queries/fileevents/filters/risk_filter.py
unparalleled-js: Thanks @alanag13  No worries :)

src/py42/choices.py
```diff
@@ -0,0 +1,14 @@
+from py42.util import get_attribute_keys_from_class
+
+
+class Choices:
+    """Helper class to provide the choices() method to a class."""
+
+    @classmethod
+    def choices(cls):
+        """Returns attribute names for the given class.
```
timabrmsn, line 9: Hmm, one thing I noticed (this has always been the case, it's not due to this change), is that this description is incorrect. `.choices()` returns attribute _values_ not names (or "keys", based on the naming of `get_attribute_keys_from_class()` func.

Example, this is what .choices() currently returns:
```
>>> EventTimestamp.choices()
['PT15M', 'P14D', 'P1D', 'PT1H', 'P7D', 'P30D', 'P3D', 'PT3H', 'PT12H']
```
if it were actually grabbing attribute names it would be:

```
>>> EventTimestamp.choices()
['FIFTEEN_MINUTES', 'FOURTEEN_DAYS', 'ONE_DAY', 'ONE_HOUR', 'SEVEN_DAYS', 'THIRTY_DAYS', 'THREE_DAYS', 'THREE_HOURS', 'TWELVE_HOURS']
```

@alanag13 @unparalleled-js what was the original intent of the choices() usage? Names or values?

alanag13: Good catch -- values, the intent is to enumerate all options for a certain value that could be sent to the API. The name of the python var doesn't matter. We should update the docs accordingly.

alanag13: @tora-kozic if we change this docstr to attribute _values_ this all LGTM.

tora-kozic: I can do that! Would it be also appropriate to rename `get_attribute_keys_from_class()` to `get_attribute_values_from_class()` or do we want to leave that as is?

alanag13 (approved):

PR #357, opened 2021-08-10
use v2 of alert aggregate data api

unparalleled-js:
### Description of Change ###

Uses the new v2 API for the alert aggregate data.
The response seems mostly the same EXCEPT:

1.) v1 of api has field named `ffsUrl` but on v2 api it is named `ffsUrlEndpoint`.... (breaking change if updating the method, that is why it aliases the one field)
2.) v2 of api has more fields.

^ number 1 is unfortunate... it is why I had to make an alias and now the response has the same property twice

Note: I need v2 for splunk, but I don't need this merged or released, I can just drop it in for now, but just need to remember to legitimately update py42 in splunk later

What should we do???

### Testing Procedure ###

Example usage:  https://github.com/unparalleled-js/code42cli-extensions/blob/main/jules42.py#L220-L227

### PR Checklist ###


timabrmsn:
I went looking at the alertservice's logs to see if v1 has been used at all recently, and there hasn't been any usage at all in US2 (for at least the past 6 months), but it seems there's something (probably automated) calling it in US1, as it's been called at least once a minute for the past month. Unfortunately the logging doesn't provide tenant or user agent data so I don't know if it's py42 usage or an internal Code42 tenant the calls are being made against. I pinged the alerts team about helping to figure out who it was and they're going to try to get updated logging for tenant and user agent pushed out today.

alanag13:
instead of a v2 method, we could consider keeping the old one and just making it so that `ffsUrl` can be accessed under both the old and new names (i.e. modifying the response to add the new prop name to it, and giving the prop the same value as the old one).

src/py42/services/alerts.py
```diff
@@ -134,11 +134,11 @@ def update_note(self, alert_id, note):
         return self._connection.post(uri, json=data)
 
     def get_aggregate_data(self, alert_id):
-        uri = self._uri_prefix.format("query-details-aggregate")
-        data = {
-            "alertId": alert_id,
-        }
-        return self._connection.post(uri, json=data)
+        uri = f"{self._uri_prefix}/v2/query-details-aggregate"
+        data = {"alertId": alert_id}
```
unparalleled-js, line 138: One other thing I noticed about this method:  we don't convert the observations to JSON like we do for the `get_details()` method.

alanag13: sad, but it might be too late to do anything different now.

alanag13 (approved):

timabrmsn:
The new logging is live, and so far today the only times the v1 endpoint has been called (6x from 2 different tenants) it's been from the workato connector. I don't know if we want to give it some more time for more data, or if it's relatively safe to say no one is using this method in py42 actively.

unparalleled-js:
> The new logging is live, and so far today the only times the v1 endpoint has been called (6x from 2 different tenants) it's been from the workato connector. I don't know if we want to give it some more time for more data, or if it's relatively safe to say no one is using this method in py42 actively.

Are you suggesting we do not do the aliasing since no one is using the old endpoint?

src/py42/services/alerts.py
unparalleled-js: If we would have went with the `v2` method approach, I would have done it then.

However, Tim is suggesting that maybe there are 0 actual users of the original method, if that changes anything

unparalleled-js:
> > Are you suggesting we do not do the aliasing since no one is using the old endpoint?
> 
> It's an option I think.



> > Are you suggesting we do not do the aliasing since no one is using the old endpoint?
> 
> It's an option I think.

I think it is better practice not to do that, because you never know who's idol script is laying around or something.

timabrmsn:
> I think it is better practice not to do that, because you never know who's idle script is laying around or something.

Sure, I'm ok with being cautious about it.

timabrmsn (approved):

PR #359, opened 2021-08-12
Feature/saved search all events

tora-kozic:
### Description of Change ###

Addition of the method `sdk.securitydata.savedsearches.execute_get_all(search_id, page_token)` which makes use of page tokens when a search query returns over 10,000 results. The method takes the required parameter `search_id` and the optional parameter `page_token` and returns a page of events matching the saved search query.  If there are greater than 10,000 resulting events the object will include a token for the next page.  If the `page token` is not given (or entered as an empty string) the first page of events will be returned. 

### Testing Procedure ###
Running the function with a query should return a response containing the events which match the saved search query, if there are over 10,000 results the response should include a token for the next page.  If provided a token on input the function should properly handle the token with or without escaped quotes, and then return the matching event page to that token (and the corresponding next page token, if applicable).

### PR Checklist ###


src/py42/services/savedsearch.py
```diff
@@ -61,3 +62,20 @@ def execute(self, search_id, page_number=None, page_size=None):
         """
         query = self.get_query(search_id, page_number=page_number, page_size=page_size)
         return self._file_event_client.search(query)
+
+    def execute_get_all(self, search_id, page_token=""):
```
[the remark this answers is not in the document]
ryan-haley-code42: I feel `execute_pagination` might be a bit better, `get_all` feels like a call once and done method or returning a stream that I interact with over having to call again with the next token.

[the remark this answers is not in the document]
ryan-haley-code42: Are we avoiding changing `execute()` to use pagination to not break anything for existing usage? If so then I think keeping the name `execute_get_all()` might make more sense to really drive home the difference between the methods.

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- New method `sdk.securitydata.savedsearches.execute_get_all()` which makes use of page tokens when a search query returns over 10,000 results.
+
+### Removed
```
unparalleled-js, line 17: This `Removed` is not needed yet

src/py42/services/savedsearch.py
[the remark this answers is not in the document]
unparalleled-js: @ryan-haley-code42 I don't think it'd be a breaking change. However, in the `securitydata` client, there are two methods `search_file_events()` and `search_all_file_events()` that have this same distinction (one support `page_token` and one does not.

Also, I should add that both technically support pagination. There are like two levels of pagination when it comes to these file event search APIs... It is just that they originally caps the number of events to 10,000 to prevent API abuse. However, we had to write scripts to get around this limitation because customers needed all the events... Thus, they added the ability to search past 10,000 via this page token.

I am not sure exactly the motive behind why there are two methods `search_file_events()/search_all_file_events()`, but I remember the original plan being that the `search_all()` one would return a generator.  It appears that is not the case. However, methods in py42 that return generators _should_ include the keyword `all` somewhere in the method name.  I suppose these file event search methods are different, but we seem to have done the same distinction via the `all` keyword.

I think it is important, given that savedsearches are completely paired with the searching file event API, that it mirrors the same method distinction of the one having `page_token` having the word `all` somewhere in the method name.

That being said,  this is a difficult method to name and I don't really have a better idea, I think `execute_get_all()` would work but am curious what others think.

Sorry, that is a lot of info

src/py42/services/util.py
```diff
@@ -13,3 +15,18 @@ def get_all_pages(func, key, *args, **kwargs):
         yield response
         page_items = response[key]
         item_count = len(page_items)
+
+
+def escape_quote_chars(token):
```
unparalleled-js, line 20: In general, this might make more sense at the root `util.py`.
However, I like that it is here because this way, it won't show up in the docs! and it can be treated as an implementation detail and changes at will for v1 of py42.

tests/services/test_savedsearch.py
```diff
@@ -1,15 +1,22 @@
...
 
 SAVED_SEARCH_GET_RESPONSE = """
     {"searches": [{"groups": [] }]}
 """
+FILE_EVENT_URI = "/forensic-search/queryservice/api/v1/fileevent"
 
 
 class TestSavedSearchService:
+    @pytest.fixture
+    def connection(self, mocker):
```
unparalleled-js, line 17: I don't think this is needed.

The `mock_connection` fixture is already being imported here, and it is basically the same thing.

src/py42/services/savedsearch.py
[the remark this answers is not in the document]
unparalleled-js: If we go this route, someone should remember to add a ticket in the py42 2.0 GitHub project that says to remove the `execute` methods

src/py42/services/savedsearch.py
```diff
@@ -61,3 +62,20 @@ def execute(self, search_id, page_number=None, page_size=None):
...
+        Args:
+            search_id (str): Unique search Id of the saved search.
+            page_token (str, optional): A token used to indicate the starting point for
+                additional page results. For the first page, do not pass ``page_token``. For
+                all consecutive pages, pass the token from the previous response from
+                field ``nextPgToken``. Defaults to empty string.
+        Returns:
+            :class:`py42.response.Py42Response`: A response containing a page of events.
+        """
+        query = self.get_query(search_id)
```
alanag13, line 79: on thought I just had -- it is unfortunate that that if you are paging through this that we will have to make the api call to get the query every time. I wonder if there's a good way could design that that would allow us to avoid that overhead.

tora-kozic: I imagine this is why in fileevents querying you pass the `Query` instead of the `search_id`?  If we wanted this one to work similarly, we could switch it to take that parameter instead - however then it would diverge from the original 'execute()' method. 

Or perhaps if we wanted to keep the param as `search_id` we could have the function make this api call and then call some helper function on that returned Query?  I'm not sure how closely we want this new method to mirror `execute()`.

timabrmsn: If we didn't make the user handle the page tokens (but instead handled that internally and just yielded pages like other `get_all` methods), we could avoid the overhead, but since "paging" requires multiple manual calls of this same method I don't know if we'd be able to safely cache the query results, since we don't know when the user is "done" in order to clear the cache. 

Any reason not to do that here? I know we're not consistent with the regular `securitydata.search_all_file_events` in how we handle page tokens, but the end user doesn't really need to know that saved search is doing the same thing automatically for them, right?

tora-kozic: So sounds like we just want to leave it as is to stay consistent with the other `search_all_file_events()` method for now, and then we'll plan on improving their design in the future

alanag13: I kind of realized that we can achieve what we want already by doing the following:
- call `get_query`
- pass the result of that to `securitydata.search_all_file_events`, use that method as normal

Adding a `search_all_file_events` method here would encourage users to use something that results in making extra, unneeded calls -- maybe we can just update the docs of the `search_file_events` and/or `get_query` method here to call out how to go about doing this, and get rid of `search_all_file_events`?

CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 The intended audience of this file is for py42 consumers -- as such, changes that don't affect
 how a consumer would use the library (e.g. adding unit tests, updating documentation, etc) are not captured here.
 
+## Unreleased
+
+### Added
+
+- New method `sdk.securitydata.savedsearches.execute_get_all()` which makes use of page tokens when a search query returns over 10,000 results.
```
alanag13, line 15: This method name has changed since this log entry was created (and maybe can just be removed: https://github.com/code42/py42/pull/359/files#r689673899)

We should also call out the new alias that we added for `execute()`.

src/py42/services/savedsearch.py
timabrmsn: Hmm, I like that idea..  make the "main" way to execute a saved search be to get the query object, then use it in the normal file event search methods. 

Maybe we could mark `.execute` as deprecated and update documentation to highlight the better way?

tora-kozic: So there would be `securitydata.search_all_file_events()` which takes a query and an optional page token and returns a response _with_ page tokens (if applicable).
There'd be securitydata.search_file_events() which takes a query and returns a response _without_ page tokens.

We'd remove `savedsearch.search_all_file_events()` entirely.
And then we'd have savedsearch.search_file_events() which takes a search id and returns a response without page tokens (currently be calling `execute()` which also calls `get_query()`.  Should we remove this method as well and plan on documenting that users simply use `get_query()` and then pass that into the existing `securitydata` file event search methods?

alanag13: >  Should we remove this method as well and plan on documenting that users simply use get_query() and then pass that into the existing securitydata file event search methods?

No, `execute` already exists and this is just a better name for it -- and removing it would be breaking.
There's still value to getting the first 10K results using just a search id, imo.

alanag13 (approved):

PR #366, opened 2021-09-13
enable tests that use exfiltration data service

alanag13:
### Description of Change ###

Re-enables some tests that were temporarily disabled after the introduction of functionality from the exfiltration data service.

https://github.com/code42/code42-mock-servers/pull/15 is a dependency of this PR so it needs to be merged first before the newly-enabled tests will pass.

### DIFF
--- a/src/py42/services/storage/exfiltrateddata.py
+++ b/src/py42/services/storage/exfiltrateddata.py
@@ -26,8 +26,9 @@ class ExfiltratedDataService(BaseService):
         params = "deviceUid={}&eventId={}&filePath={}&versionTimestamp={}"
         params = params.format(device_id, event_id, quote(file_path), timestamp)
         resource = "file-download-token"
+        headers = {"Accept": "*/*"}
         uri = f"{self._base_uri}{resource}?{params}"
-        return self._connection.get(uri)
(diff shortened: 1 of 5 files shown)


alanag13:
@timabrmsn @tora-kozic @ryan-haley-code42 this is ready now that the dependent PR has merged and i fixed the other minor snags in the checks.

src/py42/services/storage/exfiltrateddata.py
```diff
@@ -26,8 +26,9 @@ def get_download_token(self, event_id, device_id, file_path, timestamp):
         params = "deviceUid={}&eventId={}&filePath={}&versionTimestamp={}"
         params = params.format(device_id, event_id, quote(file_path), timestamp)
         resource = "file-download-token"
+        headers = {"Accept": "*/*"}
```
timabrmsn, line 29: was this header not being present breaking anything before?

alanag13: not in the production sense, but the way the mock server works is that the response type _must_ be compatible with the request's `Accept` headers, so this is needed for the integration tests to work.

timabrmsn (approved):

PR #367, opened 2021-09-13
Feature/add trusted domain methods

tora-kozic:
### Description of Change ###

- New `trustedactivities` client with the following new methods:
    - `sdk.trustedactivities.get_all()` to get all trusted activities.
    - `sdk.trustedactivities.create()` to create a new trusted activity.
    - `sdk.trustedactivities.get()` to get details about a single trusted activity.
    - `sdk.trustedactivities.update()` to update a trusted activity.
    - `sdk.trustedactivities.delete()` to delete a trusted activity

- New custom exceptions
    - `Py42TrustedActivityInvalidChangeError`
    - `Py42TrustedActivityConflictError`
    - `Py42TrustedActivityInvalidCharacterError`

### Testing Procedure ###
Users should be able to use the py42 sdk methods of the `trustedactivities` client to hit the corresponding API endpoints to add, update, delete, and retrieve trusted activity details.  Currently the two types of activity supported are `DOMAIN` (urls) and `SLACK` (slack workspaces).  

Unit tests for the client and services can be found in `tests/clients/test_trustedactivities.py` and `tests/clients/test_trustedactivities.py`, respectively.  A separate story specifies adding integration tests and adding the service to the mock-servers.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/trustedactivities.py
+++ b/src/py42/clients/trustedactivities.py
@@ -15,7 +15,7 @@ class TrustedActivitiesClient:
     def __init__(self, trusted_activities_service):
         self._trusted_activities_service = trusted_activities_service
 
-    def get_all(self, type=None):
+    def get_all(self, type=None, page_size=None):
         """Gets all trusted activities.
         `Rest documentation <https://developer.code42.com/>`
 
@@ -25,7 +25,7 @@ class TrustedActivitiesClient:
         Returns:
             :class:'py42.response.Py42Response'
         """
-        return self._trusted_activities_service.get_all(type)
+        return self._trusted_activities_service.get_all(type, page_size)
 
     def create(self, type, value, description=None):
         """Gets all trusted activities with the given type.
--- a/src/py42/clients/trustedactivities.py
+++ b/src/py42/clients/trustedactivities.py
@@ -1,0 +1,69 @@
+from py42.choices import Choices
+
+
+class TrustedActivityType(Choices):
+    DOMAIN = "DOMAIN"
+    SLACK = "SLACK"
+
+
+class TrustedActivitiesClient:
+    """A client to expose the trusted activities/data preferences API
+
+    `Rest documentation <https://developer.code42.com/api/#tag/Trusted-Activities>`__
+    """
+
+    def __init__(self, trusted_activities_service):
+        self._trusted_activities_service = trusted_activities_service
+
+    def get_all(self, type):
+        """Gets all trusted activities with the given type.
+        `Rest documentation <https://developer.code42.com/api/TODO>`
+
+        Args:
+            type (str): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.get_all(type)
+
+    def get(self, id):
+        """Retrieve trusted activity details by given resource number.
+        `Rest documentation <https://developer.code42.com/api/TODO>`
+
+        Args:
+            id (int): Resource number of the trusted activity or domain.
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.get(id)
+
+    def update(self, id, type, value, description):
+        """Updates trusted activity details by given resource number.
+        `Rest documentation <https://developer.code42.com/api/TODO>`
+
+        Args:
+            id (int): Resource number of the trusted activity.
+            type (str): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
+            value (str): The URL of the domain or name of the Slack workspace.
+            description (str): Description of the trusted activity.
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.update(
+            id=id, type=type, value=value, description=description
+        )
+
+    def delete(self, id):
+        """Deletes a trusted activity by given resource number.
+        `Rest documentation <https://developer.code42.com/api/TODO>`
+
+        Args:
+            id (int): Resource number of the trusted activity or domain.
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.delete(id)
--- a/src/py42/clients/trustedactivities.py
+++ b/src/py42/clients/trustedactivities.py
@@ -53,13 +53,12 @@ class TrustedActivitiesClient:
         """
         return self._trusted_activities_service.get(id)
 
-    def update(self, id, type=None, value=None, description=None):
+    def update(self, id, value=None, description=None):
         """Updates trusted activity details by given resource number.
         `Rest documentation <https://developer.code42.com/>`
 
         Args:
             id (int): Resource number of the trusted activity.
-            type (str, optional): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
             value (str, optional): The URL of the domain or name of the Slack workspace.
             description (str, optional): Description of the trusted activity.
 
@@ -67,7 +66,7 @@ class TrustedActivitiesClient:
             :class:'py42.response.Py42Response'
         """
         return self._trusted_activities_service.update(
-            id=id, type=type, value=value, description=description
+            id=id, value=value, description=description
         )
 
     def delete(self, id):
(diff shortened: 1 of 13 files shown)


src/py42/clients/trustedactivities.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Args:
+            id (int): Resource number of the trusted activity or domain.
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.get(id)
+
+    def update(self, id, type=None, value=None, description=None):
```
tora-kozic, line 56: Users can't change the `type` of a trusted activity (currently throws a custom `invalid change` exception), so should this not be a parameter of the `update()` method in the client, or altogether (remove it from `update() in both the client and service)?

alanag13: yeah, I would remove it, no reason to provide functionality that will either do nothing or always fail

src/py42/exceptions.py
```diff
@@ -337,6 +337,33 @@ def __init__(self, exception):
...
+        super().__init__(exception, msg)
+
+
+class Py42TrustedActivityConflictError(Py42HTTPError):
+    """An error raised when theres a conflict with a trusted activity domain URL."""
+
+    def __init__(self, exception, value):
+        msg = (
+            f"Duplicate URL or workspace name, '{value}' already exists on your trusted list.  "
+            "Please enter a unique value"
```
alanag13, line 354: let's use `provide` instead of `enter`.

src/py42/clients/trustedactivities.py
```diff
@@ -0,0 +1,83 @@
...
+        Args:
+            id (int): Resource number of the trusted activity.
+            type (str, optional): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
+            value (str, optional): The URL of the domain or name of the Slack workspace.
+            description (str, optional): Description of the trusted activity.
+
+        Returns:
+            :class:'py42.response.Py42Response'
+        """
+        return self._trusted_activities_service.update(
```
alanag13, line 69: A side effect of the way the update API works is that it will replace all properties with whatever was sent -- i.e. if you only want to update the value and don't send a description, then the description will be updated to be blank.

To avoid this oddity, we can first get the activity that we want to update and build the update request based off of that. Here's an example of how we manage that problem in the cases API:

https://github.com/code42/py42/blob/0d0a472bfce6d1ccd5ae09eb5789e8494c0da247/src/py42/services/cases.py#L110

src/py42/clients/trustedactivities.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        return self._trusted_activities_service.get(id)
+
+    def update(self, id, type=None, value=None, description=None):
+        """Updates trusted activity details by given resource number.
+        `Rest documentation <https://developer.code42.com/>`
+
+        Args:
+            id (int): Resource number of the trusted activity.
+            type (str, optional): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
```
alanag13, line 62: does IP address not have a separate type?

src/py42/exceptions.py
```diff
@@ -337,6 +337,33 @@ def __init__(self, exception):
...
+        )
+        super().__init__(exception, msg)
+
+
+class Py42TrustedActivityInvalidCharacterError(Py42BadRequestError):
+    """An error raised when an invalid change is being made to a trusted activity."""
+
+    def __init__(self, exception):
+        msg = "Domain name cannot include @"
```
alanag13, line 363: Let's make this more generic in case the service decides that there are other invalid characters besides just `@`

src/py42/clients/trustedactivities.py
tora-kozic: the cases API does this in its services method (rather than the client) - I did this same thing for the trustedactivities `update()` method its just in `services`

src/py42/clients/trustedactivities.py
tora-kozic: I don't believe so.. unless its hiding somewhere? In the console you're only prompted with DOMAIN or SLACK.  And there doesn't seem to be any kind of check that "domain" is actually a URL

src/py42/exceptions.py
tora-kozic: Changed this to `"Invalid character in domain or slack workspace name"`

src/py42/clients/trustedactivities.py
alanag13: oh i missed that thanks, nice!

src/py42/clients/trustedactivities.py
alanag13: there's a separate tab at the top of the UI to enter IP addresses.

src/py42/services/trustedactivities.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class TrustedActivitiesService(BaseService):
+
+    _uri_prefix = "/api/v1/trusted-activities"
+
+    def __init__(self, connection):
+        super().__init__(connection)
+
+    def get_all(self, type=None):
```
alanag13, line 16: does this API not support paging? i.e. this endpoint always returns everything?

tora-kozic: I believe so? The format for the API response is 

`{
  "trustResources": [
    {
      "description": "string",
      "resourceId": "string",
      "type": "string",
      "updatedAt": "2021-09-14T17:35:10.070Z",
      "updatedByUserUid": "string",
      "updatedByUsername": "string",
      "value": "string"
    }
  ]
}`

I assumed if there's no page tokens then there's no paging? Or is that not the same thing?

src/py42/clients/trustedactivities.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def get(self, id):
+        """Retrieve trusted activity details by given resource number.
+        `Rest documentation <https://developer.code42.com/>`
+
+        Args:
+            id (int): Resource number of the trusted activity or domain.
+
+        Returns:
+            :class:'py42.response.Py42Response'
```
alanag13, line 52: can we change this to be a generator?

src/py42/clients/trustedactivities.py
```diff
@@ -0,0 +1,82 @@
...
+        """Gets all trusted activities with the given type.
+        `Rest documentation <https://developer.code42.com/>`
+
+        Args:
+            type (str): Type of the trusted activity. `TrustedActivityType.DOMAIN` or `TrustedActivityType.SLACK`
+            value (str): The URL of the domain or name of the Slack workspace.
+            description (str, optional): Description of the trusted activity.
+
+        Returns:
+            :class:'py42.response.Py42Response'
```
alanag13, line 40: the character after `class:` and at the end of this line should be a backtick (`), not a single quote.

src/py42/clients/trustedactivities.py
tora-kozic: Yes! I assume you meant for the return of `get_all()` and not `get()` right?

src/py42/services/trustedactivities.py
```diff
@@ -0,0 +1,82 @@
...
+        uri = f"{self._uri_prefix}/{id}"
+        return self._connection.get(uri)
+
+    def update(self, id, value=None, description=None):
+        uri = f"{self._uri_prefix}/{id}"
+        current_activity_data = self.get(id).data
+        data = {
+            "type": current_activity_data.get("type"),
+            "value": value or current_activity_data.get("value"),
+            "description": description or current_activity_data.get("description"),
```
timabrmsn, line 57: This makes it impossible to clear the description from py42, since `""` evaluates to false. We should explicitly check for `description=None` here and only then use the current value.

alanag13: this bug is already present in the cases service (but the fix doesn't need to be handled in this PR). Good find.

src/py42/clients/trustedactivities.py
alanag13: correct, bad aim haha

src/py42/services/trustedactivities.py
alanag13: to be more clear: the fix for _cases_ doesn't need to be handled in this PR.

tora-kozic: Oh I just asked Tim this, alright!

alanag13 (approved):

PR #370, opened 2021-09-16
Chore/trust integration tests

tora-kozic:
### Description of Change ###

Added integration tests for the `trusted-activities` service endpoints

### Testing Procedure ###
Tests will not pass until the mock-servers PR which adds the trusted-activities service is merged

### PR Checklist ###

- [NA] Add unit tests to verify this change
- [NA] Add an entry to CHANGELOG.md describing this change

### DIFF
--- a/tests/integration/test_trustedactivites.py
+++ b/tests/integration/test_trustedactivites.py
@@ -16,6 +16,14 @@ class TestTrustedActivities:
             assert_successful_response(response)
             break
 
+    def test_get_all_trusted_activities_with_optional_params(
+        self, connection,
+    ):
+        page_gen = connection.trustedactivities.get_all(type="DOMAIN", page_size=1)
+        for response in page_gen:
+            assert_successful_response(response)
+            break
+
     def test_get_trusted_activity(self, connection, trusted_activity):
         response = connection.trustedactivities.get(trusted_activity["resourceId"])
         assert_successful_response(response)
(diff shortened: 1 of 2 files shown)


tests/integration/test_trustedactivites.py
```diff
@@ -0,0 +1,31 @@
...
+from tests.integration.conftest import assert_successful_response
+
+
+@pytest.mark.integration
+class TestTrustedActivities:
+    @pytest.fixture(scope="module")
+    def trusted_activity(self, connection):
+        return connection.trustedactivities.create("DOMAIN", "test.com")
+
+    def test_get_all_trusted_activities(
```
alanag13, line 11: does the mock server complain about unrecognized query params (i.e. us sending pgNum and pgSize even though they aren't defined yet?) The _real_ server won't of course, but I wasn't sure about the mock.

alanag13:
@tora-kozic you will also need to add this new service to the CI server's `/etc/hosts` in the `.github/workflows/build.yaml` file.

tests/integration/test_trustedactivites.py
tora-kozic: I haven't added the pgSize param to the mock-servers yet but it doesn't seem to complain about it as an unrecognized query param when sent in an integration test.

timabrmsn (approved):

PR #374, opened 2021-10-08
Bugfix: get all alert details empty results

timabrmsn:
### Description of Change ###

Resolves error raised when `sdk.alerts.get_all_alert_details(query)` resulted in no alerts found. If no alertIds are found in the query, the function makes no calls to `get_details()`.

### Testing Procedure ###
- Construct an AlertQuery which returns no results.
- Pass query to `result = sdk.alerts.get_all_alert_details(query)`
- Iterate over the resulting generator (e.g. `list(result)` `for item in result: print(item)`, `next(result)`)

Expected outcome: 
Iteration results in empty list or no action (depending on how the object was iterated over), no exceptions should be raised.

### PR Checklist ###

### DIFF
--- a/tests/clients/test_alerts.py
+++ b/tests/clients/test_alerts.py
@@ -170,6 +170,17 @@ def mock_alerts_service_with_pages(mocker, mock_alerts_service):
 
 
 @pytest.fixture
+def mock_alerts_service_with_no_alerts(mocker, mock_alerts_service):
+    no_alerts_page = create_mock_response(mocker, json.dumps({"alerts": []}))
+
+    def page_gen():
+        yield no_alerts_page
+
+    mock_alerts_service.search_all_pages.return_value = page_gen()
+    return mock_alerts_service
+
+
+@pytest.fixture
 def mock_details(mocker):
     detail_page_1 = create_mock_response(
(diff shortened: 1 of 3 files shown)


tests/clients/test_alerts.py
```diff
@@ -356,3 +367,23 @@ def test_alerts_client_get_all_alert_details_sorts_results_descending_when_speci
...
         results = list(alert_client.get_all_alert_details(query))
         assert results == [ALERT_F, ALERT_E, ALERT_D, ALERT_C, ALERT_B, ALERT_A]
+
+    def test_alerts_client_get_all_alert_details_returns_empty_generator_when_no_alerts_found(
+        self, mock_alerts_service_with_no_alerts, mock_alert_rules_service
+    ):
+        alert_client = AlertsClient(
+            mock_alerts_service_with_no_alerts, mock_alert_rules_service
+        )
+        query = AlertQuery()
```
tora-kozic, line 377: This is mostly just a clarifying question for my own understanding - but the `AlertQuery` constructor returns the `all()` method usually right? - And this is just returns no results because the testing env doesn't have any alerts?

tora-kozic (approved):

PR #377, opened 2021-10-15
removing diagonistics api

tora-kozic:
### Description of Change ###

Removed the diagnostics API


src/py42/services/administration.py
```diff
@@ -2,11 +2,6 @@
 
 
 class AdministrationService(BaseService):
-    def get_diagnostics(self, include_volumes=None):
```
tora-kozic, line 5: not sure if this requires a CHANGELOG entry since it sounds like no one could've been using it anyway ?

PR #388, opened 2021-11-29
Chore/update legal hold to api clients

tora-kozic:
### Description of Change ###

Add the legal hold endpoints which are supported by API Client authentication.  

To retain backwards compatibility, the existing legal hold client should be available when the SDK is initialized with any other authentication form.  If initialized with API Client auth, then the new updated legal hold client should be attached and these new endpoints should be available. 

Because this is a separate client, there are also new sets of unit and integration tests, as well as some small clarifying changes to variable names and error messages.

Breaking changes from the past client can be found in the CHANGELOG.

Because API Clients is not being released in py42 quite yet, this will target the `feature/api-clients` branch for merging.

### Testing Procedure ###
To initialize the sdk with API Client auth, use `py42.sdk.from_api_client(host, client, secret)`

the `client` and `secret` values can be created in the [console](https://console.core-int.cloud.code42.com/app/#/admin/api-clients)

### PR Checklist ###


src/py42/services/legalholdapiclient.py
```diff
@@ -0,0 +1,373 @@
...
+    _map = {True: "ACTIVE", False: "INACTIVE", None: "ALL"}
+    try:
+        return _map[active]
+    except KeyError:
+        raise Py42Error(
+            f"Invalid argument: '{active}'. active must be True, False, or None"
+        )
+
+
+class LegalHoldApiClientService(BaseService):
```
tora-kozic, line 24: It may be nicer to stick this into the `services/legalhold.py` file to prevent some redundancies and not have these wordy file names - but this follows our convention of a file for each service.

src/py42/sdk/__init__.py
```diff
@@ -359,7 +365,10 @@ def _init_services(main_connection, main_auth):
         administration=administration_svc,
         archive=ArchiveService(main_connection),
         devices=DeviceService(main_connection),
-        legalhold=LegalHoldService(main_connection),
+        # Only use updated legal hold client if initialized with API Client authorization
```
tora-kozic, line 368: Setting a flag which specifies how the sdk was initialized with the simplest solution I came up with for switching between which client is used - but I'm open to other solutions if there was something else in mind.

tora-kozic:
*Integration tests will pass once mock-server changes are approved and merged. I think CLA Assistant is failing due to the target merge branch not being `main`*

peterbriggs42:
Looks like the integration tests are failing because https://github.com/code42/code42-mock-servers/pull/20 hasn't merged yet. Is that PR ready to go? I think we should probably have that merge before this one.

src/py42/services/legalholdapiclient.py
```diff
@@ -0,0 +1,373 @@
...
+    """A service for interacting with Code42 Legal Hold APIs.
+
+    The LegalHoldService provides the ability to manage Code42 Legal Hold Policies and Matters.
+    It can:
+    - Create, view, and list all existing Policies.
+    - Create, view, deactivate, reactivate, and list all existing Matters.
+    - Add/remove Custodians from a Matter.
+    """
+
+    _uri_prefix = "/api/v27"
```
ryan-haley-code42, line 34: is v27 correct here? I know we don't follow the typical versioning, but I feel like I should still ask.

ryan-haley-code42 (approved):
Just the one v27 comment, but I assume it's correct. I'll approve here so this can be merged when the ends up being correct.

src/py42/services/legalholdapiclient.py
tora-kozic: Yeah, this isn't our versioning scheme, these endpoints are provided by the microservice (or in this case core) and don't go through the gateway.

src/py42/services/legalholdapiclient.py
alanag13: I generally err on the side of less scrolling

src/py42/sdk/__init__.py
alanag13: The general approach is fine but I think I would just use a boolean instead of defining a class/constant for this (I don't anticipate that we will be adding additional authentication methods that will require new API implementations any time soon).

PR #390, opened 2021-12-08
bugfix

tora-kozic:
The legal hold API endpoints for listing memberships and matters were formatted a little differently such that they returned a list of items without additional data.  This was throwing an error in the `get_all_pages()` method.  Adjusting this method so that it will return the response data if a key is not provided fixes this.

### DIFF
--- a/src/py42/services/legalholdapiclient.py
+++ b/src/py42/services/legalholdapiclient.py
@@ -193,7 +193,7 @@ class LegalHoldApiClientService(BaseService):
         """
         return get_all_pages(
             self.get_matters_page,
-            "legalHolds",
+            None,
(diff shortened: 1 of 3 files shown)


tora-kozic:
*CLA will fail because the target branch isn't `main`*

alanag13:
> _CLA will fail because the target branch isn't `main`_

I'm actually pretty sure that it's just failing all the time, there's something else going on with it.

alanag13:
I think the mocks in a couple of the legal hold unit tests do not currently reflect reality and this is making some tests fail.

tora-kozic:
Oops yeah I saw that, was in a meeting for a while but they should be good now

alanag13 (approved):

PR #392, opened 2021-12-10
Chore/update py42 exceptions

tora-kozic:
### Description of Change ###

Updating py42 custom exceptions to expose the args by
- Add and expose properties to each custom exception 
- Pass the raw args to the base class `__init__` method in while keeping the main string arg remains in the same position

The args are now printed in addition to the error message:
`py42.exceptions.Py42LegalHoldNotFoundOrPermissionDeniedError: ("Matter with UID '123' can not be found. Your account may not have permission to view the matter.", '123')`

Current behavior: 
`py42.exceptions.Py42LegalHoldNotFoundOrPermissionDeniedError: Matter with ID=123 can not be found. Your account may not have permission to view the matter.`

### Testing Procedure ###

Args should be visible in error message as well as accessible as properties of the exception.

### PR Checklist ###

### DIFF
--- a/tests/clients/test_detectionlists.py
+++ b/tests/clients/test_detectionlists.py
@@ -118,6 +118,7 @@ class TestDetectionListClient:
             "possibly still being created if you just recently created the Code42 "
             "user." in str(err.value)
         )
+        assert err.value.username == "testusername"
 
     def test_create_user_when_service_returns_bad_request_raises(
         self,
--- a/tests/clients/test_detectionlists.py
+++ b/tests/clients/test_detectionlists.py
@@ -113,11 +113,10 @@ class TestDetectionListClient:
             client.create_user("testusername")
 
         assert (
-            str(err.value)
-            == "Detection-list profiles are now created automatically on the server. "
+            "Detection-list profiles are now created automatically on the server. "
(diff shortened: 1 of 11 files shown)


tests/clients/test_detectionlists.py
```diff
@@ -113,11 +113,10 @@ def test_create_user_when_service_returns_cannot_find_user_bad_request_raises_un
...
 
         assert (
-            str(err.value)
-            == "Detection-list profiles are now created automatically on the server. "
+            "Detection-list profiles are now created automatically on the server. "
             "Unable to find a detection-list profile for 'testusername'. It is "
             "possibly still being created if you just recently created the Code42 "
-            "user."
+            "user." in str(err.value)
         )
```
timabrmsn, line 120: Can we also add some tests that the new properties are what we expect? Like: 

```
assert err.value.username == 'testusername'
```

timabrmsn (changes requested):

tests/clients/test_detectionlists.py
tora-kozic: oh yeah thats a good point, for sure!

timabrmsn (approved):

PR #395, opened 2021-12-23
Upgrade device

### LINKED ISSUE #394: Add support for device-upgrade v4 API
Based on some information I was able to see with developer tools, it would be nice to see the device-upgrade API be integrated into py42 or code42cli...

curl -X POST -k -H 'Content-type: application/json' -H 'Accept: application/json' https://console.us.code42.com/api/v4/device-upgrade/upgrade-device -d '{"deviceGuid":"<GUID>"}'

It would be nice if we could supply a list of email addresses, grab the GUID from the supplied list, and push the upgrade to the GUID's for the list of users.

Thanks!

ceciliastevens:
### Description of Change ###

Adds `sdk.devices.upgrade()` to upgrade an individual device. Currently administrators can do this via the console, adding it to py42 enables them to automate and have more flexibility in terms of upgrading devices in a phased manner instead of all-or-nothing.

### Issues Resolved ###

- closes #394 

### Testing Procedure ###

1. Create a device in the Code42 cloud running an older version of the client (using the delayed client upgrade functionality)
2. Call `sdk.devices.upgrade()` with that device's GUID
3. Verify that the authority received the instruction to upgrade this client

### PR Checklist ###


tests/services/test_devices.py
```diff
@@ -163,3 +164,13 @@ def test_get_page_when_org_not_found_raises_expected_error(
...
         assert err.value.org_uid == "TestOrgUid"
+
+    def test_upgrade_calls_upgrade_with_uri_and_params(
+        self, mock_connection, successful_response
+    ):
+        mock_connection.get.return_value = successful_response
+        service = DeviceService(mock_connection)
+        service.upgrade("DEVICE_ID")
+        expected_params = {"deviceGuid": "DEVICE_ID"}
+        uri = f"{UPGRADE_URI}"
```
unparalleled-js, line 175:
```suggestion
        uri = UPGRADE_URI
```

unparalleled-js: or can just pass `UPGRADE_URI` to `assert_called_once_with`

ceciliastevens: That makes sense, updated!

timabrmsn (approved):

PR #413, opened 2022-04-04
Feature/v2 file event apis

tora-kozic:
### Description of Change ###

Adds

- v2 query filters
- the py42.settings.use_v2_file_event_data settings
- toggles the API endpoint accordingly

### Testing Procedure ###

Ensure v2 file event data can be queried through the new api using the new v2 filters. 

**TODOs:**

- add links to support docs and/or developer portal API references when available
- add saved search v2 functionality (not yet released).  This includes adding a `shared` query filter term` and adding the new v2 API endpoint to the SavedSearch service. 
- Add deprecation notices and dates when those are known

### PR Checklist ###


src/py42/services/fileevent.py
```diff
@@ -24,14 +30,34 @@ def search(self, query):
...
+                "V1 file events are deprecated.  Set 'py42.settings.use_v2_file_event_data = True' to use V2 file events.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        # check if query type matches settings
+        if not isinstance(query, str):
+            if query.sort_key == "eventId" and settings.use_v2_file_event_data:
+                raise Py42Error(
+                    "Bad Request. You cannot use a V1 query with V2 settings enabled. For more details on working with V2 file event data, see the 'V2 File Events' user guide at https://py42docs.code42.com/."
```
timabrmsn, line 45: did we want to link to the guide directly here?

src/py42/services/savedsearch.py
```diff
@@ -1,36 +1,70 @@
...
+        _version = "v2" if settings.use_v2_file_event_data else "v1"
+        self._uri = f"/forensic-search/queryservice/api/{_version}/saved"
+        return self._uri
 
     def get(self):
         """Fetch details of existing saved searches.
 
+        The existing data model for file events and saved searches is deprecated.
+        To use the updated data model for file events, `update your settings <https://py42docs.code42.com/en/stable/userguides/v2apis.html>`__.
+        Retrieving saved searches with V2 settings enabled will convert existing saved search queries to the V2 data model.  Existing V1 queries that cannot be properly converted will be excluded from the response.
```
timabrmsn, line 28: Are the v1 searches still accessible via the v1 apis after conversion?

src/py42/sdk/queries/fileevents/v2/filters/__init__.py
```diff
@@ -0,0 +1,8 @@
+from py42.sdk.queries.fileevents.v2.filters.destination_filter import *
```
timabrmsn, line 1: We should import each filter class from each module individually here (and in v1), otherwise we're getting the internal stuff like `FileEventFilterStringField` and `Choices` classes imported, which just clogs up the namespace for users.

src/py42/sdk/queries/fileevents/v2/filters/destination_filter.py
```diff
@@ -0,0 +1,102 @@
+from py42.choices import Choices
+from py42.sdk.queries.fileevents.file_event_query import FileEventFilterStringField
+from py42.sdk.queries.query_filter import QueryFilterStringField
+
+
+class Destination(QueryFilterStringField):
```
timabrmsn, line 6: Can we distinguish the version in the docstring for each filter class? Since they have the same class name, it could help catch import bugs if REPL/IDE auto-complete could show you what version it was.

Maybe `"""V2 Filter Class that filters events based on the destination name."""` ?

timabrmsn:
We're not catching if you don't set the v2 setting, but try to construct + send a v2 query:
```
>>> from py42.sdk.queries.fileevents.v2 import *
>>> query = FileEventQuery(MD5.exists())
>>> 
>>> r = sdk.securitydata.search_file_events(query)
/Users/tim.abramson/IdeaProjects/py42/src/py42/clients/securitydata.py:46: DeprecationWarning: V1 file events are deprecated.  Set 'py42.settings.use_v2_file_event_data = True' to use V2 file events.
  return self._file_event_service.search(query)
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    r = sdk.securitydata.search_file_events(query)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/clients/securitydata.py", line 46, in search_file_events
    return self._file_event_service.search(query)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/services/fileevent.py", line 61, in search
    return self._connection.post(uri, json=query)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/services/_connection.py", line 141, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/services/_connection.py", line 211, in request
    _handle_error(method, url, response)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/services/_connection.py", line 289, in _handle_error
    raise_py42_error(ex)
  File "/Users/tim.abramson/IdeaProjects/py42/src/py42/exceptions.py", line 619, in raise_py42_error
py42.exceptions.Py42BadRequestError: Failure in HTTP call 400 Client Error: Bad Request for url: https://forensicsearch-default.prod.ffs.us2.code42.com/forensic-search/queryservice/api/v1/fileevent. Response content: {"timestamp":"2022-05-09T19:35
:45.310+00:00","status":400,"error":"Bad Request","path":"/forensic-search/queryservice/api/v1/fileevent"}
```

src/py42/services/fileevent.py
tora-kozic: I wasn't sure if links would make sense in something like an error message? If we think that would be appropriate we certainly could though

src/py42/sdk/queries/fileevents/v2/filters/__init__.py
tora-kozic: For v2, what do you think of switching it so that we actually import all the modules (rather than the classes from the modules) and the classes can be renamed and accessed more similarly to how they appear in the file event json?  This would also make looking for filter values easier (imo) because an IDE could prompt all possible filters relating to the `destination` module.

ie - 
`from py24.sdk.queries.fileevents.v2.filters import *`
`filter = destination.Name  # destination.py module, class Name()`

src/py42/services/fileevent.py
timabrmsn: Yeah, I'm of the mindset that error messages should be as helpful as possible, so the link is good, I was just saying we should link directly to the guide: https://py42docs.code42.com/en/stable/userguides/v2apis.html like we do in the docstrings.

src/py42/sdk/queries/fileevents/v2/filters/__init__.py
timabrmsn: I like it

src/py42/sdk/queries/fileevents/v2/filters/destination.py
```diff
@@ -0,0 +1,92 @@
+from py42.choices import Choices
+from py42.sdk.queries.fileevents.util import _FileEventFilterStringField
+from py42.sdk.queries.query_filter import _QueryFilterStringField
+
+
+class Destination:
```
tora-kozic, line 6: I'm not _crazy_ about the nested classes. The method docs also get pretty crammed. In general I think there's a better way we could organize the filters, particularly with the V2 data model.  I didn't want to completely change how they work here but maybe something to consider with the future Incydr sdk.

docs/methoddocs/fileeventqueriesv2.md
```diff
@@ -0,0 +1,141 @@
...
+construct `FilterGroup`s that will search for matching filenames or (in case someone renamed the sensitive file) the
+known MD5 hashes of the files:
+
+    from py42.sdk.queries.fileevents.v2 import *
+    filename_filter = File.Name.is_in(['confidential_plans.docx', 'confidential_plan_projections.xlsx'])
+    md5_filter = File.MD5.is_in(['133765f4fff5e3038b9352a4d14e1532', 'ea16f0cbfc76f6eba292871f8a8c794b'])
+
+See [Executing Searches](../userguides/searches.md) for more on building search queries.
+
+### Filter Group Helper Methods
```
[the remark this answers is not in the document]
tora-kozic: I agree, they were in the original docs so I just copied them over but it seemed unnecessary.  The docstrings for each method would imply someone thought they'd be useful by end-users eventually but I'm not sure that's true at this point: 
`def create_exists_filter_group(term):
    """Creates a :class:`~py42.sdk.queries.query_filter.FilterGroup` to find events where
    filter data exists. Useful for creating ``EXISTS`` filters that are not yet supported
    in py42 or programmatically crafting filter groups.`

tora-kozic:
- [ ] fix integration tests

src/py42/sdk/queries/fileevents/v2/filters/file.py
```diff
@@ -0,0 +1,87 @@
+from py42.choices import _Choices
+from py42.sdk.queries.fileevents.util import _FileEventFilterComparableField
+from py42.sdk.queries.fileevents.util import _FileEventFilterStringField
+
+
+class Category(_FileEventFilterStringField, _Choices):
```
timabrmsn, line 6: This is missing a `_term` attribute.

src/py42/services/savedsearch.py
```diff
@@ -1,80 +1,135 @@
...
+        # deprecation warning for v1 file events
+        if not use_v2:
+            warn(
+                "V1 file events and saved searches are deprecated.  Use v2 apis by passing in the optional use_v2=True arg.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+
+        self._version = "v2" if use_v2 else "v1"
+        print(self.uri)
```
timabrmsn, line 49: remove print

tests/clients/test_securitydata.py
```diff
@@ -13,18 +14,20 @@
 from py42.services.storage.exfiltrateddata import ExfiltratedDataService
 from py42.services.storage.preservationdata import StoragePreservationDataService
 
+py42.settings.use_v2_file_event_data = False
```
timabrmsn, line 17: not needed anymore

timabrmsn: since `py42.settings` is a module, it's just creating the variable `use_v2_file_event_data` in that namespace.

timabrmsn (approved):

tests/clients/test_securitydata.py
tora-kozic: ah right, module not a class

PR #421, opened 2022-04-26
Bugfix/handle empty archives

timabrmsn:
### Description of Change ###

`py42.response.Py42Response`'s handling of the "data" property fails when the key "data" exists, but the value is false-y, like an empty list. If the "data" exists, we want to return it, regardless of its truthiness. 

This came up in the `stream_to_device()` push restore method, when an archive didn't have any files in it, the list of files was empty, but because our processing assumed a list (but actually got a dict, due to this bug), it would throw the following: 

```
line 72, in _get_file_via_walking_tree
  if root["path"].lower() == path_root.lower():
TypeError: string indices must be integers
```

### Issues Resolved ###
INTEG-2160

### PR Checklist ###

### DIFF
--- a/src/py42/response.py
+++ b/src/py42/response.py
@@ -111,7 +111,8 @@ class Py42Response:
             if not self._data:
                 response_dict = json.loads(self._response.text)
                 if type(response_dict) == dict:
-                    self._data = response_dict.get("data") or response_dict
+                    data = response_dict.get("data")
+                    self._data = data if data is not None else response_dict
                 else:
                     self._data = response_dict
         except ValueError:
--- a/src/py42/response.py
+++ b/src/py42/response.py
@@ -111,8 +111,10 @@ class Py42Response:
             if not self._data:
                 response_dict = json.loads(self._response.text)
                 if type(response_dict) == dict:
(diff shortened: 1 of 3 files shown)


src/py42/response.py
```diff
@@ -111,7 +111,10 @@ def _data_root(self):
             if not self._data:
                 response_dict = json.loads(self._response.text)
                 if type(response_dict) == dict:
-                    self._data = response_dict.get("data") or response_dict
+                    if "data" in response_dict:
+                        self._data = response_dict["data"]
+                    else:
+                        self._data = response_dict
```
tora-kozic, line 117: So if there's no `"data"` key, we just set the property to the entire response dictionary?

timabrmsn: Yep, it'a a convenience feature, as a lot of APIs return results like: 
```{"metadata": ..., "data": "the_stuff_you_want"}```
so if there is a "data" field we make it directly accessible on the Response.data property, you can do `response["thing_I_want"]` instead of having to do `response["data"]["thing_I_want"]` every time.

tora-kozic (approved):

PR #422, opened 2022-04-27
Bugfix/watchlist errors

timabrmsn:
### Description of Change ###

The `included-users/add` & `included-users/delete` endpoints don't actually 404 on bad watchlistId or userId values. They return a 400 instead. Added logic to catch those and raise appropriate "not found" exceptions instead.


tora-kozic:
I know when I was testing the `get-watchlist-member()` command it returned a 404 without specifying which resource it couldn't find.  Although I should've accounted for the user_ids in the original error handling - it is odd that the included-users endpoints return something completely different. 🤔

tora-kozic (approved):

PR #423, opened 2022-04-27
adjust `userriskprofile.update`'s null date handling

timabrmsn:
### Description of Change ###

`sdk.userriskprofile.update()` when getting `None` or `""` for end_date or start_date was sending: 
```
{"startDay": {"day": null, "month": null, "year": null}
``` 
This was causing errors on the backend trying to parse the date values. 

To clear the value (or when not updating at all), we need to send just:
```
{"startDay": null}
```


tora-kozic:
Looks like the unit tests just need to be adjusted

tora-kozic (approved):

PR #424, opened 2022-04-29
support specifying `backup_set_id` in restores

timabrmsn:
### Description of Change ###

V1 archives aren't backup set aware, the client sending backup may have different sets with different file selections and priorities, but they all end up in the same file tree in a V1 archive. 

The initial py42 restore method implementations were designed for V1 archives, where it didn't matter which backup_set_id was supplied in the restore request. We just defaulted to backup_set_id=1 (the default backup set) if it existed, otherwise we picked the first from the list when calling `archive.get_backup_sets()`.

For V3 archives, backup set information is tracked in the archive itself, so if a client has two backup sets with different file selections, trying to restore a file that was backed up from backup_set_id=2 while submitting backup_set_id=1 in the restore request will result in a "file not found" error, even though the file _is_ in the archive.

So our behavior of "select the backup_set_id automatically" can make it impossible to restore some V3 archive files if they're not in the "default" backup set.

This change just adds the option to specify the backup_set_id in the restore methods. If not provided by the user, the original default behavior of selecting id=1 or else the first in the list still stands.

### PR Checklist ###

### DIFF
--- a/src/py42/clients/archive.py
+++ b/src/py42/clients/archive.py
@@ -198,7 +198,9 @@ class ArchiveClient:
         backup_set_ids = [bs["backupSetId"] for bs in backup_sets]
         if backup_set_id:
             if backup_set_id not in backup_set_ids:
-                raise Py42Error(f"backup_set_id={backup_set_id} not found in device backup sets: {backup_sets}")
+                raise Py42Error(
+                    f"backup_set_id={backup_set_id} not found in device backup sets: {backup_sets}"
+                )
             return backup_set_id
         # id=1 is the "default" backup set, use this if it exists and no explicit id
         # is provided by user
@@ -207,7 +209,7 @@ class ArchiveClient:
         elif len(backup_set_ids) > 0:
             return backup_set_ids[0]
         else:
-            raise Py42Error(f"Failed to get backup sets for device.")
+            raise Py42Error("Failed to get backup sets for device.")
 
     def get_backup_sets(self, device_guid, destination_guid):
         """Gets all backup set names/identifiers referring to a single destination for a specific
--- a/src/py42/clients/archive.py
+++ b/src/py42/clients/archive.py
@@ -50,6 +50,7 @@ class ArchiveClient:
         encryption_key=None,
         show_deleted=None,
         file_size_calc_timeout=_FILE_SIZE_CALC_TIMEOUT,
+        backup_set_id=None,
     ):
         """Streams a file from a backup archive to memory. This method uses the same endpoint
         as restoring from Console and therefore has all the same considerations.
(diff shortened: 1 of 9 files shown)


src/py42/clients/archive.py
```diff
@@ -184,16 +193,19 @@ def stream_to_device(
...
+        if backup_set_id:
+            if backup_set_id not in backup_set_ids:
+                raise Py42Error(f"{backup_set_id=} not found in device backup sets: {backup_sets}")
+            return backup_set_id
+        # id=1 is the "default" backup set, use this if it exists and no explicit id
+        # is provided by user
+        elif "1" in backup_set_ids:
+            return "1"
+        else:
+            return backup_set_ids[0]
```
tora-kozic, line 208: Do we still need to catch this case of no backup_sets being returned? Looks like it may KeyError in the `else` statement

src/py42/clients/archive.py
```diff
@@ -184,16 +193,19 @@ def stream_to_device(
...
-        for backup_set in backup_sets:
-            backup_set_id = backup_set["backupSetId"]
-            if str(backup_set_id) == _DEFAULT_BACKUP_SET_ID:
-                return backup_set_id
-
-        return backup_sets[0]["backupSetId"]
+        backup_set_ids = [bs["backupSetId"] for bs in backup_sets]
+        if backup_set_id:
+            if backup_set_id not in backup_set_ids:
+                raise Py42Error(f"{backup_set_id=} not found in device backup sets: {backup_sets}")
```
tora-kozic, line 201: `{backup_set_id=}` oops

src/py42/clients/archive.py
timabrmsn: I'm pretty sure it's not possible for a device to have _no_ backup sets at all, and if we do return `None` the restore requests would fail anyway as it's a required value. 

I changed it to raise an exception if no sets are found, to fail fast and explicitly if we do ever hit this case.

src/py42/clients/archive.py
timabrmsn: That's actually a feature of f-strings:

```
Python 3.8.10 (default, Jun  9 2021, 16:13:34) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> x = 1
>>> print(f"{x=}")
x=1
```

But in looking it up again I did realize it was introduced in python3.8 so we can't use it here anyway.

tora-kozic: Whoops yeah should’ve mentioned it looked like it was only failing on earlier python versions 👍

tora-kozic (approved):

PR #425, opened 2022-05-11
Bugfix/ffs rate limit handler

### LINKED ISSUE #225: [Enhancement] Support Rate Limit Throttling
## Summary
Per [Code 42 Documentation](https://support.code42.com/Administrator/Cloud/Monitoring_and_managing/Code42_API_resources/Forensic_Search_API#:~:text=Request%20rate%3A%20To%20ensure%20optimal,requests%20to%20120%20per%20minute.), there is a limit on API requests that can be made. That request is fairly low and as a result folks are likely to run into it. This will result in `py42.exceptions.Py42HTTPError: Failure in HTTP call 429 Client Error: Too Many Requests for url`. 

## Proposed API

The easiest way to fix this would be to make a specific exception for rate limits so those could be easily caught, of course the more elegant solution is to allow a configurable rate limit and back off parameter. 

## Intended Use Case
Supporting automation using the API.

timabrmsn:
### Description of Change ###

Adds a custom HTTPAdapter mounted to just the FileEventService url that automatically retries 429 (too many requests) errors, and logs at INFO level that we've hit the rate limit and will sleep for _n_ seconds (_n_ being either what the FFS response header tells us, or if that's not present, a 5 second exponential backoff).

### Issues Resolved ###
- closes #225

### Testing Procedure ###
Construct a simple FFS query, and rapidly make the same request ~130 times (the rate limit kicks in at 120 requests per minute), e.g.:

```
import py42.settings.debug as debug
debug.level = "INFO"

query = FileEventQuery(FileName.eq("test.txt"))
for i in range(130):
   print(i)
   r = sdk.securitydata.search_file_events(query)
```
Without this fix, a `Py42TooManyRequestsError` will be raised.

With this fix, at the 120th request, a message will be logged that the rate limit was hit, and the request will be retried after n seconds. Eventually all 130 requests will succeed without any intervention.

### PR Checklist ###

- [N/A] Add unit tests to verify this change


src/py42/services/fileevent.py
```diff
@@ -1,18 +1,48 @@
...
 
 
+class FFSQueryRetryStrategy(Retry):
+    """The forensic search service helpfully responds with a 'retry-after' header, telling us how long until the rate
+    limiter is reset. We subclass :class:`urllib3.Retry` just to add a bit of logging so the user can tell why the
+    request might look like it's hanging.
+    """
+
+    def get_retry_after(self, response):
+        retry_after = super().get_retry_after(response)
```
tora-kozic, line 20: Its cool that theres already a `Retry` class for this sort of thing !

tora-kozic (approved):

PR #427, opened 2022-05-13
alias filter classes

tora-kozic:
Undo class renaming in favor of aliasing within the filter modules - for backwards compatibility with imports


timabrmsn (approved):

tora-kozic:
Going to add this aliasing to the alerts filters as well

PR #430, opened 2022-05-23
Feature/api client users

tora-kozic:
### Description of Change ###

Switches the `users` client over to API client compatible endpoints.  Additional lookups are used to preserve backwards compatibility.

### PR Checklist ###


tests/services/test_users.py
```diff
@@ -375,3 +498,101 @@ def test_update_user_when_get_unhandled_internal_server_error_raises_base_error(
         mock_connection.put.side_effect = internal_server_error
         with pytest.raises(Py42InternalServerError):
             user_service.update_user("123", username="foo")
+
+    @patch.object(
```
tora-kozic, line 502: Just found out about the `patch.object` decorator/method!  Very helpful for patching out a helper method that makes an api call

tests/services/test_users.py
```diff
@@ -192,27 +266,76 @@ def test_get_available_roles_calls_get_with_expected_uri(self, mock_connection):
...
-        assert mock_connection.post.call_args[0][0] == uri
-        assert mock_connection.post.call_args[1]["json"]["roleName"] == "Test Role Name"
-        assert mock_connection.post.call_args[1]["json"]["userId"] == 12345
+        role_ids = ["role-1", "role.name.2"]
+        service._update_roles(12345, role_ids)
+        uri = f"/api/v3/users/{TEST_USER_UID}/roles"
+        data = {"roleIds": role_ids}
+        mock_connection.put.assert_called_once_with(uri, json=data)
+
+    def test_update_role_ids_calls_get_available_roles_and_returns_role_ids_when_add(
```
tora-kozic, line 310: opted to unit test the helper methods rather than the full `add/remove_role` functions due to all the additional lookups necessary

tora-kozic:
- [ ] Integration tests failing until mock-servers are updated

src/py42/services/users.py
```diff
@@ -242,10 +251,10 @@ def deactivate(self, user_id, block_user=None):
         Returns:
             :class:`py42.response.Py42Response`
         """
-        uri = f"/api/v1/UserDeactivation/{user_id}"
+        uri = f"/api/v3/users/{self._get_user_uid_by_id(user_id)}/deactivate"
         data = {"blockUser": block_user}
```
timabrmsn, line 255: the query params for this and reactivate got changed on the new apis to just `block` and `unblock` instead of `blockUser`/`unblockUser`

tora-kozic: Oh cool, they must be backwards compatible or something because they were working as is, but I'll update it to the new values 👍

timabrmsn: It wouldn't error, but the incorrect params would just be ignored. If you tried `deactivate(user_id, block_user=False)` (opposite of default), the user would still be blocked on deactivation.

tora-kozic: Ohh I see, I must've been testing it against the default values without knowing so the output looked right 🧐

src/py42/services/users.py
```diff
@@ -380,3 +393,29 @@ def update_user(
...
+            if role["roleName"] == role_name:
+                if add:
+                    role_ids.append(role["roleId"])
+                else:
+                    role_ids.remove(role["roleId"])
+                break
+
+        return role_ids
+
+    def _update_roles(self, user_id, role_ids):
```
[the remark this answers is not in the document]
timabrmsn: Since we're already doing the lookup either way, we could just accept role-id _or_ role-name values and use them if they're valid, just make the check `if (role["roleName"] == role_name) or (role["roleId"] == role_name):`

Then just update the docs like:

```
role_name (str): The name or id of the role to assign to the user. e.g. "Desktop User" (name) or "desktop-user" (id)
```

This keeps it backwards compatible, but makes it easy to re-use the role-id values. I was testing these changes in the CLI and noticed the CLI only returns `role-id` in user results, but requires `role-name` in the arguments, making this change would make that awkward inconsistency go away.

timabrmsn (approved):

PR #432, opened 2022-05-25
Feature/api clients orgs

tora-kozic:
### Description of Change ###

Replaces the APIs for the following methods with the new `v3/orgs` endpoints with api client support.
- create
- block/unblock
- activate/deactivate

The GET and PUT endpoints for v3/orgs are not yet released, but the existing endpoints do support API clients.  TODO's in the code indicate places where those future changes will need to be made.  TBD whether waiting for those is in the scope of this PR.

### PR Checklist ###

- [NA] Add an entry to CHANGELOG.md describing this change

### DIFF
--- a/src/py42/services/orgs.py
+++ b/src/py42/services/orgs.py
@@ -209,8 +209,8 @@ class OrgService(BaseService):
         """
         return self.get_agent_state(org_id, "fullDiskAccess")
 
-    # TODO - status of this org settings with api clients
-
+    # TODO - this method won't be supported by API clients.  A plan for how to handle this and similar methods
+    # TODO - will need to be determined prior to a full py42 api-client support release.
     def get_settings(self, org_id):
         """Gets setting data for an org and returns an `OrgSettingsManager` for the target org.
 
@@ -227,6 +227,8 @@ class OrgService(BaseService):
         t_settings = self._connection.get(uri)
         return OrgSettings(org_settings.data, t_settings.data)
 
+    # TODO - this method won't be supported by API clients.  A plan for how to handle this and similar methods
+    # TODO - will need to be determined prior to a full py42 api-client support release.
     def update_settings(self, org_settings):
         """Updates an org's settings based on changes to the passed in `OrgSettings` instance.
 
--- a/src/py42/services/orgs.py
+++ b/src/py42/services/orgs.py
@@ -2,7 +2,8 @@ from collections import namedtuple
 
 from py42 import settings
 from py42.clients.settings.org_settings import OrgSettings
-from py42.exceptions import Py42Error, Py42InternalServerError
+from py42.exceptions import Py42Error
+from py42.exceptions import Py42InternalServerError
 from py42.services import BaseService
 from py42.services.util import get_all_pages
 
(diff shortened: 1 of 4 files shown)


src/py42/services/orgs.py
```diff
@@ -33,12 +34,16 @@ def create_org(self, org_name, org_ext_ref=None, notes=None, parent_org_uid=None
         Returns:
             :class:`py42.response.Py42Response`
         """
-        uri = "/api/v1/Org/"
+
+        # get parent org GUID from UID :)
+        parent_org_guid = self._get_guid_by_id(parent_org_uid, id_key="orgUid")
```
tora-kozic, line 39: We should add more kinds of ids 🙂

tora-kozic:
- [ ] green build pending mock-server update https://github.com/code42/code42-mock-servers/pull/30

src/py42/services/orgs.py
```diff
@@ -232,3 +260,13 @@ def update_settings(self, org_settings):
             org_response=org_response,
             org_settings_response=org_settings_response,
         )
+
+    def _get_guid_by_id(self, org_id, id_key="orgId"):
```
timabrmsn, line 264: I wonder if we should cache the results of this method, so we're not looping through every org for every individual org interaction.

tora-kozic: Thats a good idea,  so like every time you call it, it stores an entry of `{"orgId": "orgGuid"}` in a prop like we did on watchlists? (seems like the Uid's are used rarely enough that that caching wouldn't be super necessary)

src/py42/services/orgs.py
```diff
@@ -182,6 +214,8 @@ def get_agent_full_disk_access_states(self, org_id):
         """
         return self.get_agent_state(org_id, "fullDiskAccess")
 
+    # TODO - status of this org settings with api clients
```
tora-kozic, line 217: The is an important `TODO`.  With the latest apis, a lot of the old org settings code doesn't work well with artemis only tenants, because they lack the expected backup settings. 

We should decide how we want to handle `get_settings` and `update_settings` for api-clients. I would favor on the side of marking them as deprecating them/marking them incompatible with api-clients, especially since I'm not sure its even possible to make them fully backwards compatible - but I'm open to suggestions. We could certainly make them _as close to backwards compatible as possible_, but the new endpoint for updating orgs doesn't include any backup configuration.

tora-kozic:
- [ ] update mock-servers to add get/put org apis (released in authority 22.0.0)

src/py42/services/orgs.py
```diff
@@ -79,8 +103,8 @@ def get_page(self, page_num, page_size=None, **kwargs):
         Returns:
             :class:`py42.response.Py42Response`
         """
+        uri = "/api/v3/orgs"
```
tora-kozic, line 106: I don't think this new api implements paging, so I'll add some docs around that

src/py42/services/orgs.py
timabrmsn: I'm fine with just leaving this as a TODO for now. Before we do a release with api client support, we'll need to come up with a strategy for disabling/raising on things that don't/won't support them (like this and the web/push restore methods). But we don't quite need to come up with that strategy _now_.

src/py42/services/orgs.py
timabrmsn: Does it just always return all results in a single request?

tora-kozic: yeah, that's my understanding

PR #442, opened 2022-07-27
Chore/pl 104371 replace login token

timabrmsn:
### Description of Change ###

Removes the LoginToken auth dance with the storage node, now storage Connections can use just the same authority oauth token to validate storage API calls. 

### Testing Procedure ###
Confirm `sdk.archive.stream_from_backup()` and `sdk.archive.stream_to_device()` methods work with the updated changes.

### PR Checklist ###

- [N/A] Add an entry to CHANGELOG.md describing this change


tora-kozic:
looks like you'll need a mock-server update for a passing build

tora-kozic:
so basically the change is that previously, we got a separate auth token to connect to the storage node, and now it can just use the same oauth token as the other api calls?

timabrmsn:
> so basically the change is that previously, we got a separate auth token to connect to the storage node, and now it can just use the same oauth token as the other api calls?

Yep, previous flow was: 
- Call /LoginToken api on authority
- Use that token to call AuthToken on the storage node itself
- Storage node would validate login token and issue its own auth token for future storage node api calls

Now the storage node just accepts the same oauth token we use for authority calls.

tora-kozic (approved):

PR #466, opened 2024-11-26
Integ 2834/remove file download by path

ceciliastevens:
### Description of Change ###

Updates the stream file methods in the securitydata client to use the hash when requesting a download token.

### PR Checklist ###

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ how a consumer would use the library (e.g. adding unit tests, updating documenta
 #### Fixed
 
 - An issue where the page size of file event queries was higher than intended.
+- An issue where streaming file by hash did not use the hash.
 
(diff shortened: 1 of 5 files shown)


tora-kozic:
The changes generally look good & minimal, but ideally we do get all the tests passing in the build tasks here.  
That said, CLA is broken and can be ignored, and the `style` here task can be looked at/fixed locally by running the command `tox -e style`

ceciliastevens:
> The changes generally look good & minimal, but ideally we do get all the tests passing in the build tasks here. That said, CLA is broken and can be ignored, and the `style` here task can be looked at/fixed locally by running the command `tox -e style`

Ok, updated. We should be good on the tests and style checks now.

johnsonjben (approved):

PR #471, opened 2025-03-18
deprecate incydr functionality

ceciliastevens:
Marks functionality that is superseded by the Incydr SDK (https://developer.code42.com/) as deprecated.


tora-kozic (commented):
Have we tested this alongside the cli update? Do the py42 deprecations carry over to that? 

If we add a deprecation warning to the client itself, does every command also print that out (ie. if we add it to the client and the command does it do it twice?)

ceciliastevens:
> Have we tested this alongside the cli update? Do the py42 deprecations carry over to that?
> 
> If we add a deprecation warning to the client itself, does every command also print that out (ie. if we add it to the client and the command does it do it twice?)

Yes, the code42cli prints out warnings generated by its dependencies, including py42, so if we make these changes together both the SDK's warning and the CLI's warning will print.

jdullum (approved):

tora-kozic (approved):

PR #478, opened 2025-05-05
download files from EDS instead of PDS

ceciliastevens:
Updates the file download methods to use EDS instead of PDS. Also removes the old fallback to downloading from backup archives, as that is no longer used now that we're on EDS.


ceciliastevens:
Builds are failing because the mock-servers used for integration testing don't yet handle the EDS url in the key-value store.

src/py42/services/storage/exfiltrateddata.py
```diff
@@ -1,37 +1,25 @@
...
-    def __init__(self, main_session, streaming_session):
-        super().__init__(main_session)
+    def __init__(self, main_connection, streaming_session):
+        super().__init__(main_connection)
         self._streaming_session = streaming_session
 
-    def get_download_token(
-        self, event_id, device_id, file_path, file_sha256, timestamp
-    ):
+    def get_download_token(self, downloadRequestUrl):
```
tora-kozic, line 12: Is this the only customer using this as far as we know? Changing the signature is technically a breaking change - but if its currently broken and/or no ones using it, then its probably okay.  
Maybe we could add details about how to adopt the fix in the changelog?

tora-kozic:
If we're planning to release this right after merging, feel free to do the version bump and the changelog release date in the same PR

src/py42/services/storage/exfiltrateddata.py
ceciliastevens: Added a note to the changelog.

tora-kozic (approved):

